PR #14, opened 2023-09-21
QEMU

macpijan:
1. Start QEMU with (fw from: https://docs.dasharo.com/variants/qemu_q35/building-manual/#procedure)

```
#!/usr/bin/env bash

qemu-system-x86_64 \
  -machine q35 \
  -drive if=pflash,format=raw,readonly=on,file=/usr/share/OVMF/OVMF_CODE.fd \
  -drive if=pflash,format=raw,file=/tmp/OVMF_VARS.fd \
  -net none \
  -qmp unix:/tmp/qmp-socket,server,nowait \
  -serial telnet:localhost:1234,server,nowait \
  -nographic
  ```

2. Start test with

```
robot -L TRACE -v config:qemu -v rte_ip:127.0.0.1 -v snipeit:no dasharo-compatibility/uefi-shell.robot

Uefi-Shell                                                                    
USH001.001 UEFI Shell :: Check whether the DUT has the ability to ... | PASS |
Uefi-Shell                                                            | PASS |
1 test, 1 passed, 0 failed
Output:  /home/macpijan/projects/github/dasharo/open-source-firmware-validation/output.xml
Log:     /home/macpijan/projects/github/dasharo/open-source-firmware-validation/log.html
Report:  /home/macpijan/projects/github/dasharo/open-source-firmware-validation/report.html
```

Current problems:
- we are really tied to the RTE in our setup - we expect each platform uses it, we expect we are able to SSH into it, verify CPU ID and stuff,
- we use OVMF, not edk2 payload, and intial phase of boot looks different on serial - especially the `Press KEY to enter setup` etc is not printed, so all our keywords won't work; current FW boots straight into EFI shell (we can interrupt with DEL, but there is no message printed about such possibility on serial)

### DIFF
--- a/.github/workflows/qemu-self-test.yml
+++ b/.github/workflows/qemu-self-test.yml
@@ -1,0 +1,55 @@
+name: Keywords self-tests with QEMU
+
+on: [push, pull_request]
+
+jobs:
+  qemu:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repository and submodules
+      uses: actions/checkout@v2
+      with:
+        submodules: 'recursive'
+
+    - name: Set up QEMU
+      run: |
+         sudo apt-get update
+         sudo apt-get install qemu-system-x86-64
+
+    - name: Download firmware files for QEMU
+      run: |
+        wget -O OVMF_CODE.fd https://cloud.3mdeb.com/index.php/s/iXmyJxKLBB2Eb94/download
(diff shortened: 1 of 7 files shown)


macpijan:
@pietrushnic With these changes: https://github.com/Dasharo/edk2/pull/91
I am able to run the same uefi-shell basic test as above, but on OVMF built from our code

PR #22, opened 2023-09-26
Flashing test for Yocto system using SD Wire setup

WiktorG351:
This PR is still a work in progress, it attempts to add some keywords related to SD Wire flashing, in order to build a base on which we will test Rpi related products.


keywords.robot
```diff
@@ -2978,3 +2978,26 @@ Should Contain All
     FOR    ${substring}    IN    @{substrings}
         Should Contain    ${string}    ${substring}
     END
+
+Flash SD Wire
```
artur-rs, line 2982: @WiktorG351 name is misleading because we are not flashing SD Wire, but "with/via" this device

artur-rs: @WiktorG351 another issue is that we are missing `[Arguments]` section

keywords.robot
```diff
@@ -2978,3 +2978,26 @@ Should Contain All
...
+    Variable Should Exist    ${file_gz}
+
+    SSHLibrary.Put File    ${file_bmap}    /data/    scp=ON
+    SSHLibrary.Put File    ${file_gz}    /data/    scp=ON
+
+    # sonoff turn off power should be here
+
+    SSHLibrary.Execute Command    sd-mux-ctrl -e=sd-wire_01-80 --ts
+    SSHLibrary.Execute Command    umount /dev/sda*
+    SSHLibrary.Execute Command    bmaptool copy --bmap /data/${file_bmap} /data/${file_gz} /dev/sda
```
artur-rs, line 2997: @WiktorG351 each command should have separate pass/fail check, without this approach it will hard to debug the possible problems with procedure

keywords.robot
```diff
@@ -2978,3 +2978,26 @@ Should Contain All
...
+
+    SSHLibrary.Execute Command    sd-mux-ctrl -e=sd-wire_01-80 --dut
+
+    # sonoff turn on power should be here
+
+    SSHLibrary.Close All Connections
```
artur-rs, line 3003: @WiktorG351 this should be moved to test suite teardown, because can break test-cases/test-suites

keywords.robot
WiktorG351: @artur-rs adressed [here](https://github.com/Dasharo/open-source-firmware-validation/blob/a5528a512d760cae06aa057a36a7ecb3293e5146/keywords.robot#L2982) and [here](https://github.com/Dasharo/open-source-firmware-validation/blob/a5528a512d760cae06aa057a36a7ecb3293e5146/keywords.robot#L2985)

keywords.robot
WiktorG351: @artur-rs i added a [Get Status Of SD Wire](https://github.com/Dasharo/open-source-firmware-validation/blob/a5528a512d760cae06aa057a36a7ecb3293e5146/keywords.robot#L3034) that checks whether the command executes successfully. I am calling it each time after the test changes the status, precisely [here](https://github.com/Dasharo/open-source-firmware-validation/blob/a5528a512d760cae06aa057a36a7ecb3293e5146/keywords.robot#L2997-L2999) and [here](https://github.com/Dasharo/open-source-firmware-validation/blob/a5528a512d760cae06aa057a36a7ecb3293e5146/keywords.robot#L3004-L3006)

keywords.robot
```diff
@@ -2978,3 +2978,70 @@ Should Contain All
     FOR    ${substring}    IN    @{substrings}
         Should Contain    ${string}    ${substring}
     END
+
```
macpijan, line 2981: Please conform to: https://github.com/Dasharo/open-source-firmware-validation#guidelines before moving from draft

macpijan:
@WiktorG351 You should be using pre-commit even during development. It is really helpful.

WiktorG351:
hi @macpijan , I have made changes to make the PR adhere to contributing guidelines, mainly moved the SD Wire related keywords from `keywords.robot` to a separate resource `.robot` file in `mobiqam` directory. I also added a RTE to the lab, and included it in `variables.robot`.

requirements-openbmc.txt
```diff
@@ -22,3 +22,4 @@ robotframework-angularjs
 robotframework-xvfb
 websocket-client
 pyotp
+pre-commit
```
artur-rs, line 25: @macpijan should I remove such change?

requirements.txt
```diff
@@ -15,3 +15,4 @@ unidecode
 robotframework-robocop
 robotframework-tidy
 robotframework-sherlock
+pre-commit
```
artur-rs, line 18: @macpijan should I remove such change?

yocto-base/00_flash_and_verify.robot
```diff
@@ -0,0 +1,50 @@
...
+Suite Teardown      Log Out And Close Connection
+
+
+*** Test Cases ***
+Verify number of connected SD Wire devices
+    [Documentation]    Test confirms the number of connected SD Wire devices to
+    ...    the DUT platform.
+    ${sd_wire_id_list}=    Get List Of SD Wire Ids
+    ${length_of_sd_wire_id_list}=    Get Length    ${sd_wire_id_list}
+    # we are currently doing this because we only expect one SD Wire
```
artur-rs, line 23: @WiktorG351 let's rephrase this comment, to include recent changes supporting multiple connections and value specified in the device config

yocto-base/00_flash_and_verify.robot
```diff
@@ -0,0 +1,50 @@
...
+    Telnet.Set Prompt    :~#
+    Telnet.Read Until    login:
+    Telnet.Write Bare    \n
+    Telnet.Login    root    ${DUT_PASSWORD}
+    ${output}=    Telnet.Execute Command    sh -c "cat /etc/os-release | grep VERSION_ID"
+    ${lines}=    Split To Lines    ${output}
+    ${actual_version}=    Evaluate    "${lines[0]}".split("=")[-1]
+    # get version from file name
+    ${gz_name}=    Evaluate    "${FILE_GZ}".split("/")[-1]
+    ${flashed_version}=    Replace String    ${gz_name}    .gz    ${EMPTY}
```
artur-rs, line 49: @WiktorG351 This assumes that the filename always consists of `version.gz`, but what if we have names like `platform-version.gz` or `manufacturer-version.gz`. Can we introduce a naming rule usable in more filename cases? The standard approach would be regexp operations.
@TomaszAIR we should confirm the method with you

yocto-base/00_flash_and_verify.robot
```diff
@@ -0,0 +1,50 @@
...
+        Fail    'Incorrect number of connected SD Wire devices.'
+    END
+    Variable Should Exist    ${DUT_PASSWORD}
+    # flashing
+    Flash SD Card Via SD Wire    ${FILE_BMAP}    ${FILE_GZ}    ${SD_WIRE_SERIAL1}
+    # telnet
+    Telnet.Set Prompt    :~#
+    Telnet.Read Until    login:
+    Telnet.Write Bare    \n
+    Telnet.Login    root    ${DUT_PASSWORD}
```
artur-rs, line 43: @WiktorG351 this seems like a standard keyword we are using in the keywords.robot library. Please check if we can re-use actual keyword.

yocto-base/00_flash_and_verify.robot
WiktorG351: @artur-rs fixed [here](https://github.com/Dasharo/open-source-firmware-validation/pull/22/commits/0639db54fa4d298b32ad5d3106e4e34db1493cc1)

yocto-base/00_flash_and_verify.robot
WiktorG351: @artur-rs unfortunately the only two keywords in `keywords.robot` that use `Telnet.Login` (which we need for this) are `Serial Root Login Linux` and `Serial User Login Linux`, which both have hard-coded prompts.  There is no standard keyword for logging into Telnet.

requirements-openbmc.txt
macpijan: If you use it this way it can stay

yocto-base/00_flash_and_verify.robot
```diff
@@ -0,0 +1,49 @@
...
+
+Flash platform and verify
+    [Documentation]    This test flashes the DUT connected to the RTE through
+    ...    the SD Wire, then attempts to log into it over serial to see whether
+    ...    it works.
+    IF    '${PREV_TEST_STATUS}' == 'FAIL'
+        Fail    'Incorrect number of connected SD Wire devices.'
+    END
+    Variable Should Exist    ${DUT_PASSWORD}
+    # flashing
```
macpijan, line 38: This comment is redundant

yocto-base/00_flash_and_verify.robot
```diff
@@ -0,0 +1,49 @@
...
+    Flash SD Card Via SD Wire    ${FILE_BMAP}    ${FILE_GZ}    ${SD_WIRE_SERIAL1}
+    # telnet
```
macpijan, line 40: as above

yocto-base/00_flash_and_verify.robot
```diff
@@ -0,0 +1,49 @@
...
+    Telnet.Set Prompt    :~#
```
macpijan, line 41: You should not be using Telnet keywords directly if not strictly necessary. See point 3/4 from: https://github.com/Dasharo/open-source-firmware-validation#keywords

yocto-base/00_flash_and_verify.robot
```diff
@@ -0,0 +1,49 @@
...
+    Telnet.Read Until    login:
+    Telnet.Write Bare    \n
+    Telnet.Login    root    ${DUT_PASSWORD}
```
macpijan, line 44: There are existing keywords for logging in to Linux. It is very common operation. Perhaps these can be reused.

yocto-base/00_flash_and_verify.robot
```diff
@@ -0,0 +1,49 @@
...
+    ${output}=    Telnet.Execute Command    sh -c "cat /etc/os-release | grep VERSION_ID"
```
macpijan, line 45: If you already put complex shell expression in, you may just as well put one more and save two lines of robot thanks to that:

```
cat /etc/os-release | grep VERSION_ID | cut -d '=' -f 2
```

macpijan: BTW, checking for version of Linux version sounds like a generic Linux keyword to be defined, so it can be reused.

yocto-base/00_flash_and_verify.robot
```diff
@@ -0,0 +1,49 @@
...
+    ${lines}=    Split To Lines    ${output}
+    ${actual_version}=    Evaluate    "${lines[0]}".split("=")[-1]
+    ${file_gz}=    Evaluate    "${FILE_GZ}".split("/")[-1]
```
macpijan, line 48: I do not understand why do we have to split it. If file_gz contains version, then FILE_GZ as well and should be enough to this check? Can you explain?

yocto-base/00_flash_and_verify.robot
```diff
@@ -0,0 +1,49 @@
...
+    END
+
+Flash platform and verify
+    [Documentation]    This test flashes the DUT connected to the RTE through
+    ...    the SD Wire, then attempts to log into it over serial to see whether
+    ...    it works.
+    IF    '${PREV_TEST_STATUS}' == 'FAIL'
+        Fail    'Incorrect number of connected SD Wire devices.'
+    END
+    Variable Should Exist    ${DUT_PASSWORD}
```
macpijan, line 37: We check for one global variable, but not for others? The FILE_BMAP, and so on, must also exist. Do we check for them somewhere else?

yocto-base/00_flash_and_verify.robot
```diff
@@ -0,0 +1,49 @@
...
+Resource            ../../variables.robot
+Resource            ../../keywords.robot
+Resource            ../lib/sd-wire.robot
+
+Suite Setup         Prepare Test Suite
+Suite Teardown      Log Out And Close Connection
+
+
+*** Test Cases ***
+Verify number of connected SD Wire devices
```
macpijan, line 18: I'm not sure if this is a real test. It does nothing with the DUT itself, it simply checks our config file in variables.robot. Moreover, this is likely to be reused whenever SD Wire is in use.
It looks to me like a keyword to be put in SD wire library, called whenever we need to use SD Wire, or in some SD Wire init keyword. Perhaps as part of the Suite Setup as we do with RTE checking.

platform-configs/rpi-3b.robot
```diff
@@ -0,0 +1,272 @@
...
+${TCG_OPAL_DISK_PASSWORD_SUPPORT}=                  ${FALSE}
+${BIOS_LOCK_SUPPORT}=                               ${FALSE}
+${SMM_WRITE_PROTECTION_SUPPORT}=                    ${FALSE}
+${WIFI_BLUETOOTH_CARD_SWITCH_SUPPORT}=              ${FALSE}
+${CAMERA_SWITCH_SUPPORT}=                           ${FALSE}
+${EARLY_BOOT_DMA_SUPPORT}=                          ${FALSE}
+${UEFI_PASSWORD_SUPPORT}=                           ${FALSE}
+
+# Test module: dasharo-performance
+${SERIAL_BOOT_MEASURE}=                             ${FALSE}
```
macpijan, line 171: Many of these flags does not make sense for Yocto/RPi testing. We need to come up with a way to manage this, not necessarily in this MR.

README.md
```diff
@@ -48,6 +48,7 @@ and running tests have been implemented.
 | Protectli    | VP4650               | Dasharo                  |  `protectli-vp4650`                    |
 | Protectli    | VP4670               | Dasharo                  |  `protectli-vp4670`                    |
 | Raptor-CS    | TalosII              | Dasharo                  |  `raptor-cs_talos2`                    |
+| N/A          | RaspberryPi 3B       | Yocto                    |  `rpi-3b`                              |
```
macpijan, line 51: The vendor can simply by Raspberry Pi, not N/A

macpijan (changes requested):

yocto-base/00_flash_and_verify.robot
macpijan: Why do we care about the naming format? Just check if version string is withing the file name string?

WiktorG351: @macpijan that is precisely what we are doing [here](https://github.com/Dasharo/open-source-firmware-validation/blob/d8b5ae3b5c0183e6bd8be7879181733cf8cfcbd0/yocto-base/00_flash_and_verify.robot#L49)

yocto-base/00_flash_and_verify.robot
WiktorG351: @macpijan adressed [here](https://github.com/Dasharo/open-source-firmware-validation/pull/22/commits/3ee525c2fcb5358836d49cacd883548c2fc6504d#diff-322efff20647f11c797d3f64ba9852d27c8a23969970ad17b19b7b9e7c7d7029L38)

yocto-base/00_flash_and_verify.robot
WiktorG351: @macpijan adressed [here](https://github.com/Dasharo/open-source-firmware-validation/pull/22/commits/3ee525c2fcb5358836d49cacd883548c2fc6504d#diff-322efff20647f11c797d3f64ba9852d27c8a23969970ad17b19b7b9e7c7d7029L40)

yocto-base/00_flash_and_verify.robot
WiktorG351: @macpijan adressed [here](https://github.com/Dasharo/open-source-firmware-validation/pull/22/commits/299c9c0ba1181a5870c8002d6b1b5dd11a1c69ea#diff-322efff20647f11c797d3f64ba9852d27c8a23969970ad17b19b7b9e7c7d7029L39-R39), I reused a keyword that already existed and tested it to make sure it works.

yocto-base/00_flash_and_verify.robot
WiktorG351: @macpijan same as above, adressed [here](https://github.com/Dasharo/open-source-firmware-validation/pull/22/commits/299c9c0ba1181a5870c8002d6b1b5dd11a1c69ea#diff-322efff20647f11c797d3f64ba9852d27c8a23969970ad17b19b7b9e7c7d7029L39-R39)

yocto-base/00_flash_and_verify.robot
WiktorG351: @macpijan adressed [here](https://github.com/Dasharo/open-source-firmware-validation/pull/22/commits/909324f803fe083d61d8fbfc84429dbb1b2224b5#diff-322efff20647f11c797d3f64ba9852d27c8a23969970ad17b19b7b9e7c7d7029L40-R43). I do not understand the second comment. Are You suggesting that I should add a new keyword to `keywords.robot` that checks the version?

yocto-base/00_flash_and_verify.robot
WiktorG351: @macpijan Yes. So basically my idea is that this needs to be done to prevent false positives. For example someone can make a directory with the version name, and then accidentally put a file in it that doesn't correspond to the directory name. Then the test would still pass, because the `Should Contain` keyword at the end of the test would return true, because it found the version name in the full path to the file. In order to avoid this false positive, it is best to extract just the file name, and then check against that.

yocto-base/00_flash_and_verify.robot
WiktorG351: @macpijan no we don't. I have added a check for those two variables in [this commit](https://github.com/Dasharo/open-source-firmware-validation/pull/22/commits/a28991e1148500be410e9f25d60701baf902d741)

README.md
WiktorG351: @macpijan i added the vendor info in [this commit](https://github.com/Dasharo/open-source-firmware-validation/pull/22/commits/fe41929bdd7591ca142da1a439ed7e332afaf33a)

yocto-base/00_flash_and_verify.robot
WiktorG351: @macpijan I adressed this in [this commit](https://github.com/Dasharo/open-source-firmware-validation/pull/22/commits/483a919567615d797ef8a4f37e73c9e8b1449650). This commit:
* moves the `Verify Number Of Connected SD Wire Devices` to the `sd-wire.robot` resource
* changes a keyword in `keywords.robot`, to make it when the config is `rpi-3b` (the thing that uses the `sd-wire.robot` resource), the keyword `Verify Number Of Connected SD Wire Devices` is ran
* also adds a separate `IF ELSE` statement, for when the config is `rpi-3b`

WiktorG351: @macpijan also sorry, in the above comment i forgot to remove a section of code i had commented out. I did that in [this commit](https://github.com/Dasharo/open-source-firmware-validation/pull/22/commits/f0df6a7cb46271996f139bb710e2e975321b800b)

yocto-base/00_flash_and_verify.robot
macpijan: It would be ideal to use Login to Linux, but I see it is not generic enough. Please create issue in this repo linking to this comment that we should use the generic keyword in this test. Then you can close the discussion.

https://github.com/Dasharo/open-source-firmware-validation/blob/main/keywords.robot#L93

yocto-base/00_flash_and_verify.robot
macpijan: Yes. It should be very common activity to do to retreve Linux distro version. So you should create a keyword that just returns Linux version from the OS. 
But we are not adding new kewords in keyword.robot. Please read the guidelines carefully. You should create a new library, e.g. `lib/linux.robot`.

yocto-base/00_flash_and_verify.robot
macpijan: Ok, so the correct way to solve it is to use: https://manpages.ubuntu.com/manpages/trusty/man1/basename.1.html
Or even better, look up for something similar in RF libraries. Maybe in this one: https://robotframework.org/robotframework/latest/libraries/OperatingSystem.html

This is very common activity to extract basename from path, no need to invent it once again.

yocto-base/00_flash_and_verify.robot
macpijan: It is against guidelines to modify/add keywords in keywords.robot. But let's say we can ignore this rule this time.

platform-configs/rpi-3b.robot
macpijan: @WiktorG351 Please at least create issue and link here, so we can close this thread.

yocto-base/00_flash_and_verify.robot
WiktorG351: @macpijan are you sure? it would be a keyword that only has two lines:
```
    ${output}=    Telnet.Execute Command    sh -c "cat /etc/os-release | grep VERSION_ID | cut -d '=' -f 2"
    ${output}=    Get Line    ${output}    0
    [Return]    ${output}
```
and it wouldn't be super generic (aka wouldnt work over ssh, and there would need to be a separate keyword for ssh). Is a new keyword necessary for this?

platform-configs/rpi-3b.robot
WiktorG351: @macpijan yes, [here](https://github.com/Dasharo/open-source-firmware-validation/issues/33) is the new issue

yocto-base/00_flash_and_verify.robot
WiktorG351: @macpijan okay, i used keyword from os library, in [this](https://github.com/Dasharo/open-source-firmware-validation/pull/22/commits/f75318784f351a23bddcf3fea7dce2e612baf6b0) commit

yocto-base/00_flash_and_verify.robot
artur-rs: @WiktorG351 maybe we can determine the correct connection based on the platform-config file? Such as `${DUT_CONNECTION_METHOD}` which we already use in our tests?

Then, based on the value of the variable, you can extend the keyword to handle telnet and ssh command execution, returning the same variable in both cases.

platform-configs/rpi-3b.robot
artur-rs: added 1 comment as well - marking as resolved

yocto-base/00_flash_and_verify.robot
WiktorG351: @artur-rs okay, adressed [here](https://github.com/Dasharo/open-source-firmware-validation/pull/22/commits/4db9447d6aca6f1c3a625514b42a6cdaa9c34d88)

artur-rs (approved):

PR #23, opened 2023-09-27
Selective OpenROM tests

Mixss:

PR #28, opened 2023-09-29
Resizeable BARs test

Mixss:


lib/bios/menus.robot
```diff
@@ -0,0 +1,58 @@
+*** Settings ***
```
miczyg1, line 1: It's a third PR where I see this file being added. It makes the review process to be confusing.

miczyg1:
i guess it depends on changes from #23 , so i will wait with review until #23 is merged

Mixss:
@miczyg1 
> @Mixss any test report I can see from this code?
I guess you mean test logs: https://cloud.3mdeb.com/index.php/s/DDJEYArjXTTtDje

SergiiDmytruk (commented):
If we don't need to check resetting to default value (#23 doesn't do it either), then LGTM.  This also needs a rebase though.

PR #29, opened 2023-09-29
Power State After Power Failure test

filipleple:


lib/power-after-fail-lib.robot
```diff
@@ -0,0 +1,31 @@
...
+    [Arguments]    ${state}
+    ${menu_construction}=    Get Setup SubMenu Construction
+    Select Option From List    ${menu_construction}    Power state after power    ${state}
+
+Simulate Power Failure
+    [Documentation]    This keyword simulates a power failure to the DUT.
+    Sonoff Power Off
+    Sleep    10s
+    Sonoff Power On
+    Sleep    5s
```
miczyg1, line 31: Do not sleep here, otherwise robot may miss some strings if they are parsed right after this keyword. Also there is no need to sleep at all after restoring power.

keywords.robot
```diff
@@ -5,6 +5,7 @@ Library         osfv-scripts/osfv_cli/osfv_cli/snipeit_robot.py
 Resource        keys-and-keywords/flashrom.robot
 Resource        pikvm-rest-api/pikvm_comm.robot
 Resource        lib/secure-boot-lib.robot
+Resource        lib/power-after-fail-lib.robot
```
miczyg1, line 8: Is it needed anywhere else besides `dasharo-compatibility/power-after-fail.robot`? I don't think so. it could be imported directly in `dasharo-compatibility/power-after-fail.robot` only. Note that you create a circular include:

1. `lib/power-after-fail-lib.robot` includes `keywords.robot`
2. `keywords.robot` includes `lib/power-after-fail-lib.robot`

In such case, does `lib/power-after-fail-lib.robot` need to include `keywords.robot`? I thought `keywords.robot` is included in every test suite

dasharo-compatibility/power-after-fail.robot
```diff
@@ -0,0 +1,94 @@
...
+*** Test Cases ***
+PSF001.001 Check Power State After Power Failure default state (firmware)
+    [Documentation]    This test ensures that the option is present, and the
+    ...    default state of this option after flashing is correct.
+    Skip If    not ${TESTS_IN_FIRMWARE_SUPPORT}    PSF001.001 not supported
+    Power On
+    Enter Dasharo System Features
+    Enter Power Management Options
+    ${psf_state}=    Read From Terminal Until    ESC to exit.
+    Should Contain    ${psf_state}    Powered Off
```
miczyg1, line 37: The option default is firmware build dependent. You would have to parse the coreboot config file to check what should be the default state:

CONFIG_MAINBOARD_POWER_FAILURE_STATE

0 - is power off
1 - is power on
2 - is keep previous state (aka. `The state at the moment of power failure`)

dasharo-compatibility/power-after-fail.robot
```diff
@@ -0,0 +1,94 @@
...
+    Skip If    not ${TESTS_IN_FIRMWARE_SUPPORT}    PSF002.001 not supported
+    Power On
+    Enter Dasharo System Features
+    Enter Power Management Options
+    Set Power After Fail State    Powered Off
+    Save Changes And Reset    2    4
+    Simulate Power Failure
+    Sleep    5s
+    ${output}=    Read From Terminal
+    Length Should Be    ${output}    1
```
miczyg1, line 51: Why do we sleep?

We should just wait for the prompt from the firmware using `Read From Terminal Until    <prompt>` or `Enter Setup Menu Tianocore`  and expect failure/timeout (because nothing should be printed as machine should stay powered off).

dasharo-compatibility/power-after-fail.robot
```diff
@@ -0,0 +1,94 @@
...
+
+PSF002.001 Powered Off State Restoration Test
+    [Documentation]    This test ensures that the feature is able to
+    ...    keep the DUT powered off after power failure.
+    Skip If    not ${TESTS_IN_FIRMWARE_SUPPORT}    PSF002.001 not supported
+    Power On
+    Enter Dasharo System Features
+    Enter Power Management Options
+    Set Power After Fail State    Powered Off
+    Save Changes And Reset    2    4
```
miczyg1, line 47:
```suggestion
    Save Changes And Reset    2    4
    Enter Setup Menu Tianocore
```

Let's ensure the platform boots to setup at least and applies the configuration first. Then simulate the power failure.

dasharo-compatibility/power-after-fail.robot
```diff
@@ -0,0 +1,94 @@
...
+    [Documentation]    This test ensures that the feature is able to correctly
+    ...    restore the power state from the moment of failure, in this case to
+    ...    keep the DUT powered off.
+    Skip If    not ${TESTS_IN_FIRMWARE_SUPPORT}    PSF004.001 not supported
+    Power On
+    Enter Dasharo System Features
+    Enter Power Management Options
+    Set Power After Fail State    The state at the moment of power failure
+    Save Changes And Reset    2    4
+    RteCtrl Power Off
```
miczyg1, line 76:
```suggestion
    Enter Setup Menu Tianocore
    RteCtrl Power On
```

Simple power button press should power off every machine supported in Dasharo, as log nas the machine is booted. No need to hold the power button for 5 or 6 seconds (the platform will immediately power off anyways in the first second). but before that, boot up to the setup first so the changes will be applied.

dasharo-compatibility/power-after-fail.robot
```diff
@@ -0,0 +1,94 @@
...
+    Sleep    2s
+    Simulate Power Failure
+    ${output}=    Read From Terminal
+    Length Should Be    ${output}    1
```
miczyg1, line 80: As earlier.

We should just wait for the prompt from the firmware using Read From Terminal Until    <prompt> or Enter Setup Menu Tianocore and expect failure/timeout (because nothing should be printed as machine should stay powered off).

miczyg1 (changes requested):

dasharo-compatibility/power-after-fail.robot
filipleple: I think the simplest option would be adding that as a setting for the compatibility module to the mainboard config? IMO we don't know if we'll be able to readily access the build config file from the position of the test file

dasharo-compatibility/power-after-fail.robot
filipleple: Fixed in [recent commit](https://github.com/Dasharo/open-source-firmware-validation/commit/e84e768780bae0482929f7309386c111a401257c)

dasharo-compatibility/power-after-fail.robot
filipleple: Fixed in [recent commit](https://github.com/Dasharo/open-source-firmware-validation/commit/e84e768780bae0482929f7309386c111a401257c)

dasharo-compatibility/power-after-fail.robot
filipleple: Fixed in [recent commit](https://github.com/Dasharo/open-source-firmware-validation/commit/e84e768780bae0482929f7309386c111a401257c)

dasharo-compatibility/power-after-fail.robot
filipleple: Fixed in [recent commit](https://github.com/Dasharo/open-source-firmware-validation/commit/e84e768780bae0482929f7309386c111a401257c)

keywords.robot
filipleple: Fixed in [recent commit](https://github.com/Dasharo/open-source-firmware-validation/commit/e84e768780bae0482929f7309386c111a401257c)

lib/power-after-fail-lib.robot
filipleple: Fixed in [recent commit](https://github.com/Dasharo/open-source-firmware-validation/commit/e84e768780bae0482929f7309386c111a401257c)

dasharo-compatibility/power-after-fail.robot
miczyg1: I gave a generic method that will always work with any binary and it is the most recommended option. Let;s not hardcode things which do not have to be hardcoded.

miczyg1:
@filipleple please do not resolve the threads yourselves, because I lose the track what has been addressed and what not. Also if you force-push the commits and always point to a single commit, it defeats the purpose of linking it ( I still have to scroll and look for the line where the change has been introduced)... At least please link the line number

lib/power-after-fail-lib.robot
miczyg1: @filipleple my comment appplied only to the line 31: `Sleep    5s`. We need to give a couple of seconds before restoring the power so `Sleep    10s` between sonoff calls is desired and should be brought back.

dasharo-compatibility/power-after-fail.robot
```diff
@@ -0,0 +1,97 @@
...
+    ...    restore the power state from the moment of failure, in this case to
+    ...    keep the DUT powered off.
+    Skip If    not ${TESTS_IN_FIRMWARE_SUPPORT}    PSF004.001 not supported
+    Power On
+    Enter Dasharo System Features
+    Enter Power Management Options
+    Set Power After Fail State    The state at the moment of power failure
+    Save Changes And Reset    2    4
+    Enter Setup Menu Tianocore
+    Power On
```
miczyg1, line 79:
```suggestion
```

No need for Power On since we are powered on if entered the setup

dasharo-compatibility/power-after-fail.robot
macpijan: @miczyg1 @filipleple I am fine with defining this default in platform config, just as we do for the default Secure Boot state as far as I remember.

It is true that we could extract such things from coreboot config and expect based on that, but I'm not totally convinced on this approach.

One of the goal of testing is to also define an independent set of expected outcomes and verify that the firmware product delivers that. If we define in the platform config that platform X is expected to ship with feature Y enabled by default, we can check if the firmware developer used correct configuration as well.

dasharo-compatibility/power-after-fail.robot
filipleple: Fixed [here](https://github.com/Dasharo/open-source-firmware-validation/pull/29/files#diff-b9e98ddabefa0c5dd70c8174e598bd616df7ae602633134a00d0debe605e2398R79)

dasharo-compatibility/power-after-fail.robot
filipleple: So, config [option](https://github.com/Dasharo/open-source-firmware-validation/pull/29/files#diff-924a28816186a815414d0e7ac57b47fab4603000ee13fbd0adb8a4952e0b7b3dR160) added and checked for [in the test case](https://github.com/Dasharo/open-source-firmware-validation/pull/29/files#diff-b9e98ddabefa0c5dd70c8174e598bd616df7ae602633134a00d0debe605e2398R38)

lib/power-after-fail-lib.robot
filipleple: OK the 10s sleep [line](https://github.com/Dasharo/open-source-firmware-validation/commit/4e0c68c288a75411d2cdfa55e8cc1ba38cfaa79a#diff-ee78c06a5fbccca05b7cef40ff817bff0d7d9e5526ddf1f2127b69c97d91ac49R27) is brought back

dasharo-compatibility/power-after-fail.robot
macpijan: @filipleple It shall be added to all MSI boards, at least.

filipleple: @macpijan added to all platform configs

miczyg1: > One of the goal of testing is to also define an independent set of expected outcomes and verify that the firmware product delivers that. If we define in the platform config that platform X is expected to ship with feature Y enabled by default, we can check if the firmware developer used correct configuration as well.

Then why not have both? The hardcoded EXPECTED state of feature and then the feature state in the binary. Then we could also detect misconfiguration of the binary. Maybe even do a preliminary test to check feature state before starting the real tests

lib/power-after-fail-lib.robot
```diff
@@ -0,0 +1,28 @@
...
+    Press Key N Times And Enter    ${system_index}    ${ARROW_DOWN}
+
+Set Power After Fail State
+    [Documentation]    Sets the Power State After Power Fail option to a
+    ...    desired state declared in ${state}. Available options are:
+    ...    Powered Off
+    ...    Powered On
+    ...    The state at the moment of failure
+    [Arguments]    ${state}
+    ${menu_construction}=    Get Setup SubMenu Construction
```
macpijan, line 21: No need for dedicated keyword just for that. The `Set Option Value` can be directly used in tests like here: https://github.com/Dasharo/open-source-firmware-validation/blob/rework-terminal-kwds/dasharo-security/me-neuter.robot#L50

lib/power-after-fail-lib.robot
```diff
@@ -0,0 +1,28 @@
...
+    Select Option From List    ${menu_construction}    Power state after power    ${state}
+
+Simulate Power Failure
+    [Documentation]    This keyword simulates a power failure to the DUT.
+    Sonoff Power Off
```
macpijan, line 26: What if there is no sonoff, just RTE? It must be more generic.

lib/power-after-fail-lib.robot
```diff
@@ -0,0 +1,28 @@
+*** Settings ***
+Documentation       Collection of keywords related to the Power State After
+...                 Power Fail option
+
+
+*** Keywords ***
+Enter Power Management Options
```
macpijan, line 7: We do not need this kwd. Simply use `Enter Dasharo Submenu` as here: https://github.com/Dasharo/open-source-firmware-validation/blob/rework-terminal-kwds/dasharo-security/me-neuter.robot#L49

macpijan (changes requested):
Please try to rebase to https://github.com/Dasharo/open-source-firmware-validation/pull/55 and apply suggested changes to use the reworked keywords.

dasharo-compatibility/power-after-fail.robot
macpijan: Issue for this enhancement: https://github.com/Dasharo/open-source-firmware-validation/issues/82

macpijan:
Replaced by: https://github.com/Dasharo/open-source-firmware-validation/pull/83

PR #45, opened 2023-10-11
Add test: Auto boot time-out

Psotas:


dasharo-compatibility/boot-maintenance-manager.robot
```diff
@@ -0,0 +1,73 @@
...
+    Reset To Defaults Tianocore    checkpoint=${DASHARO_EXIT_PROMPT}
+    Press Key N Times    1    ${F10}
+    Write Bare Into Terminal    y
+    Read From Terminal Until    ${DASHARO_EXIT_PROMPT}
+    Restart The DUT
+    Enter Setup Menu Tianocore
+    Enter Dasharo Submenu    Boot Maintenance Manager
+    Refresh Serial Screen In BIOS Editable Settings Menu
+    ${value}=    Get Option Value    Auto Boot Time-out    checkpoint=${DASHARO_EXIT_PROMPT}
+    Should Be Equal    ${value}    ${DEFAULTS_VALUE_AUTO_BOOT_TIME_OUT}
```
miczyg1, line 66:
```suggestion
    Should Be Equal    ${value}    ${AUTO_BOOT_TIME_OUT_DEFAULT_VALUE}
```

Please apply to all instances of the variable.

miczyg1 (changes requested):

dasharo-compatibility/boot-maintenance-manager.robot
Psotas: Done

Psotas:
@miczyg1 mam poprawić błędy znalezione przez `Smoke tests on selected platforms / build (pull_request)` czy potem zrobisz force moich zmian?

platform-configs/protectli-vp2420.robot
```diff
@@ -42,6 +42,9 @@ ${USB_SHORT_NAME}=                                  ${EMPTY}
 ${ME_INTERFACE}=                                    Intel Corporation Device 4b70 (rev 11)
 ${INITIAL_FAN_RPM}=                                 ${EMPTY}
 ${ACCEPTED_%_NEAR_INITIAL_RPM}=                     ${EMPTY}
+${AUTO_BOOT_TIME_OUT_DEFAULT_VALUE}=                6
+${DASHARO_ENTER_PROMPT}=                            <Enter>=Select Entry
```
macpijan, line 46: @Psotas I really do not see why need these to be customizable. Can you give 2 platforms that have these strings set differently? We use the same payload everywhere.

You open an issue on problem with variables: https://github.com/Dasharo/dasharo-issues/issues/519 that are constantly missing in platform configs, while you introduce here another set of new variables, and to not add them to all configs, rendering new instances of this problem.

lib/bios/menus.robot
```diff
@@ -59,6 +59,25 @@ Enter Dasharo System Features Submenu Snapshot
     ${menu_construction}=    Get Setup Submenu Construction
     RETURN    ${menu_construction}
 
+Enter Dasharo Submenu
```
macpijan, line 62: @Psotas We cannot accept adding new keywords that provide redundant functionality anymore. This is the exact approach that was used here in this repo, that led to the current state that we have 8 keywords for logging into operating system...

If you opt for improvement of the keyword, you shall improve/replace the existing one, and all of it's occurrences. If this is too much, create an issue describing your proposal, but in these MR you should use existing ones.

macpijan:
@Psotas Please use English.

Let me take a look at the SnipeIT error.

macpijan:
@Psotas Please rebase on top of the latest master. The snipeit import shall be fixed in: https://github.com/Dasharo/open-source-firmware-validation/commit/6787a7455de5f7200c22ab45a9038f6dc289a864

platform-configs/protectli-vp2420.robot
macpijan: @Psotas I can see you are keep adding these new variables to configs. But I am still not convinced on the necessity of their existence. Can you please respond to the above first?

Psotas: I misunderstood what the tests I was based on were waiting for (reset-to-defaults.robot).
Not every menu (Boot Mainteance Manager does not have one) contains `ESC to exit`.
I'm already removing unnecessary variables.

image

lib/bios/menus.robot
Psotas: OK, I will try to improve the keywords.
I was planning to do it in a separate PR, now I know to do it in the same one.

dasharo-compatibility/boot-maintenance-manager.robot
```diff
@@ -0,0 +1,73 @@
...
+    Reset To Defaults Tianocore    Esc=Exit
+    Press Key N Times    1    ${F10}
+    Write Bare Into Terminal    y
+    Read From Terminal Until    Esc=Exit
+    Restart The DUT
+    Enter Setup Menu Tianocore
+    Enter Dasharo Submenu    Boot Maintenance Manager
+    Refresh Serial Screen In BIOS Editable Settings Menu
+    ${value}=    Get Option Value    Auto Boot Time-out    Esc=Exit
+    Should Be Equal    ${value}    ${AUTO_BOOT_TIME_OUT_DEFAULT_VALUE}
```
macpijan, line 66: @Psotas I wonder how this could have worked, if the `[6]` is being compared with `6` from config. In the above test you specifically compared it with `[7]`. So I just suspect it has not worked.

dasharo-compatibility/boot-maintenance-manager.robot
```diff
@@ -0,0 +1,73 @@
...
+    Change Numeric Value Of Setting    Auto Boot Time-out    7    Esc=Exit
+    Press Key N Times    1    ${F10}
+    Write Bare Into Terminal    y
+    Read From Terminal Until    Esc=Exit
+    Restart The DUT
+    Enter Setup Menu Tianocore
+    Enter Dasharo Submenu    Boot Maintenance Manager
+    Refresh Serial Screen In BIOS Editable Settings Menu
+    ${value}=    Get Option Value    Auto Boot Time-out    Esc=Exit
+    Should Be Equal    ${value}    [7]
```
miczyg1, line 46:
```suggestion
    Should Be Equal    ${value}    7
```

Shouldn't the BIOS menu keywords strip the square brackets and extract the value of option?

dasharo-compatibility/boot-maintenance-manager.robot
miczyg1: The first test aims to change the value to an arbitrary one and compare if the change was applied.

Second test check if timeout resets back to the default, which is hardcoded in config.

I don't see any error with that approach @macpijan except the square brackets which should be rather stripped by BIOS menu keywords

miczyg1 (changes requested):

dasharo-compatibility/boot-maintenance-manager.robot
macpijan: The test with improved keywords is moved here: https://github.com/Dasharo/open-source-firmware-validation/pull/55/files#diff-3c157e9d4762250f95224f014fa3acb1f979e1aca0c9fe3808a4156f980d7cc7R28

macpijan: > I don't see any error with that approach @macpijan except the square brackets which should be rather stripped by BIOS menu keywords 

I was just only saying that it couldn't have worked, since one test case assume they are stripped, the other assumes otherwise. And @Psotas claimed the both worked.

macpijan:
This has been integrated into: https://github.com/Dasharo/open-source-firmware-validation/pull/5

dasharo-compatibility/boot-maintenance-manager.robot
miczyg1: > Please do not waste time on this MR. 

Unfortunately this won't always work in my workmodel. I go through each of the notifications in order. So, depending on what notifications I go through first (whether mail or matrix), I may, or may not, notice such announcements.

macpijan: If it would be closed (as I asked) you would not have reviewed close MR I assume.

PR #52, opened 2023-10-12
Qemu improvements

macpijan:


README.md
```diff
@@ -49,6 +46,7 @@ and running tests have been implemented.
 | Protectli    | VP4670               | Dasharo                  |  `protectli-vp4670`                    |
 | Raptor-CS    | TalosII              | Dasharo                  |  `raptor-cs_talos2`                    |
 | Raspberry Pi | RaspberryPi 3B       | Yocto                    |  `rpi-3b`                              |
+| Emulation    | QEMU x86             | Dasharo (OVMF)           |  `qemu`                                |
```
pietrushnic, line 49: @macpijan, for some reason, only QEMU has the framework payload naming scheme (`Dasharo (framework+payload)`). I would appreciate the consistency of this table with the [DCS Roadmap](https://github.com/Dasharo/presentations/blob/main/dasharo_roadmap.md#dasharo-emulation-roadmap)
as well as naming, which we introduce in [Dasharo Universe](https://docs.dasharo.com/variants/overview/).

README.md
```diff
@@ -173,6 +171,40 @@ Parameters should be defined as follows:
...
+options enabled or disabled). In such a case, please refer to:
+* [Building Manual in Dasharo for QEMU documentation](https://docs.dasharo.com/variants/qemu_q35/building-manual/)
+* [Development section in Dasharo for QEMU documentation](https://docs.dasharo.com/variants/qemu_q35/development/)
+
+Following tests are proven to work with QEMU right now:
+
+```shell
+robot -L TRACE -v config:qemu -v rte_ip:127.0.0.1 -v snipeit:no dasharo-compatibility/uefi-shell.robot
+robot -L TRACE -v config:qemu -v rte_ip:127.0.0.1 -v snipeit:no dasharo-compatibility/custom-boot-menu-key.robot
+```
```
pietrushnic, line 203: The assumption here is that someone has correctly configured the environment. Maybe a link would be needed to the above section if one would be linked directly here?

scripts/ci/qemu-run.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+
+if [ ! -f "./OVMF_CODE.fd" ]; then
+    wget -O ./OVMF_CODE.fd https://cloud.3mdeb.com/index.php/s/iXmyJxKLBB2Eb94/download
+fi
+
+if [ ! -f "./OVMF_VARS.fd" ]; then
+    wget -O ./OVMF_VARS.fd https://cloud.3mdeb.com/index.php/s/Mn2n92DZLnZQZ8N/download
+fi
```
pietrushnic, line 9: @macpijan I guess this is some release candidate, but in the future (with the v0.1.0 release), we plan to put that something going directly from dasharo.com.

pietrushnic (changes requested):

scripts/ci/qemu-run.sh
macpijan: @pietrushnic Release v0.1.0 was apparently a couple of months ago, but no binaries wre published: https://docs.dasharo.com/variants/qemu_q35/releases/

Now that we have CI enabled, maybe I will just integrate one more GitHub action to publish github release on tag and link this instead?

README.md
macpijan: 2346889fbe5b87eb0a4d184cef4b0dc3fd6a035e

README.md
macpijan: 2346889fbe5b87eb0a4d184cef4b0dc3fd6a035e

README.md
macpijan: @pietrushnic I just wanted to indicate the difference, that no coreboot is used here, and OvmfPkg (not UefiPayloadPkg) from edk2 is used here. How do you suggest to indicate that?

pietrushnic: @macpijan I'm fine with Dasharo (UEFI). The fact the release is for QEMU Q35 means it has to be OVMF because there is no other way to support EDK II (also incorrectly known as UEFI).

Please check the rules: https://docs.dasharo.com/dev-proc/smbios-rules/#bios-version-field

My comment is not only about Dasharo but also about the fact of how that table is designed. IMHO, it is incorrect and consistent with how we did it in Dasharo Universe. I don't know if we can resolve everything now, but the naming scheme used on Dasharo Universe suggests QEMU should be at the same level as Protectli/Novacustom, and in the platform, you should have Q35 because that machine (in QEMU terminology platform) is emulated and validated.

pietrushnic: Fixed that line in 38a35a1f4d46

macpijan:
This is already part of the: https://github.com/Dasharo/open-source-firmware-validation/pull/71

PR #63, opened 2023-10-23
Qemu q35 v0.1.0 rc1

pietrushnic:

PR #70, opened 2023-10-25
Rework terminal kwds + qemu network boot

macpijan:
This PR replaces: https://github.com/Dasharo/open-source-firmware-validation/pull/63


macpijan:
I propose to merge it, so we do not diverge too much. Changes introduce here do not break anything on the target branch.

We will still need to work on keywords rework in the tests, but it does need to happen in this MR.

PR #71, opened 2023-10-25
[WIP] Rework terminal kwds qemu sb

macpijan:
- SB tests added to self-tests/setup-and-boot-menus.robot
  and should work
- The SBO001.001 should work in QEMU
- The SBO002.001 must be worked on - currently, the keys are correctly
  reset when needed (so option to enable SB becomes available), but
  this option does not work in QEMU (Attempt Secure Boot never becomes
  accessible, which must be fixed first)
- Then, we should continue with reworking the rest of the keywords
  in lib/secure-boot.robot
- Another problem is this suite was prepared for PiKVM only - we need
  to find another way for provifing USB disk images with signed/unsigned
  EFI files for testing

### DIFF
--- a/dasharo-security/secure-boot.robot
+++ b/dasharo-security/secure-boot.robot
@@ -41,14 +41,11 @@ SBO001.001 Check Secure Boot default state (firmware)
     ${device_mgr_menu}=    Enter Submenu From Snapshot And Return Construction
     ...    ${setup_menu}
     ...    Device Manager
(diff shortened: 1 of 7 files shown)


macpijan:
The target branch has been merged already.
An updated version of this MR, including more changes and adjustments to the recent SB layout: https://github.com/Dasharo/open-source-firmware-validation/pull/95

PR #96, opened 2023-11-09
lib/bios/menus.robot: improve reliability of parsing menu into constr…

macpijan:
…uction

We have added F9/F10 hotkeys to be available globally, which breaked some of the kwds. With the following change, we can properly parse both menu styles (before and after this change).


macpijan:
Menus parsing work both in CI (older menus layout) and from this MR: https://github.com/Dasharo/edk2/pull/109 (new menu layout with F9/F10 appearing in all menu screens).

PR #100, opened 2023-11-10
lib/power-after-fail-lib.robot: wait 15 seconds before power on

SergiiDmytruk:
Use 15 seconds delay because if power is absent for less than roughly 10 seconds the platform powers on regardless of the settings (original firmware behaves the same) and waiting 10 seconds doesn't produce stable results.

Fixes https://github.com/Dasharo/dasharo-issues/issues/557.

### DIFF
--- a/lib/dl-cache.robot
+++ b/lib/dl-cache.robot
@@ -23,9 +23,7 @@ Download To Host Cache
     END
     Log    Downloading ${url} ...
     ${wget_rc}=    Run And Return RC    wget -O ${local_path} ${url}
-    IF    ${wget_rc} != 0
-        Fail    Download failed with exit code: ${wget_rc}
-    END
+    IF    ${wget_rc} != 0    Fail    Download failed with exit code: ${wget_rc}
 
 Calculate SHA256 Sum
     [Arguments]    ${file_path}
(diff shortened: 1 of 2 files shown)


SergiiDmytruk:
`robotidy` check fails with such worktree changes:
```
diff --git a/lib/dl-cache.robot b/lib/dl-cache.robot
index b8a99cb7..651e546c 100644
--- a/lib/dl-cache.robot
+++ b/lib/dl-cache.robot
@@ -23,9 +23,7 @@ Download To Host Cache
     END
     Log    Downloading ${url} ...
     ${wget_rc}=    Run And Return RC    wget -O ${local_path} ${url}
-    IF    ${wget_rc} != 0
-        Fail    Download failed with exit code: ${wget_rc}
-    END
+    IF    ${wget_rc} != 0    Fail    Download failed with exit code: ${wget_rc}

 Calculate SHA256 Sum
     [Arguments]    ${file_path}

```

Don't know if that's intended, the syntax is unreadable even on multiple lines, let alone on a single one.

macpijan:
Thanks for getting down to the bottom of the problem.

I went ahead and applied robotidy suggestion.

macpijan (approved):

PR #102, opened 2023-11-12
openqa-hw-tests: add configuration scripts

SergiiDmytruk:
For the lack of better place, sending openQA-related scripts here.


openqa-hw-tests/openqa-qubesos-setup.png
pietrushnic, line 1: @SergiiDmytruk, nice diagram. What tool do you use to create it? Do you have the source for that diagram?

openqa-hw-tests/README.md
```diff
@@ -0,0 +1,81 @@
+# Configuration for testing of Qubes OS via PiKVM and RTE
```
pietrushnic, line 1: @SergiiDmytruk aren't break DRY? I would expect all that stuff was already made by the Qubes OS Team and if we testing Qubes maybe we should consider upstream those procedures.

pietrushnic (changes requested):

openqa-hw-tests/openqa-qubesos-setup.png
SergiiDmytruk: It was made by [ditaa](https://github.com/stathissideris/ditaa) source is [here](https://github.com/3mdeb/news-and-ideas/pull/547/files#diff-8f767932e18bd14210866ef9087fa2755db2c88c20cecc7a3d48c26bb6cde9b6R97-R131).

openqa-hw-tests/README.md
SergiiDmytruk: We're reusing part of what was done by Marek, but scripts and configuration are specific in each testing setup.  Marek runs worker on PiKVM where he installed openSUSE, he doesn't use it for anything else but such tests, there is no Sonoff or RTE, etc.

openqa-hw-tests/openqa-qubesos-setup.png
pietrushnic: Thanks, diagram looks neat.

openqa-hw-tests/README.md
pietrushnic: @SergiiDmytruk, I think it needs better justification. I read this:
https://github.com/3mdeb/news-and-ideas/pull/547/files#diff-8f767932e18bd14210866ef9087fa2755db2c88c20cecc7a3d48c26bb6cde9b6R22

Even after reading the blog post and your answer here, it is unclear to me why we created something different than Marek. I roughly understand it is related to different hardware setups, which depends on our unique needs as open-source firmware developers, but why can't we change/improve our or Marek's approach to cover both situations? We should always aim for a unified approach, especially if we deal with partner projects like Qubes OS. Of course, there still may be strong justification for why we are doing things differently and why we have to diverge from the original approach, but it would require Marek's review as well as a very good justification from our side that the community and ourselves can understand in 3-5 years.

SergiiDmytruk: > why can't we change/improve our or Marek's approach to cover both situations?

Due to setup differences it will be quite complicated and make https://github.com/QubesOS/tests-hw-setup unmaintainable and probably fragile.  Knowledge of upstream setup is embedded into scripts and tests (to some degree) along with hardware-specific workarounds.

I'm not saying that it can't be done, it just seems unfeasible.  It reminds me of how Doxygen is implemented: to obey DRY (maybe for some other reason), 10+ of quite different languages are parsed by the same code cost of change of which is sky high.  I saw the same in SrcML project, where I wanted but couldn't fix a bug because they combined various grammars into one for simplicity and created an unmanageable monster (I think they have funding issues because of that).  In some cases DRY is much better to avoid as it creates problems of its own.

If original configuration was designed to be customizable, things would be simpler, but as it currently stands Marek's configuration likely needs quite an overhaul to allow supporting third-party setups.  And that setup is barely documented.

pietrushnic: @SergiiDmytruk I agree with your justification, but it should be documented, and we should really ask @marmarek what he thinks about that.

>  In some cases DRY is much better to avoid as it creates problems of its own.

Argee, DRY at all costs doesn't make sense., but neglecting DRY without sufficient explanation sooner or later has to raise questions.

openqa-hw-tests/power
```diff
@@ -0,0 +1,82 @@
...
+function rte_cmd() {
+    sshpass -pmeta-rte \
+            ssh -q -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null "root@$rte_ip" \
+                "$@"
+}
+
+function setup_video() {
+    if pikvm_cmd v4l2-ctl --get-dv-timings | grep -q 'Active width: 0'; then
+        echo "Setting EDID"
+        pikvm_cmd v4l2-ctl --set-edid 'file=$HOME/openqa/edid.1024x768'
```
marmarek, line 22: This needs to be done once after boot, I have a systemd unit for it: https://github.com/QubesOS/tests-hw-setup/blob/5de30dfe12d0e9399b1d085fda738c1b0fc992b5/states/openqa-worker.sls#L273-L288

openqa-hw-tests/README.md
marmarek: Generally, I'm open accepting PRs that will make https://github.com/QubesOS/tests-hw-setup more flexible.  One part that I already have, is `/etc/openqa/hw-control.conf` with device specific parameters (for example how to press power button or access serial port, which varies between systems a lot). Example for MSI one:
```
[system_state]
# GPIO pin + value
power = led 24 1

[disk]
bootpart = /dev/nvme0n1p2
efipart = /dev/nvme0n1p1
lvmpart = /dev/nvme0n1p3
disk = /dev/nvme0n1

[serial]
console = ttyAMA1

[buttons]
# those are GPIO pin numbers + values
power = switch 23 0 1
reset = switch 27 0 1

[console]
serial = ttyAMA1
```

But indeed it seems the changes required to adjust for a very different architecture (running openqa-worker on a separate device and controlling PiKVM via API) will require quite a bit of changes. In my setup, the function of PiKVM API serves a `testbed-control` script called via SSH, so a very different API. That means most of the `GENERAL_HW_*_CMD` scripts will be different. At this point, there are very few common parts I think. Maybe `gadget-control` script - if you really want to avoid duplication, maybe you can use git submodule.

Note that relatively small part of the tests-hw-setup repo is about openQA integration itself. Most is about setting up RPi system and various auxiliary tools via salt (script to press power button via servo motor, hostapd, gitlab-runner for Xen tests, RPi-specific kernel). I don't see any of those in here and most is likely out of scope (you have it either covered by PiKVM software, RTE, or do the setup steps manually).

BTW I have considered using more of the PiKVM software stack, especially the video part, but VNC encoding is not compatible with openQA, so accessing /dev/video0 directly was easier. At this point, having openqa-worker on RPi (instead of RPi + some extra system) seems easier at the scale, but adding openqa-worker to PiKVM system was much more complex than adding (parts of) PiKVM software to openSUSE with openqa-worker. So here we are.

openqa-hw-tests/power
SergiiDmytruk: Thanks, but having maximum resolution on boot may be useful to run some manual tests, I didn't want to fix it to 1024x768.

SergiiDmytruk:
Closing in favor of [a branch in TrenchBoot/openqa-tests-qubesos](https://github.com/TrenchBoot/openqa-tests-qubesos/tree/3mdeb-lab/generalhw).  The scripts are used by this version of the test suite, so keeping them together makes more sense.

PR #106, opened 2023-11-14
dasharo-compatibility/reset-to-defaults.robot: add test case

filipleple:


dasharo-compatibility/reset-to-defaults.robot
```diff
@@ -272,3 +272,25 @@ RTD014.001 F9 resets Memory SPD Profile to JEDEC
     ${memory_menu}=    Enter Dasharo Submenu    ${dasharo_menu}    Memory Configuration
     ${memory_profile}=    Get Option State    ${memory_menu}    Memory SPD Profile
     Should Be Equal    ${memory_profile}    JEDEC (safe
+
+RTD015.001 F9 reset is globally effective
```
macpijan, line 276: This test checks is it is effective in the Dasharo System Features, whuch is what we need, but not globally effective.
We need another test case like:
- changing stuff in DSF menu
- entering into device manager, or boot maintenance manager,
- resetting there,
- checking options state in DSF

Ideally, also the other way around, but at least the one scenario above should be there.

filipleple: Two more test cases [added](https://github.com/Dasharo/open-source-firmware-validation/pull/106/commits/1b6da908c045c632b945cd6b7c5f519580404ae4#diff-a97bf45e762edcf01a621e3d9a07b4ea56a09bd141df894f76f0f0cf46ff5804R298)

macpijan (approved):

dasharo-compatibility/reset-to-defaults.robot
macpijan: thanks

PR #116, opened 2023-11-16
Qemu install os

macpijan:
Tested with Ubuntu 22.04.03 installer, manual installation.

OvmgPkg from: https://github.com/Dasharo/edk2/pull/111

Was able to boot with SB enabled.

Without that, installer could not boot at all and/or performed very poorly.

### DIFF
--- a/scripts/ci/qemu-run.sh
+++ b/scripts/ci/qemu-run.sh
@@ -2,12 +2,8 @@
 
 INIT_DIR="$(pwd)"
 
-# Check if a directory was provided as the first parameter
-if [ -n "$1" ]; then
-    DIR="$1"
-else
-    DIR="$(pwd)"
-fi
+# Optionally, accept DIR as environmental variable. If not given, use current directory.
+DIR="${DIR:-$PWD}"
 
 # Check if the directory exists
 if [ ! -d "$DIR" ]; then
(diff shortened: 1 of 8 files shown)


scripts/ci/qemu-run.sh
```diff
@@ -52,8 +64,12 @@ case "$2" in
 	vnc)
 		qemu-system-x86_64 $q35_params -nographic -vnc :0
 		;;
+	install_ubuntu)
+		qemu-system-x86_64 $q35_params -cdrom qemu-data/ubuntu.iso
+		;;
 	*)
-    qemu-system-x86_64 $q35_params
+    qemu-system-x86_64 $q35_params -display gtk,window-close=off -vga virtio
```
pietrushnic, line 71: @macpijan this conflicts with vnc feature, I'm not running QEMU locally but on remote machine and would like to vnc to it.

scripts/ci/qemu-run.sh
```diff
@@ -52,8 +64,12 @@ case "$2" in
 	vnc)
 		qemu-system-x86_64 $q35_params -nographic -vnc :0
 		;;
+	install_ubuntu)
+		qemu-system-x86_64 $q35_params -cdrom qemu-data/ubuntu.iso
```
pietrushnic, line 68: Where I should get `qemu-data/ubuntu.iso` ? When I'm running this option I get  

```
qemu-system-x86_64: -drive file=qemu-data/hdd.qcow2,if=virtio: Could not open 'qemu-data/hdd.qcow2': No such file or directory
```

pietrushnic (changes requested):
@macpijan our wrapper to QEMU very quickly growing and I think it would be unmanagable soon. We should think about more scalable way of providing parameters to QEMU, maybe we should leverage QMP as you did in other place?

scripts/ci/qemu-run.sh
macpijan: I simply used 22.04.3 standard desktop installer and installed manually.
I wanted to use autoinstaller from https://github.com/Dasharo/preseeds/blob/main/ubuntu/create_image.sh but it seems it has worked for previous ISO version only. I wanted to focus on SB testing, not on OS installation testing here, so moved this task for later.

scripts/ci/qemu-run.sh
macpijan: But vnc is another command, so it would not be used. This gives proper resolution and fullscreen when running graphical interface locally. Not sure if all parameters are needed TBH. We may still need to work on this, I took some known working configuration and based on that to be able to install OS.

macpijan:
>  our wrapper to QEMU very quickly growing and I think it would be unmanagable soon. We should think about more scalable way of providing parameters to QEMU

I'm afraid of that too...

> maybe we should leverage QMP as you did in other place?

You mean the KW liibrary https://github.com/Dasharo/open-source-firmware-validation/blob/main/lib/QemuMonitor.py ?
We use some of it already, e.g. in secure boo tests to attach different USB devices with test files.

scripts/ci/qemu-run.sh
macpijan: @pietrushnic Can you elaborate here?

pietrushnic:
> You mean the KW liibrary https://github.com/Dasharo/open-source-firmware-validation/blob/main/lib/QemuMonitor.py ?

Yes.

scripts/ci/qemu-run.sh
macpijan: @pietrushnic But you used the `vnc` branch of this switch, not `*`, or not?

macpijan:
@pietrushnic Proposed some improvements on how to handle different modes of operation.

I have not immediate idea how to manage this better.

scripts/ci/qemu-run.sh
macpijan: I reworked the script a bit more

pietrushnic (approved):

pietrushnic:
@macpijan I approved changes

PR #144, opened 2023-12-11
Intermediate cert test

WiktorG351:
This PR adds a test which tests the enrolling of an intermediate key.


dasharo-security/secure-boot.robot
```diff
@@ -233,6 +233,29 @@ SBO008.001 Attempt to enroll the key in the incorrect format (firmware)
     Select File In File Explorer    DB.txt
     Read From Terminal Until    ERROR: Unsupported file type!
 
+SBO009.001 Attempt to boot file signed for intermediate certificate
+    [Documentation]    This test verifies that a file signed with an
```
TomaszAIR, line 237: @WiktorG351 Could you elaborate on the description here about how you generated the chain of certs and then how signed the file?

TomaszAIR: I want to be sure that we are really testing what is stated in the name of test.

TomaszAIR: Ok, here we create intermediate certificate by signing csr using PK priv key.

TomaszAIR (approved):

PR #161, opened 2023-12-19
dasharo-security/secure-boot.robot: Add test for key enrollment

m-iwanicki:
and file signing from inside operating system using sbctl

### DIFF
--- a/dasharo-security/secure-boot.robot
+++ b/dasharo-security/secure-boot.robot
@@ -409,6 +409,48 @@ SBO011.001 Attempt to enroll expired certificate and boot signed image
     ${out}=    Execute File In UEFI Shell    hello-signed-with-expired-cert.efi
     Should Contain    ${out}    Access Denied
 
+SBO012.001 Boot OS Signed And Enrolled From Inside System (Ubuntu 22.04)
+    [Documentation]    This test verifies that OS boots after enrolling keys
+    ...    and signing system from inside
+    Skip If    not ${SECURE_BOOT_SUPPORT}    SBO009.001 not supported
+    Skip If    not ${TESTS_IN_FIRMWARE_SUPPORT}    SBO009.001 not supported
+    Skip If    not ${TESTS_IN_UBUNTU_SUPPORT}    SBO009.001 not supported
+
+    Power On
+    # 1. Make sure we are in Setup Mode
+    ${sb_menu}=    Enter Secure Boot Menu And Return Construction
+    ${advanced_menu}=    Enter Advanced Secure Boot Keys Management And Return Construction    ${sb_menu}
+    Erase All Secure Boot Keys    ${advanced_menu}
+    Exit From Current Menu
+    Save Changes And Reset    2
+
+    Boot System Or From Connected Disk    ubuntu
+    Login To Linux
+    Switch To Root User
+
+    # 2. Remove old keys then generate and enroll new keys
+    Remove Old Secure Boot Keys
+    ${sb_status}=    Generate Secure Boot Keys
+    Should Be True    ${sb_status}
+    ${sb_status}=    Enroll Secure Boot Keys
+    Should Be True    ${sb_status}
+
+    # 3. Sign all components
+    Sign All Boot Components
+    Execute Reboot Command
+
+    # 4. Enable Secure Boot
+    ${sb_menu}=    Enter Secure Boot Menu And Return Construction
+    Enable Secure Boot    ${sb_menu}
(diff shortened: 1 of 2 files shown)


lib/secure-boot-lib.robot
```diff
@@ -218,3 +218,26 @@ Execute File In UEFI Shell
     Press Enter
     ${out}=    Read From Terminal Until    FS0:\\>
     RETURN    ${out}
+
+Remove Old Secure Boot Keys
```
TomaszAIR, line 222: @m-iwanicki lets add `in OS` to the names of those keywords so we could tell that e.g. we are doing `Remove Old Secure Boot Keys in OS` because `Remove Old Secure Boot Keys` could be also done from BIOS menu

lib/secure-boot-lib.robot
```diff
@@ -218,3 +218,26 @@ Execute File In UEFI Shell
...
+    [Documentation]    Removes all files and directories in
```
TomaszAIR, line 223: @m-iwanicki documentation section of those keywords could tell that we are using sbctl there

lib/secure-boot-lib.robot
```diff
@@ -218,3 +218,26 @@ Execute File In UEFI Shell
...
+Enroll Secure Boot Keys
+    [Documentation]    Enrolls current keys to EFI. Returns true if succeeded.
+    ${out}=    Execute Linux Command    sbctl enroll-keys --yes-this-might-brick-my-machine
+    ${sb_status}=    Run Keyword And Return Status
+    ...    Should Contain    ${out}    Enrolled keys to the EFI variables!
+    RETURN    ${sb_status}
+
+Sign All Boot Components
+    [Documentation]    Signs boot components with current keys
+    Execute Linux Command    sbctl verify | awk -F ' ' '{print $2}' | tail -n+2 | xargs -I "#" sbctl sign "#"
```
TomaszAIR, line 243: @m-iwanicki it will try to sign even those components that are already signed?

TomaszAIR (commented):
@m-iwanicki I left a review, could you address and make sure that this is rebased on latest uefi-sb-release branch?

lib/secure-boot-lib.robot
m-iwanicki: Yes. I just checked and it will sign multiple times each time increasing it's size.
It's possible to delete signature with `sudo sbattach --signum 1 --remove /path/to/signed/efi/file`
```
# sudo sbctl sign /boot/efi/EFI/BOOT/BOOTX64.EFI
Signed /boot/efi/EFI/BOOT/BOOTX64.EFI
# sudo sbattach --signum 1 --remove /boot/efi/EFI/BOOT/BOOTX64.EFI
warning: data remaining[827536 vs 953088]: gaps between PE/COFF sections?
```
But size is a little bigger than before signing so it's not perfect.

TomaszAIR: If we are sure that every file will be signed then I am ok with that for this test.

lib/secure-boot-lib.robot
m-iwanicki: added in [5b71c47](https://github.com/Dasharo/open-source-firmware-validation/pull/161/commits/5b71c477c2d2bf39168f700972f6a9e92a46964c)

m-iwanicki:
@TomaszAIR requested changes added in [5b71c47](https://github.com/Dasharo/open-source-firmware-validation/pull/161/commits/5b71c477c2d2bf39168f700972f6a9e92a46964c)

dasharo-security/secure-boot.robot
```diff
@@ -409,6 +409,48 @@ SBO011.001 Attempt to enroll expired certificate and boot signed image
...
+
+    Boot System Or From Connected Disk    ubuntu
+    Login To Linux
+    Switch To Root User
+
+    # 2. Remove old keys then generate and enroll new keys
+    Remove Old Secure Boot Keys In Os
+    ${sb_status}=    Remove Old Secure Boot Keys In Os
+    Should Be True    ${sb_status}
+    ${sb_status}=    Remove Old Secure Boot Keys In Os
```
TomaszAIR, line 435: @m-iwanicki copy&paste error? We now have `Remove Old Secure Boot Keys in Os`. can we replace Os to OS while fixing this?

m-iwanicki: @TomaszAIR fixed both in 8c33406, made mistake with find&replace

PR #167, opened 2023-12-20
Outdated certs test

WiktorG351:
This tests whether we can enroll an outdated certificate, and run a `hello-world` program signed with it. So far it seems that thats possible.

### DIFF
--- a/dasharo-security/secure-boot.robot
+++ b/dasharo-security/secure-boot.robot
@@ -385,9 +385,9 @@ SBO010.006 Check support for ecdsa521 signed certificates
     ${out}=    Execute File In UEFI Shell    hello_ecdsa521.efi
     Should Contain    ${out}    Hello, world!
 
-SBO011.001 Attempt to enroll expired certificate
-    [Documentation]    This test verifies that an expired certificate can/cannot
-    ...    be enrolled.
+SBO011.001 Attempt to enroll expired certificate and boot signed image
+    [Documentation]    This test verifies that an expired certificate can be
+    ...    used to verify booted image.
     Skip If    not ${SECURE_BOOT_SUPPORT}    SBO004.001 not supported
     Skip If    not ${TESTS_IN_FIRMWARE_SUPPORT}    SBO004.001 not supported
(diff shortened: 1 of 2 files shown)


dasharo-security/secure-boot.robot
```diff
@@ -233,6 +233,30 @@ SBO008.001 Attempt to enroll the key in the incorrect format (firmware)
     Select File In File Explorer    DB.txt
     Read From Terminal Until    ERROR: Unsupported file type!
 
+SBO010.001 Attempt to enroll expired certificate
```
TomaszAIR, line 236: @WiktorG351 please, split it into two tests, one that attempt to enroll expired  cert, which should fail in case where we are able to enroll the certificate

Second that try to boot image signed with expired cert (assuming the cert was enrolled before end of date). This should also fail if the image boots.

TomaszAIR: After thinking about it, second test would repeat things the first one does. So lets have one test here.

TomaszAIR: I rebased it and make small changes. Merging.

TomaszAIR:
Merging.

Result
```
λ robot -L TRACE -v rte_ip:127.0.0.1 -v snipeit:no -v config:qemu -t "SBO011.001*" dasharo-security/secure-boot.robot
Secure-Boot                                                                                                            
SBO011.001 Attempt to enroll expired certificate and boot signed i... | FAIL |                                                                                                                                                                
'                                                                                                                      
Hello, world!                                                                                                          
FS0:\>' does not contain 'Access Denied'                                                                                                                                                                                                      
Secure-Boot                                                           | FAIL |                                         
1 test, 0 passed, 1 failed                                                                                                                                                                                                                    
Output:  /home/tzyjewski/files/open-source-firmware-validation/output.xml                                              
Log:     /home/tzyjewski/files/open-source-firmware-validation/log.html                                                
Report:  /home/tzyjewski/files/open-source-firmware-validation/report.html
```

I think that image signed with outdated cert should not be able to boot in qemu but when using Dasharo the hello-world efi app starts even when signed with outdated cert. That is why the test is failing.

TomaszAIR (approved):

PR #171, opened 2023-12-22
Cpu freq fixes

mkopec:
Fix for this false failure:

```
CPF004.005 CPU with load runs on expected frequency (Ubuntu 22.04)... | FAIL |
Several failures occurred:

1) '4400.0 > 4400.0' should be true.

2) '400.0 < 400.0' should be true.
```

Also bump the number of threads in CPU stress test to actually load the CPU

### DIFF
--- a/lib/CPU-performance-lib.robot
+++ b/lib/CPU-performance-lib.robot
@@ -92,7 +92,7 @@ Stress Test
     [Documentation]    Proceed with the stress test.
     [Arguments]    ${time}=60s
     Detect Or Install Package    stress-ng
-    Execute Command In Terminal    stress-ng --cpu 1 --timeout ${time} &> /dev/null &
+    Execute Command In Terminal    stress-ng --cpu 16 --timeout ${time} &> /dev/null &
 
 Check Power Supply
     ${laptop_platform}=    Check The Platform Is A Laptop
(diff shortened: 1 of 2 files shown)


lib/CPU-performance-lib.robot
```diff
@@ -85,14 +85,14 @@ Check CPU Frequency In Windows
...
-        ...    Should Be True    ${freq_max} > ${freq_current}
+        ...    Should Be True    ${freq_max} >= ${freq_current}
     END
 
 Stress Test
     [Documentation]    Proceed with the stress test.
     [Arguments]    ${time}=60s
     Detect Or Install Package    stress-ng
-    Execute Command In Terminal    stress-ng --cpu 1 --timeout ${time} &> /dev/null &
+    Execute Command In Terminal    stress-ng --cpu 16 --timeout ${time} &> /dev/null &
```
macpijan, line 95: @mkopec Perhaps nproc or some variation of it?

macpijan:
@filipleple You might be interested in this

lib/CPU-performance-lib.robot
mkopec: Right, added in https://github.com/Dasharo/open-source-firmware-validation/pull/171/commits/afefd3d45039730357d887da689824bc78d8a75d

PR #186, opened 2024-01-05
DRAFT: Novacustom configuration

Psotas:


macpijan:
@mkopec We have some old changes here, mostly regardin WoL. Do we need them with the current test setup?

mkopec:
@macpijan I don't think so, no

PR #192, opened 2024-01-10
Make sb images

WiktorG351:


scripts/secure-boot/generate-images/intermediate.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+
+# go into a separate directory to make cleanup easier
+mkdir -p something
```
TomaszAIR, line 4: @WiktorG351 in this situations we usually create dir named `tmp`. Even better is to run `mktemp -d`

scripts/secure-boot/generate-images/intermediate.sh
```diff
@@ -0,0 +1,47 @@
...
+cd something
```
TomaszAIR, line 5: @WiktorG351 let's use variable for that instead of hardcoded name

scripts/secure-boot/generate-images/intermediate.sh
```diff
@@ -0,0 +1,47 @@
...
+
+# make the new image
+dd if=/dev/zero of=image.img bs=1M count=8
```
TomaszAIR, line 8: @WiktorG351 count could be a variable, please use variables instead of hardcoded values

scripts/secure-boot/generate-images/intermediate.sh
```diff
@@ -0,0 +1,47 @@
...
+# sign the CSR with the second key pair
+# its necessary to `touch intermediate-db.ext`, otherwise it says its not there
+touch intermediate-db.ext
+openssl x509 -req -in intermediate-db.csr -CA PK.crt -CAkey PK.key -out intermediate-db.crt -CAcreateserial -extfile intermediate-db.ext
+
+# convert the cert which we will enroll to the right format
+openssl x509 -in intermediate-db.crt -outform der -out intermediate-db.der
+
+# sign hello.efi
+cp ../scripts/secure-boot/generate-images/hello.efi .
```
TomaszAIR, line 29: @WiktorG351 better use variable for paths and use something like https://github.com/Dasharo/preseeds/blob/ubuntu-22.04.03/ubuntu/create_image.sh#L7 this way it will work no matter from which location we will execute the script

scripts/secure-boot/generate-images/intermediate.sh
```diff
@@ -0,0 +1,47 @@
...
+sbsign --key intermediate-db.key --cert intermediate-db.crt --output signed-hello.efi hello.efi
+
+# copy the necessary things to the image
+mkdir -p mnt/
```
TomaszAIR, line 33: @WiktorG351 can we mount it under another folder created with mktemp ?

scripts/secure-boot/generate-images/intermediate.sh
```diff
@@ -0,0 +1,47 @@
...
+sudo mount -o loop image.img mnt/
+# simple check to see if it works
+echo "this image was created automatically" > check.txt
+sudo cp check.txt mnt/
+# end of simple check
+sudo cp hello.efi intermediate-db.der signed-hello.efi mnt/
```
TomaszAIR, line 39: @WiktorG351 would be good if every command has some kind of error check, e.g. https://github.com/Dasharo/meta-dts/blob/main/meta-dts-distro/recipes-dts/dts/dts/dts-functions.sh#L83

scripts/secure-boot/generate-images/intermediate.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
```
TomaszAIR, line 1: @WiktorG351 would be nice if we would have one script with functions that allows to prepare needed image based on the provided arguments

dasharo-security/secure-boot.robot
```diff
@@ -239,10 +239,7 @@ SBO009.001 Attempt to boot file signed for intermediate certificate
     Skip If    not ${SECURE_BOOT_SUPPORT}    SBO009.001 not supported
     Skip If    not ${TESTS_IN_FIRMWARE_SUPPORT}    SBO009.001 not supported
     Skip If    not ${TESTS_IN_UBUNTU_SUPPORT}    SBO009.001 not supported
-    Download ISO And Mount As USB
-    ...    ${DL_CACHE_DIR}/${INTERMEDIATE_NAME}
-    ...    ${INTERMEDIATE_URL}
-    ...    ${INTERMEDIATE_SHA256}
+    Make ISO And Mount As USB    intermediate
```
TomaszAIR, line 242: @WiktorG351 this could go to suite setup so we would need to run it once, before running tests, this way we would prepare all needed images, additionally we could add variable so by default those images would not be created (user would need to force creation only once), something similar is here https://github.com/Dasharo/open-source-firmware-validation/pull/193/files

lib/usb-hid-msc-lib.robot
```diff
@@ -33,3 +33,24 @@ Download ISO And Mount As USB
             Skip    unsupported
         END
     END
+
+Make ISO And Mount As USB
+    [Documentation]    first creates the right image using a script,
+    ...    then mounts it
+    [Arguments]    ${img_name}
+
+    ${out}=    Run    sudo ./scripts/secure-boot/generate-images/${img_name}.sh
```
TomaszAIR, line 42: @WiktorG351 do we need to run it with sudo? Maybe there is some keyword builtin RF for such situations? https://robotframework.org/robotframework/

TomaszAIR (changes requested):
@WiktorG351 I added initial review

lib/usb-hid-msc-lib.robot
WiktorG351: @TomaszAIR unfortunately yes. Because we need sudo to `mount`/`umount` the image.

dasharo-security/secure-boot.robot
WiktorG351: @TomaszAIR what if someone is only running one of these tests? Then it would be unnecessary to create all the other images as part of suite setup.

Maybe since the scripts require `sudo` anyway, one solution I can think of is this:
* the repo by default does not have the images
* the tests themselves dont generate the images, but require them
* if you run the tests, and dont have the images, they will fail and tell you that you need to use the script to generate them manually
* then the tests can run
that way we will the generation (and sudo) out of the RF code, and it will be super clean.
Just one keyword like `Mount Image As USB` or something.

scripts/secure-boot/generate-images/intermediate.sh
WiktorG351: @TomaszAIR adressed [here](https://github.com/Dasharo/open-source-firmware-validation/pull/192/commits/d5daafb1fbd4dbfc97289aa9eebda2b7e6c8f7a6#diff-00f53487a7d91b72da783fcf7286011bd2c964cb60496c9c6e607a55e9ae748dR3)

scripts/secure-boot/generate-images/intermediate.sh
WiktorG351: @TomaszAIR ive changed it to use `mktemp -d` like [this](https://github.com/Dasharo/open-source-firmware-validation/pull/192/commits/d5daafb1fbd4dbfc97289aa9eebda2b7e6c8f7a6#diff-00f53487a7d91b72da783fcf7286011bd2c964cb60496c9c6e607a55e9ae748dR9-R10), so no variable neccessary

scripts/secure-boot/generate-images/intermediate.sh
WiktorG351: @TomaszAIR it now uses `$SCRIPTDIR`, for example [here](https://github.com/Dasharo/open-source-firmware-validation/pull/192/commits/d5daafb1fbd4dbfc97289aa9eebda2b7e6c8f7a6#diff-00f53487a7d91b72da783fcf7286011bd2c964cb60496c9c6e607a55e9ae748dR90)

scripts/secure-boot/generate-images/intermediate.sh
WiktorG351: @TomaszAIR since the parent folder is now being made with `mktemp -d`, then it should be fine to keep it as it (?)

dasharo-security/secure-boot.robot
```diff
@@ -252,137 +250,129 @@ SBO009.001 Attempt to boot file signed for intermediate certificate
...
     Power On
 
     ${sb_menu}=    Enter Secure Boot Menu And Return Construction
     Enable Secure Boot    ${sb_menu}
     ${sb_menu}=    Reenter Menu And Return Construction
     ${advanced_menu}=    Enter Advanced Secure Boot Keys Management And Return Construction    ${sb_menu}
     Enter Enroll DB Signature Using File In DB Options    ${advanced_menu}
-    Enter Volume In File Explorer    RSA2K_LABEL
-    Select File In File Explorer    DB-RSA2048.der
+    Enter Volume In File Explorer    NO VOLUME LABEL
```
TomaszAIR, line 272: @WiktorG351 I think @m-iwanicki in his latest PR sets some labels for this images, maybe let's use them?

scripts/secure-boot/generate-images/generate-image.sh
```diff
@@ -0,0 +1,93 @@
...
+    touch cert.ext
+    openssl x509 -req -in cert.csr -CA PK.crt -CAkey PK.key -out cert.crt -CAcreateserial -extfile cert.ext
+    openssl x509 -in cert.crt -outform der -out cert.der
+
+elif [ "$TESTNAME" = "RSA" ]; then
+
+    openssl req -new -x509 -newkey rsa:$N -subj "/CN=PK/" -keyout cert.key -out cert.crt -days 3650 -nodes -sha256
+    openssl x509 -in cert.crt -outform der -out cert.der
+
+elif [ "$TESTNAME" = "ECDSA256" ]; then
```
TomaszAIR, line 38: @WiktorG351 are you sure that we can't prepare one function for those cases?

WiktorG351: @TomaszAIR this was added in [this commit](https://github.com/Dasharo/open-source-firmware-validation/pull/192/commits/73b95c5b65c10a70a68646905f85f706c358f235)

dasharo-security/secure-boot.robot
WiktorG351: @TomaszAIR adressed, i added labels in [this commit](https://github.com/Dasharo/open-source-firmware-validation/pull/192/commits/4de28123ddcbff4af3746e7b54496cdab3c7163f)

lib/usb-hid-msc-lib.robot
macpijan: @TomaszAIR I think it is perfectly fine that RB script just check if the images are available, and if not, just ask to run some preparation script first. Running of this preparation script, can be one of the steps in the `getting started`.

scripts/secure-boot/generate-images/generate-image.sh
```diff
@@ -0,0 +1,94 @@
...
+sbsign --key cert.key --cert cert.crt --output signed-hello.efi hello.efi
+
+# copy all things to the image
+mkdir -p mnt/
+sudo mount -o loop image.img mnt/
+# simple check to see if it works
+echo "this image was created automatically" > check.txt
+# copy everything to the image (except for the image itself)
+sudo find ./ -maxdepth 1 -mindepth 1 ! -name 'image.img' ! -name 'mnt/' -exec cp {} mnt/ \;
+sudo umount image.img
```
m-iwanicki, line 95: @WiktorG351 I'm not sure using sudo inside test is good idea, right now we would need to write password each time.
You could change it to:

```
udisksctl loop-setup -f image.img
```

drop sudo from `find` and add $LABEL variable with image label because udisksctl mounts image at `/run/media/<username>/<label>` or in script `/run/media/$(whoami)/$LABEL`

m-iwanicki (changes requested):

scripts/secure-boot/generate-images/generate-image.sh
m-iwanicki: If you remove those 3 sudo then you could most likely remove them also in `Make Certificate Images For Tests` keyword and `wrapper.sh` file.

lib/secure-boot-lib.robot
```diff
@@ -44,6 +44,11 @@ ${EXPIRED_SHA256}=          831c4f0ec9bf4e40a80e15deb16d76a60524e91bd9e9ebe07847
 
 
 *** Keywords ***
+Make Certificate Images For Tests
+    [Documentation]    This keyword generates all the necessary images
+    ...    for secure boot tests.
+    Run    sudo ./scripts/secure-boot/generate-images/wrapper.sh
```
m-iwanicki, line 50: You should probably check that return code is 0 (finished correctly)

```
    ${rc}=    Run And Return RC    ./scripts/secure-boot/generate-images/wrapper.sh
    Should Be Equal As Integers   ${rc}   0
```

You would also need to modify scripts so they exit if command fails. You can do that by adding `set -e` at the beginning of each script (wrapper and generate-image). It will make whole script fail immediately if any command fails.
This can be disabled using `set +e` before commands you want to allow to fail and then using `set -e` again after them.

m-iwanicki (changes requested):
Couple more changes, mostly so script won't show errors.

scripts/secure-boot/generate-images/generate-image.sh
```diff
@@ -0,0 +1,102 @@
+#!/bin/bash
+
+SCRIPTDIR=$(readlink -f $(dirname "$0"))
```
m-iwanicki, line 3:
```
In scripts/secure-boot/generate-images/generate-image.sh line 3:
SCRIPTDIR=$(readlink -f $(dirname "$0"))
                        ^------------^ SC2046 (warning): Quote this to prevent word splitting.
```

scripts/secure-boot/generate-images/generate-image.sh
```diff
@@ -0,0 +1,102 @@
...
+COUNT=$1
+TESTNAME=$2
+N=$3
+
+# go into a separate directory to make cleanup easier
+TEMPDIR=$(mktemp -d)
+cd "$TEMPDIR"
```
m-iwanicki, line 10:
```
In scripts/secure-boot/generate-images/generate-image.sh line 11:
cd "$TEMPDIR"
^-----------^ SC2164 (warning): Use 'cd ... || exit' or 'cd ... || return' in case cd fails.
```

You could also use set -e as mentioned earlier.

scripts/secure-boot/generate-images/wrapper.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+SCRIPTDIR=$(readlink -f $(dirname "$0"))
```
m-iwanicki, line 3:
```
In scripts/secure-boot/generate-images/wrapper.sh line 3:
SCRIPTDIR=$(readlink -f $(dirname "$0"))
                        ^-------------^ SC2046 (warning): Quote this to prevent word splitting.
```

m-iwanicki (changes requested):

scripts/secure-boot/generate-images/wrapper.sh
WiktorG351: pre-commit now passes, changed [this line](https://github.com/Dasharo/open-source-firmware-validation/pull/192/commits/bce10d72eb89079d32fb6db71846782fa2f3e593#diff-ed91284b570a2b939ac0f9b2f2920b30663fb3a865b8503338410983fd0a84c2R5)

scripts/secure-boot/generate-images/generate-image.sh
WiktorG351: we now use [set -e](https://github.com/Dasharo/open-source-firmware-validation/pull/192/commits/bce10d72eb89079d32fb6db71846782fa2f3e593#diff-00f53487a7d91b72da783fcf7286011bd2c964cb60496c9c6e607a55e9ae748dR3), in both scripts

scripts/secure-boot/generate-images/generate-image.sh
WiktorG351: fixed [here](https://github.com/Dasharo/open-source-firmware-validation/pull/192/commits/bce10d72eb89079d32fb6db71846782fa2f3e593#diff-00f53487a7d91b72da783fcf7286011bd2c964cb60496c9c6e607a55e9ae748dR5), pre-commit passes

lib/secure-boot-lib.robot
WiktorG351: adressed in [this commit](https://github.com/Dasharo/open-source-firmware-validation/pull/192/commits/2c746fe7fb34c331de12a48ef66a0aa9eaa5cd63)

scripts/secure-boot/generate-images/generate-image.sh
```diff
@@ -0,0 +1,102 @@
...
+sudo mount -o loop image.img mnt/
+# simple check to see if it works
+echo "this image was created automatically" > check.txt
+# copy everything to the image (except for the image itself)
+sudo find ./ -maxdepth 1 -mindepth 1 ! -name 'image.img' ! -name 'mnt/' -exec cp {} mnt/ \;
+sudo umount image.img
+
+# now the image is ready, all we have to do is copy it to the desired location
+IMAGENAME=$TESTNAME$N
+cp image.img "$SCRIPTDIR"/../images/$IMAGENAME.img
```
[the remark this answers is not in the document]
WiktorG351: adressed in [this commit](https://github.com/Dasharo/open-source-firmware-validation/pull/192/commits/e550192fd9659acbc3a55a8c40b33542d7688153)

scripts/secure-boot/generate-images/generate-image.sh
```diff
@@ -0,0 +1,102 @@
...
+cp "$SCRIPTDIR"/hello.efi .
+sbsign --key cert.key --cert cert.crt --output signed-hello.efi hello.efi
+
+# copy all things to the image
+mkdir -p mnt/
+sudo mount -o loop image.img mnt/
+# simple check to see if it works
+echo "this image was created automatically" > check.txt
+# copy everything to the image (except for the image itself)
+sudo find ./ -maxdepth 1 -mindepth 1 ! -name 'image.img' ! -name 'mnt/' -exec cp {} mnt/ \;
```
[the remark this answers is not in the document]
WiktorG351: this is now being done with a [for loop instead](https://github.com/Dasharo/open-source-firmware-validation/pull/192/commits/bce10d72eb89079d32fb6db71846782fa2f3e593#diff-00f53487a7d91b72da783fcf7286011bd2c964cb60496c9c6e607a55e9ae748dR99-R105)

scripts/secure-boot/generate-images/generate-image.sh
```diff
@@ -0,0 +1,102 @@
...
+elif [ "$TESTNAME" = "EXPIRED" ]; then
+
+    faketime '10 days ago' openssl req -new -x509 -newkey rsa:2048 -subj "/CN=ExpiredCert/" -keyout cert.key -out cert.crt -days 1 -nodes -sha256
+    openssl x509 -in cert.crt -outform der -out cert.der
+
+fi
+
+# sign hello.efi
+cp "$SCRIPTDIR"/hello.efi .
+sbsign --key cert.key --cert cert.crt --output signed-hello.efi hello.efi
```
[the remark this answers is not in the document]
WiktorG351: added this [here](https://github.com/Dasharo/open-source-firmware-validation/pull/192/commits/bce10d72eb89079d32fb6db71846782fa2f3e593#diff-00f53487a7d91b72da783fcf7286011bd2c964cb60496c9c6e607a55e9ae748dR88-R90)

scripts/secure-boot/generate-images/generate-image.sh
WiktorG351: turns out [0.2 seconds](https://github.com/Dasharo/open-source-firmware-validation/pull/192/commits/bce10d72eb89079d32fb6db71846782fa2f3e593#diff-00f53487a7d91b72da783fcf7286011bd2c964cb60496c9c6e607a55e9ae748dR95) is enough :)

lib/secure-boot-lib.robot
m-iwanicki: Works but maybe change to: `${rc}    ${output}=    Run And Return RC And Output`.
This way when there is error you could see why in logs. Currently there is no information.

scripts/secure-boot/generate-images/generate-image.sh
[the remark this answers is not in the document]
WiktorG351: @m-iwanicki  sorry, that was added before `$SCRIPTDIR` is defined, which caused the script to fail. Actually fixed [here](https://github.com/Dasharo/open-source-firmware-validation/pull/192/commits/1027b0e910c1cebb459ac7b4880dea008bafd43f)

lib/secure-boot-lib.robot
WiktorG351: this is a great idea - added [here](https://github.com/Dasharo/open-source-firmware-validation/pull/192/commits/531a320643f8b7fb2630b6fdaa0630fe1f044f66)

m-iwanicki (approved):

dasharo-security/secure-boot.robot
TomaszAIR: This is better approach, we should just add needed information in README how to prepare env to run test (e.g. run script before running tests)

lib/usb-hid-msc-lib.robot
TomaszAIR: At the end this is a way that we chose. Info added to README https://github.com/Dasharo/open-source-firmware-validation/pull/192/commits/16058f263b77da1e7919c2a1d5eb0b4018ddde46

TomaszAIR:
@WiktorG351 I made couple of changes, kw now only checks if given image is available, if not then suite fails with info that sb-img-wrapper.sh script should be used to create images.

Test results with images created from script. Tested on QEMU.

```
λ robot -b file.txt -L TRACE -v ansible_config:yes -v rte_ip:127.0.0.1 -v snipeit:no -v config:qemu dasharo-security/secure-boot.robot                          
Secure-Boot                                                                                                                                                                                                                                   
SBO001.001 Check Secure Boot default state (firmware) :: This test... | PASS |                                                                                                                                                                
SBO002.001 UEFI Secure Boot (Ubuntu 22.04) :: This test verifies t... | PASS |                                         
SBO002.002 UEFI Secure Boot (Windows 11) :: This test verifies tha... | SKIP |                                                                                                                                                                
SBO002.002 not supported                                                                                                                                                                                                                      
SBO003.001 Attempt to boot file with the correct key from Shell (f... | PASS |                                         
SBO004.001 Attempt to boot file without the key from Shell (firmwa... | PASS |                                         
SBO005.001 Attempt to boot file with the wrong-signed key from She... | PASS |                                         
SBO006.001 Reset Secure Boot Keys option availability (firmware) :... | PASS |                                                                                                                                                                
SBO007.001 Attempt to boot the file after restoring keys to defaul... | PASS |                                                                                                                                                                
SBO008.001 Attempt to enroll the key in the incorrect format (firm... | PASS |                                                                                                                                                                
SBO009.001 Attempt to boot file signed for intermediate certificat... | PASS |                                                                                                                                                                
SBO010.001 Check support for rsa2k signed certificates :: PEM gene... | PASS |                                         
SBO010.002 Check support for rsa3k signed certificates :: PEM gene... | PASS |                                                                                                                                                                
SBO010.003 Check support for rsa4k signed certificates :: PEM gene... | PASS |                                                                                                                                                                
SBO010.004 Check support for ecdsa256 signed certificates :: PEM g... | FAIL |                                         
'                                                                                                                      
Command Error Status: Access Denied                                                                                                                                                                                                           
FS0:\>' does not contain 'Hello, world!'                                                                               
SBO010.005 Check support for ecdsa384 signed certificates :: PEM g... | FAIL |
'
Command Error Status: Access Denied
FS0:\>' does not contain 'Hello, world!'
SBO010.006 Check support for ecdsa521 signed certificates :: PEM g... | FAIL |
'
Command Error Status: Access Denied
FS0:\>' does not contain 'Hello, world!'
SBO011.001 Attempt to enroll expired certificate and boot signed i... | FAIL |
'
Hello, world!
FS0:\>' does not contain 'Access Denied'
```

last four can fail, in case of ecdsa BIOS cannot enroll ecdsa certs and with expired image, BIOS still boots it even if certificate is expired - so in this cases we have FAIL.

If one of the images is missing, tests fails immediately.

```
λ rm scripts/secure-boot/images/GOOD_KEYS.img
λ robot -b file.txt -L TRACE -v ansible_config:yes -v rte_ip:127.0.0.1 -v snipeit:no -v config:qemu dasharo-security/secure-boot.robot
Secure-Boot                                                                   
SBO001.001 Check Secure Boot default state (firmware) :: This test... | FAIL |                                         
Parent suite setup failed:                                                                                             
Image GOOD_KEYS.img does not exist! Please run ./scripts/secure-boot/generate-images/sb-img-wrapper.sh script.
```

TomaszAIR (approved):

PR #214, opened 2024-03-06
keywords.robot: list tests revision in log and report

krystian-hebel:
This is done to keep track of which version of test was actually run, as well as whether local changes were made. It should make it possible to repeat the same tests in the future and hopefully get the same results.

Link to source code of test is also added to suite metadata, but this may be just an approximation in case of local modifications. In such case, reported revision will have '-dirty' appended to the digest.


krystian-hebel:
This is how it looks like:
image

macpijan:
Thanks @krystian-hebel , very useful improvement

macpijan (approved):

pietrushnic:
@krystian-hebel Yes, this is very useful. It should primarily improve future analysis, and IIUC can prevent releasing based on test results, which are made on a modified version of the repository.

PR #220, opened 2024-03-08
os-config/ubuntu-credentials.robot: use correct credentials

pkubaj:


macpijan:
@pkubaj no need to contribute from fork, you should be able to create branches directly in dasharo.

macpijan (approved):

PR #223, opened 2024-03-13
Add configs for apu{2,3,6}

pkubaj:


macpijan:
@pkubaj Some pre-commit failures slipped through the cracks: https://results.pre-commit.ci/run/github/554712842/1710315604.E2-RLA3JShmb559hG6i0Xg

It is not from your code, but I would appreciate, if you could fix it as well.

Changes LGTM, we just need some logs proving they work fine.

pkubaj:
I needed to switch the boot menu shortcut to F10, because APU2 uses F10.

mkopec:
Ah right, I tested a build from before the boot menu keys were changed.

pkubaj:
[logs.tar.gz](https://github.com/Dasharo/open-source-firmware-validation/files/14589955/logs.tar.gz)
Here are logs for apu2. For apu3, the cable used for flashing doesn't work, so I couldn't test it there, but since @mkopec agrees to the boot key change, I assume it will be fine as well.

mkopec (approved):
I had already addressed the CI errors in https://github.com/Dasharo/open-source-firmware-validation/pull/221 but it's ok, I'll just drop these commits in there.

pkubaj:
Logs from APU6
[logs.tar.gz](https://github.com/Dasharo/open-source-firmware-validation/files/14603600/logs.tar.gz)

APU3 wasn't tested because of no output on the serial (even after flashing the last public coreboot release).

PR #227, opened 2024-03-14
Add a test for CPB

pkubaj:
keys.robot is included in the pcengines.robot file because keys are actually used there.


pkubaj:
Closing, source branch is not OK.

PR #229, opened 2024-03-15
Wdat wdt tests

mkopec:
Based on PCE WDT tests.

### DIFF
--- a/dasharo-compatibility/watchdog.robot
+++ b/dasharo-compatibility/watchdog.robot
@@ -1,0 +1,107 @@
+*** Settings ***
+Library             Collections
+Library             OperatingSystem
+Library             Process
+Library             String
+Library             Telnet    timeout=20 seconds    connection_timeout=120 seconds
+Library             SSHLibrary    timeout=90 seconds
+Library             RequestsLibrary
+# TODO: maybe have a single file to include if we need to include the same
+# stuff in all test cases
+Resource            ../lib/bios/menus.robot
+Resource            ../sonoff-rest-api/sonoff-api.robot
+Resource            ../rtectrl-rest-api/rtectrl.robot
+Resource            ../variables.robot
+Resource            ../keywords.robot
+Resource            ../keys.robot
+
+# TODO:
+# - document which setup/teardown keywords to use and what are they doing
+# - go through them and make sure they are doing what the name suggests (not
+#    exactly the case right now)
+Suite Setup         Run Keyword
+...                     Prepare Test Suite
+Suite Teardown      Run Keywords
+...                     Log Out And Close Connection
+
+
+*** Test Cases ***
+WDT001.001 Check if watchdog option is available
+    [Documentation]    Check if the watchdog timer can be enabled in the chipset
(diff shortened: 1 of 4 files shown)


mkopec:
The only real difference between these and https://github.com/Dasharo/open-source-firmware-validation/pull/221 are the options being located in different submenus, and the default timeout values. These tests could probably be combined with a bit of effort.

dasharo-compatibility/watchdog.robot
```diff
@@ -0,0 +1,108 @@
...
+    ...    configuration submenu.
+    Skip If    not ${TESTS_IN_FIRMWARE_SUPPORT}    APU001.001 not supported
+    Skip If    not ${WATCHDOG_SUPPORT}    Watchdog tests not supported.
+    Power On
+    ${setup_menu}=    Enter Setup Menu Tianocore And Return Construction
+    ${dasharo_menu}=    Enter Dasharo System Features    ${setup_menu}
+    ${chipset_menu}=    Enter Dasharo Submenu    ${dasharo_menu}    Chipset Configuration
+    Should Contain Match    ${chipset_menu}    Enable watchdog*
+
+WDT002.001 Enable watchdog
```
macpijan, line 40: @BeataZdunczyk @mkopec I think it would be beneficial to test WDT in OS as well:
- if there is no unwanted reset in OS with our default OS installation
- if the WDT triggers reset when on certain situations (such as kernel crash)
- if the reset reason is correct when WDT has triggered

If not here, then maybe create issue and add when possible.

mkopec: I've still got some time for this task, will try to add some tests for this. At least Ubuntu has the wdat_wdt module OOTB (but doesn't load it by default for some reason 🙄)

macpijan: > (but doesn't load it by default for some reason 🙄)


That sounds a bit problematic. Also, any idea on BSD?

mkopec: Supposedly FreeBSD does have a driver: https://www.mail-archive.com/dev-commits-src-main@freebsd.org/msg19415.html

Ubuntu has the driver, just doesn't load it by default. It works anyway after `modprobe wdat_wdt`. Arch loads the module by default, so it's probably something Ubuntu is doing. Windows also works.

We should probably not enable the watchdog by default, but expose the option in settings

macpijan: > We should probably not enable the watchdog by default, but expose the option in settings

Definitely

mkopec: Working on it, need to untangle the knob to show the watchdog options from the actual watchdog state first, it was all written with the assumption the watchdog is always enabled by default

mkopec: Implemented in https://github.com/Dasharo/coreboot/pull/472 , https://github.com/Dasharo/edk2/pull/124 and https://github.com/Dasharo/DasharoModulePkg/pull/42

PR #231, opened 2024-03-15
README.md: add OSFV requirements diagrams

artur-rs:


pietrushnic:
@macpijan let's get back here in context of Introductory Lab Assembly course

pietrushnic:
@macpijan I see you merged it without any comments. Also, I completely forgot to submit mine despite it being ready on 15th March.

We have to get back to this topic. One of my goals for hardwear.io training is an `Overview of hardware theory, including flashing firmware, recovery methods, and reading from serial`, the Lab Assembly Guide. I want to train new personnel with that short session, but first, I would like to agree on the scope to maximize use for Dasharo and Zarhus Team. Ideally, we would have one such session, which would be recorded and edited for online consumption and remote access, or someone from the organization could take over from me for that training. Still, first, we would have to ensure we teach helpful stuff. Please help me maximize the synergy of that effort.

@miczyg1 since you often question the quality of education in the hardware-facing areas, your input would be invaluable.

miczyg1:
The diagrams' format does not seem to be the right match for purposes like these. As I can see we mostly have a decision flow (each step is a "question? -> yes/no -> action"). it might be more readable to have a flow diagram which goes downwards like this:

pietrushnic:
@miczyg1, no other comments than what tools and how were used? If the type of diagram were changed, would you have anything to add?

macpijan:
> I see you merged it without any comments. Also, I completely forgot to submit mine despite it being ready on 15th March.

Yes, it was up too long and we can always improve.

macpijan:
> The diagrams' format does not seem to be the right match for purposes like these (at least for me, getting the conclusion out of theese diagrams is difficult).

That was always my comment for these UML diagrams as well, I think I gave up at some point

Issue #257, opened 2024-04-02
Use OSFV python lib to implement low-level interfaces, like flashing


macpijan:
**Test environment version**
* Branch: 
* Commit: 

**Dasharo version (if applicable)**

**Dasharo variant (if applicable)**

**Affected component(s) or functionality (if applicable)**

**Brief summary**

Implement DRY strategy.
Avoid reimplementing low-level flashing and platform control login in multiple places:
- test repo
- CLI repo
- flashing scripts on the platform

Keep the low-level details only in one place, and provide an interface to be used in tests.

**Additional context**

WIP work:
- https://github.com/Dasharo/open-source-firmware-validation/tree/osfv-integration
- https://github.com/Dasharo/osfv-scripts/tree/osfv-integration

macpijan:
@PLangowski 

I have tried various way of sharing this, and probably the best way is as in
the linked branches. I would start from there adn check if we can call
basic functions (like flash probe/read) from osfv-script repo in all these cases:
- from osfv_cli in poetry shell: https://github.com/Dasharo/osfv-scripts/tree/main/osfv_cli#development
- from osfv_cli when installed through Make install
- from robot tests, through the RobotRTE library (can be check on any of the available platforms)

If the answer is yes, and we cannot see probles with this approach, we can continue with
cleanup, further improvements, and replacing the previous Rte library with the calls
from osfv-scripts libs.

High-level TODOs:
- replace calls to https://github.com/3mdeb/rtectrl-rest-api with these to
  osfv-scripts libs 
- replace calls to https://github.com/3mdeb/sonoff-rest-ap with these to
  osfv-scripts libs
- drop the low-level hardware access procedures from open-source-firmware-validation
  and replace them with higher-level calls from osfv-scripts; especially the flashing,
  perhaps also the Power On, and similar
- use snipeit from osfv-scripts package, not imported from path like here:
  https://github.com/Dasharo/open-source-firmware-validation/commit/c08c808be6878e462587bb940a42bdb9c0a4f6e0#diff-2bd69584d51e03632b6b478cae660b9d98acb614bedfcf8ad005f876c9d6744eR528
- in the open-source-firmware-validation repo we've had the `-v snipeit:no` switch
  which allowed us to run the scripts without snipeit; this should still be somehow
  supported, but the snipeit calls are somehow convoluted in the osfv-scripts
  already; we need a proposal on how to handle that
[...]

PR #258, opened 2024-04-03
[DRAFT] Implement read firmware

PLangowski:


lib/flash.robot
```diff
@@ -185,3 +185,11 @@ Compare Write Protection Ranges
     IF    ${set_length}!=${length}
         FAIL    Declared and currently set protection lengths are not the same
     END
+
+Read Firmware
+    [Documentation]    Read platform firmware to file specified in the argument.
+    [Arguments]    ${file}    ${flags}=""
+    RteCtrl Power Off
+    Sleep    2s
+    SSHLibrary.Execute Command    flashrom -p linux_spi:dev=/dev/spidev1.0,spispeed=16000 -r /tmp/coreboot.rom ${flags}
```
macpijan, line 194: Unfortunately, it will work only in a  very specific case, for APU platforms setup with RTE rev v1.0.0 or lower.
For different RTE revision, or different platform, it will not work.

Only after implementing: https://github.com/Dasharo/open-source-firmware-validation/issues/257 we will be able to have more generic read/write keywords in this repo I am afraid.

Please link to this issue and detect the platform type. If it is different than any of the apu, fail with a message that this kwd is not supported for this platform yet.

PLangowski: Is this what you meant by linking the issue?: 90f743f7b6f3b46b360c57f2725f60a7854b7013

keywords.robot
```diff
@@ -1321,7 +1321,7 @@ Firmware Version Verification From Binary
     [Documentation]    Check whether the DUT firmware version is the same as it
     ...    is expected by checking it with dmidecode and comparing
     ...    with a value get from binary.
-    Read Firmware    ${TEMPDIR}${/}coreboot.rom
+    Read Firmware    ${TEMPDIR}${/}coreboot.rom    ${FLASHROM_FLAGS}
```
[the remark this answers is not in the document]
PLangowski: @macpijan Not yet, but if we ever need to use this keyword for a different platform, it might be useful

macpijan (approved):

PR #266, opened 2024-04-12
Vp2420 fixes

macpijan:
We still need to take a look at:
- [ ]  https://github.com/Dasharo/dasharo-issues/issues/740 - likely a change in tests
- [ ] Proper frequency calculation and offset as discussed in: https://github.com/Dasharo/dasharo-issues/issues/763

### DIFF
--- a/platform-configs/novacustom-ns50mu.robot
+++ b/platform-configs/novacustom-ns50mu.robot
@@ -4,19 +4,33 @@ Resource    include/novacustom-tgl.robot
 
 
 *** Variables ***
-# CPU
-${CPU}=                         Intel(R) Core(TM) i7-1165G7 CPU
+${DUT_CONNECTION_METHOD}=               SSH
+${PAYLOAD}=                             tianocore
+${RTE_S2_N_PORT}=                       ${EMPTY}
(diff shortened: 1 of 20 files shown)


filipleple (approved):
LGTM - does fix `Boot Time Measure`

filipleple:
Windows tests fail after 5 mins, probably due to autosleep - which is strange, since I recall disabling it.  Will investigate and fix

platform-configs/novacustom-ns50mu.robot
```diff
@@ -17,306 +17,18 @@ ${IPXE_BOOT_ENTRY}=                                 iPXE Network boot
 ${AUTO_BOOT_TIME_OUT_DEFAULT_VALUE}=                ${EMPTY}
 ${LAPTOP_EC_SERIAL_WORKAROUND}=                     ${TRUE}
 
-# Platform flashing flags
```
pkubaj, line 20: Is it necessary to remove 90% of this file?

macpijan: I see only indentation changes here. I assume this comment was for older patch version.

Issue #132, opened 2023-12-04
"Make Sure That Network Boot Is Enabled" cannot be used in Suite Setup because of QEMU not working 


macpijan:
### Device

QEMU

### RTE version

N/A

### Affected component(s) or functionality

`Make Sure That Network Boot Is Enabled` keyword

### Brief summary

`Make Sure That Network Boot Is Enabled` keyword cannot be used in Suite Setup

### How reproducible

100%

### How to reproduce

Move it from `Test Setup` to `Suite Setup` in: `dasharo-compatibility/network-boot.robot`

### Expected behavior

kwd can be used in a test suite

### Actual behavior

When running test on HW, there is no problem.
When running test on QEMU it hangs somewhere in `Power On` / `Read From Terminal`: https://matrix.to/#/!MwWOJhMJzlxIdAQxae:matrix.3mdeb.com/$CIyMWfI4F7tcQ0PESicObkPftDcqfaTH-bRP72SMXG4?via=matrix.org&via=matrix.3mdeb.com&via=infosec.exchange

### Link to screenshots or logs

N/A

macpijan:
Should be already fixed by this commit: https://github.com/Dasharo/open-source-firmware-validation/commit/924c5499fa408ab631bc00520f9588d57ce8efdb

macpijan:
The kwd has been moved from Test Setup into Suite Setup here: f103dfaaee00e50757762b8aa3c10bf1f21534dc

With successful results in QEMU

PR #267, opened 2024-04-15
Dts update test

WiktorG351:
This test is a proof of concept.

The test scenario is the same as in https://github.com/Dasharo/meta-dts/pull/114

Unfortunately it fails because of lack of cloud keys. The `dts-boot` script thats in `unit-tests` of `meta-dts` contains a section thats responsible for getting the right keys. The release version of DTS obviously can't have that, and I think that further editing of `dts` script on qemu through robotframework is a bad idea.

### DIFF
--- a/dasharo-compatibility/dts-update.robot
+++ b/dasharo-compatibility/dts-update.robot
@@ -13,12 +13,11 @@ Resource            ../sonoff-rest-api/sonoff-api.robot
 Resource            ../variables.robot
 
 Suite Setup         Run Keywords
-...                     Prepare Test Suite
+...                     Prepare Test Suite    AND
+...    Make Sure That Network Boot Is Enabled    AND
+...    Restore Initial DUT Connection Method
 Suite Teardown      Run Keyword
 ...                     Log Out And Close Connection
-# This must be in Test Setup, not Suite Setup, because of a known problem
-# with QEMU: https://github.com/Dasharo/open-source-firmware-validation/issues/132
-Test Setup          Run Keywords    Make Sure That Network Boot Is Enabled    AND    Restore Initial DUT Connection Method
 
 
 *** Test Cases ***
--- a/dasharo-compatibility/dts-update.robot
+++ b/dasharo-compatibility/dts-update.robot
@@ -1,0 +1,94 @@
+*** Settings ***
+Library             Collections
+Library             OperatingSystem
+Library             Process
+Library             String
+Library             Telnet    timeout=20 seconds    connection_timeout=120 seconds
+Library             SSHLibrary    timeout=90 seconds
+Library             RequestsLibrary
+Resource            ../keywords.robot
+Resource            ../keys.robot
+Resource            ../rtectrl-rest-api/rtectrl.robot
+Resource            ../sonoff-rest-api/sonoff-api.robot
+Resource            ../variables.robot
+
+Suite Setup         Run Keywords
+...                     Prepare Test Suite
+Suite Teardown      Run Keyword
+...                     Log Out And Close Connection
+# This must be in Test Setup, not Suite Setup, because of a known problem
+# with QEMU: https://github.com/Dasharo/open-source-firmware-validation/issues/132
+Test Setup          Run Keywords    Make Sure That Network Boot Is Enabled    AND    Restore Initial DUT Connection Method
+
+
+*** Test Cases ***
+DTS-2 NovaCustom Dasharo v1.7.2
+    [Documentation]    This test aims to verify that Dasharo v1.7.2 on NV4x_PZ
+    ...    eligible for updates to heads with heads DES and regular update
+    ...    works as expected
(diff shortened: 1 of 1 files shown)


dasharo-compatibility/dts-update.robot
```diff
@@ -0,0 +1,94 @@
...
+Resource            ../keys.robot
+Resource            ../rtectrl-rest-api/rtectrl.robot
+Resource            ../sonoff-rest-api/sonoff-api.robot
+Resource            ../variables.robot
+
+Suite Setup         Run Keywords
+...                     Prepare Test Suite
+Suite Teardown      Run Keyword
+...                     Log Out And Close Connection
+# This must be in Test Setup, not Suite Setup, because of a known problem
```
macpijan, line 19: Should be already fixed: https://github.com/Dasharo/open-source-firmware-validation/issues/132#issuecomment-2063498232
Can you please verify?

dasharo-compatibility/dts-update.robot
```diff
@@ -0,0 +1,94 @@
...
+    Boot Dasharo Tools Suite    iPXE
+    Write Into Terminal    9
+    Read From Terminal Until Regexp    bash-\\d\\.\\d#
+
+    # We have to do this so that dts-environment doesnt ruin our mock exports
+    # I am putting this in a script to break up the command and escape
+    # robotframeworks variable expansion. Unfortunately running this as one
+    # command is impossible.
+    # This could be removed once the new dts release contains this change
+    # (PR for this is submitted already)
```
macpijan, line 42: I would expect link to PR here if we mention it.

macpijan, line 42: But on another note, do we really must have these changes in release image prior running tests? We need to be able to test the changes we have locally, using QEMU as an environment. Cannot we deploy our local copy of the scripts (via SSH for instance), and then run the tests on modified scripts?

dasharo-compatibility/dts-update.robot
```diff
@@ -0,0 +1,94 @@
...
+    [Documentation]    This test aims to verify that Dasharo v1.7.2 on NV4x_PZ
+    ...    eligible for updates to heads with heads DES and regular update
+    ...    works as expected
+    Skip If    not ${TESTS_IN_FIRMWARE_SUPPORT}    DTS005.001 not supported
+    Skip If    not ${DTS_SUPPORT}    DTS005.001 not supported
+
+    Power On
+    Boot Dasharo Tools Suite    iPXE
+    Write Into Terminal    9
+    Read From Terminal Until Regexp    bash-\\d\\.\\d#
```
macpijan, line 35: This looks like prompt, why not set the prompt and Read Until Prompt?

macpijan: I **think** it could be problematic in a way that the prompt changes on bash version change.
But we can change the prompt in DTS. Can you create an issue for that, so the prompt is persistent, not related to bash version?

dasharo-compatibility/dts-update.robot
```diff
@@ -0,0 +1,94 @@
...
+    Log    ${out}
+
+    ${out}=    Write Into Terminal    ./usr/sbin/dts-boot
+
+    ${out}=    Read From Terminal Until    Enter an option:
+    Write Into Terminal    5
+    Log    ${out}
+
+    ${out}=    Read From Terminal Until    Would you like to switch to Dasharo heads firmware? (Y|n)
+    # test fails here, because it doesnt have the right cloud keys
```
macpijan, line 75: It means we should first provide them, right?
How do we provide them when testing locally?

dasharo-compatibility/dts-update.robot
WiktorG351: well originally it was [here](https://github.com/Dasharo/meta-dts/pull/116) but then I remembered that the scripts have been moved so I made a [new one](https://github.com/Dasharo/dts-scripts/pull/9)

dasharo-compatibility/dts-update.robot
WiktorG351: Yes.

Locally its a lot different since we use [this version of dts-boot](https://github.com/Dasharo/meta-dts/blob/main/unit_tests/dts-boot) from `meta-dts/unit-tests`, and it automatically takes care of the keys.

macpijan: Does it have to be a lot different? Cannot we send the files with keys (or event mount the repository into qemu) and use them in the similar way?

dasharo-compatibility/dts-update.robot
WiktorG351: removed the unnecessary test setup [here](https://github.com/Dasharo/open-source-firmware-validation/pull/267/commits/a47ea84380a7aca7847890baa0e1e8d0406c83d7)

artur-rs:
@macpijan this can be closed based on the work introduced here https://github.com/Dasharo/open-source-firmware-validation/pull/333 ?

PR #273, opened 2024-04-22
fix issue #247

### LINKED ISSUE #247: APU: Option Secure Boot Mode not found in the list
### Device

APU2

### RTE version

1.1.0

### Affected component(s) or functionality

_No response_

### Brief summary

APU: Option Secure Boot Mode not found in the list

### How reproducible

_No response_

### How to reproduce

Try executing SBO* tests.

### Expected behavior

Tests should run.

### Actual behavior

Option Secure Boot Mode not found in the list

### Link to screenshots or logs

image

### Additional context

_No response_

### Solutions you've tried

_No response_

PLangowski:


macpijan:
@PLangowski Can you please provide some context on why it fixes the problem? It happens only on the APU2, or the other platforms as well (especially QEMU)?

macpijan:
@PLangowski I would appreciate a response here

PLangowski:
@macpijan This fixes 2 problems:
- The `Reenter menu ...` keyword sometimes returns the previous menu's construction despite the fact that it did reenter the menu. The duplicated keyword is a workaround to this problem - when it's present, all my tests passed.
- Sometimes APU reads the ` Secure Boot Configuration` menu title as an option, which breaks the next steps. This is why I remove it from the returned construction

The problem seems to only appear on APU2 (I haven't checked APU3 yet, because it is not available right now).

PLangowski:
@macpijan I tested it on APU3 as well and the problem did not occur there.

macpijan:
I am still not a huge fan of this workaround.

> The Reenter menu ... keyword sometimes returns the previous menu's construction despite the fact that it did reenter the menu.

Can we isolate this problem further? For instance, have a test suite stressing the problematic keyword, show what the ratio of failures is, attach logs from both PASS and FAIL scenarios.

> Sometimes APU reads the  Secure Boot Configuration menu title as an option, which breaks the next steps. This is why I remove it from the returned construction

If anything, this should be removed much higher in a parsing keyword, not in this particular case. What you say here, implies that there is a problem with keyword, so it must be resolved in a more generic way.

We will not progress much on quality with such approach, we need a deeper understanding of the problems, rather than adding workarounds on top of each other.

PLangowski:
@macpijan I have added a test that tests the `Reenter menu` keyword: https://github.com/Dasharo/open-source-firmware-validation/pull/329

I ran the test on APU2, and to my surprise it passed. I then ran `SBO001.001 Check Secure Boot default state (firmware)` several times on the latest `develop` and it also works. It's quite baffling that the error no longer occurs. Maybe somehting changed on `develop`? I'm attaching logs.

[reener-menu-test.txt](https://github.com/user-attachments/files/15986070/reener-menu-test.txt)
[sb-default-state.txt](https://github.com/user-attachments/files/15986071/sb-default-state.txt)

PR #280, opened 2024-04-25
platform-configs: add expected CPU frequency ranges

filipleple:


macpijan:
@filipleple It was still a draft, where do we need this?

@BeataZdunczyk CC

filipleple:
@macpijan grepping through the latest develop it looks like these variables are already there, not sure why github can't recognize this. The changes were probably copied over and merged via another PR

PR #285, opened 2024-05-06
SD card fixes

pkubaj:
1. Check for SD card type instead of a given vendor and model.
2. Check for mmcblkX apart from sdX.


pkubaj:
I forgot that I previously created this PR, closing.

PR #286, opened 2024-05-06
Detect average temperature in CPU tests

pkubaj:
Not run-time tested yet, because our 6670's are currently checked out.

### DIFF
--- a/dasharo-performance/cpu-temperature.robot
+++ b/dasharo-performance/cpu-temperature.robot
@@ -124,15 +124,21 @@ CPU Temperature Without Load (Ubuntu 22.04)
     Detect Or Install Package    lm-sensors
     Execute Command In Terminal    sensors-detect --auto
     ${timer}=    Convert To Integer    0
-    FOR    ${i}    IN RANGE    (${TEMPERATURE_TEST_DURATION} / ${TEMPERATURE_TEST_MEASURE_INTERVAL}) + 1
+    @{temperature_list}=    Create List
+    ${sum}=    Convert To Integer    0
+    FOR    ${i}    IN RANGE    (${TEMPERATURE_TEST_DURATION} * 60 / ${TEMPERATURE_TEST_MEASURE_INTERVAL}) + 1
         Log To Console    \n ----------------------------------------------------------------
-        Log To Console    ${timer} min.
+        Log To Console    ${timer}s
         ${temperature}=    Get CPU Temperature CURRENT
+        Append To List    ${temperature_list}    ${temperature}
+        ${sum}=    Evaluate    ${sum} + ${temperature}
         Log To Console    Current temperature: ${temperature}°C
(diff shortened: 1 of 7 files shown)


macpijan:
@pkubaj can you adjust the commit name as we typically do in this repo?

E.g.: `cpu-temperature.robot: allow for spikes of higher temp"

dasharo-performance/cpu-temperature.robot
```diff
@@ -124,12 +124,19 @@ CPU Temperature Without Load (Ubuntu 22.04)
...
     Execute Command In Terminal    sensors-detect --auto
     ${timer}=    Convert To Integer    0
+    ${overheating_cnt}=    Convert To Integer    0
     FOR    ${i}    IN RANGE    (${TEMPERATURE_TEST_DURATION} / ${TEMPERATURE_TEST_MEASURE_INTERVAL}) + 1
         Log To Console    \n ----------------------------------------------------------------
         Log To Console    ${timer} min.
         ${temperature}=    Get CPU Temperature CURRENT
         Log To Console    Current temperature: ${temperature}°C
-        Should Be True    ${temperature} < ${MAX_CPU_TEMP}
+        IF    ${temperature} > ${MAX_CPU_TEMP}
```
macpijan, line 133: @pkubaj I wonder what's the real problem is, and if it really fixes this.

It feels to me that the whole measurement procedure is not perfect, as we measure CPU temperature, which can rapidly change, only in 1 minute intervals, and we take the current value (not e.g. the mean value from last minute).

pkubaj: Yes, the measurement procedure is not perfect, but that's how the CPU temperature monitoring works, that is it checks the current temperature. Checking the mean value from last minute is one thing, but if we're checking the mean, why stop at 1 minute when the test lasts an hour? We could just read the temperature 3600 times and get the mean after one hour, which should be much simpler to implement (provided that Robot supports arrays with 3600 elements).

pkubaj:
Run time tested on V1210.

dasharo-performance/cpu-temperature.robot
```diff
@@ -143,12 +149,18 @@ CPU Temperature After Stress Test (Ubuntu 22.04)
     Execute Command In Terminal    sensors-detect --auto
     Stress Test    ${TEMPERATURE_TEST_DURATION}m
```
mkopec, line 150: This should be updated, otherwise the tests will run for 3600 minutes, right?
```suggestion
    Stress Test    ${TEMPERATURE_TEST_DURATION}s
```

mkopec (approved):

mkopec (changes requested):
bah, misclick

dasharo-performance/cpu-temperature.robot
mkopec: Thanks

PR #288, opened 2024-05-07
Watchdog test (reset after kernel crash)

PLangowski:


dasharo-compatibility/watchdog.robot
```diff
@@ -135,3 +135,35 @@ WDT005.001 Watchdog is detected by OS (Ubuntu 22.04)
...
+    Set Option State    ${chipset_menu}    Enable watchdog    ${TRUE}
+    # Refresh menu, in case the watchdog timeout wasn't already available
+    ${chipset_menu}=    Reenter Menu And Return Construction
+    Set Option State    ${chipset_menu}    Watchdog timeout value    300
+    Save Changes And Reset
+    Boot System Or From Connected Disk    ubuntu
+    Login To Linux
+    Switch To Root User
+    # Verify that there are entries from wdt or watchdog in kernel logs
+    ${out}=    Execute Linux Command    dmesg | grep -E 'watchdog|wdt'
```
macpijan, line 157: Point 4 in: https://github.com/Dasharo/open-source-firmware-validation?tab=readme-ov-file#keywords

PLangowski: @macpijan Why is `Execute Linux Command` wrong here? We use it in many tests after logging into Linux.

macpijan: Because we want to achieve a goal of reducing keyword duplication. Adding new tests using duplicated keywords will not help here.

PR #297, opened 2024-05-27
195 tests fail due to missing packages

JanPrusinowski:
Added installation of missing packets.


macpijan:
@JanPrusinowski When closing, please provide a reference on **why** we close it.

PR #305, opened 2024-06-04
Use DCU for UEFI settings on platforms w/o serial console

mkopec:
Introduce libraries for setting options. Tests will use one of these to configure options, and the actual mechanism (serial, DCU...) will be opaque.

### DIFF
--- a/lib/options/dcu.robot
+++ b/lib/options/dcu.robot
@@ -23,9 +23,7 @@ Set UEFI Option
     Execute Linux Command    flashrom -p internal -r coreboot.rom --fmap -i FMAP -i SMMSTORE &> /dev/null
     SSHLibrary.Get File    coreboot.rom    dcu/coreboot.rom
     # TODO error handling
-    Run    cd dcu
-    Run    ./dcu v coreboot.rom --set "${option_name}" --value "${value}"
-    Run    cd ..
+    Run    cd dcu && ./dcu v coreboot.rom --set "${option_name}" --value "${value}"
     SSHLibrary.Put File    dcu/coreboot.rom    coreboot.rom
     Execute Linux Command    flashrom -p internal -w coreboot.rom --fmap -i SMMSTORE --noverify-all &> /dev/null
     Power On
--- a/lib/options/dcu.robot
+++ b/lib/options/dcu.robot
@@ -15,7 +15,6 @@ Set UEFI Option
     Run    git clone https://github.com/Dasharo/dcu
     # TODO: Remove once smmstore support is merged
     Run    cd dcu && git checkout smmstore > /dev/null 2>&1 && cd ..
-    Power On
     Boot System Or From Connected Disk    ubuntu
     Login To Linux
     Switch To Root User
@@ -26,4 +25,8 @@ Set UEFI Option
     Run    cd dcu && ./dcu v coreboot.rom --set "${option_name}" --value "${value}"
     SSHLibrary.Put File    dcu/coreboot.rom    coreboot.rom
     Execute Linux Command    flashrom -p internal -w coreboot.rom --fmap -i SMMSTORE --noverify-all &> /dev/null
-    Power On
+    Execute Reboot Command
+    # Assume we don't have serial to tell us that we've rebooted, so just wait
+    # 20s for shutdown to finish, to prevent subsequent kwds from running before
+    # reboot.
+    Sleep    20s
--- a/lib/options/dcu.robot
+++ b/lib/options/dcu.robot
@@ -4,6 +4,7 @@ Documentation       Library for UEFI configuration using Dasharo Configuration
 ...                 available.
 
 Library             Collections
+Library             OperatingSystem
 Library             String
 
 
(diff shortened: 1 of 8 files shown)


mkopec:
Log from ME disable test suite on NS5x TGL: [log.zip](https://github.com/user-attachments/files/15570175/log.zip)

lib/options/dcu.robot
```diff
@@ -0,0 +1,29 @@
...
+    Power On
+    Boot System Or From Connected Disk    ubuntu
+    Login To Linux
+    Switch To Root User
+    Get Flashrom From Cloud
+    Execute Linux Command    flashrom -p internal -r coreboot.rom --fmap -i FMAP -i SMMSTORE &> /dev/null
+    SSHLibrary.Get File    coreboot.rom    dcu/coreboot.rom
+    # TODO error handling
+    Run    cd dcu && ./dcu v coreboot.rom --set "${option_name}" --value "${value}"
+    SSHLibrary.Put File    dcu/coreboot.rom    coreboot.rom
```
macpijan, line 27: I can see we have `Send File to DUT` kwd, but I don't see one for Getting a file.

lib/options/dcu.robot
```diff
@@ -0,0 +1,29 @@
...
+    [Arguments]    ${option_name}    ${value}
+    Run    git clone https://github.com/Dasharo/dcu
+    # TODO: Remove once smmstore support is merged
+    Run    cd dcu && git checkout smmstore > /dev/null 2>&1 && cd ..
+    Power On
+    Boot System Or From Connected Disk    ubuntu
+    Login To Linux
+    Switch To Root User
+    Get Flashrom From Cloud
+    Execute Linux Command    flashrom -p internal -r coreboot.rom --fmap -i FMAP -i SMMSTORE &> /dev/null
```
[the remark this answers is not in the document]
mkopec: It could work, but if we want to make variables available for modification in OS, we can expose them to https://github.com/rhboot/efivar . I believe the only reason this doesn't work right now is because all Dasharo variables have `EFI_VARIABLE_BOOTSERVICE_ACCESS` attribute set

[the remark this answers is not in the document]
macpijan: Can we use `Execute Command In Terminal` here? We have multiple kwds with similar role, trying to get rid of it.

https://github.com/Dasharo/open-source-firmware-validation?tab=readme-ov-file#keywords

lib/options/dcu.robot
```diff
@@ -0,0 +1,29 @@
...
+
+
+*** Keywords ***
+Set UEFI Option
+    [Documentation]    Set an UEFI option to a value.
+    [Arguments]    ${option_name}    ${value}
+    Run    git clone https://github.com/Dasharo/dcu
+    # TODO: Remove once smmstore support is merged
+    Run    cd dcu && git checkout smmstore > /dev/null 2>&1 && cd ..
+    Power On
```
macpijan, line 18: Do we still need power control, even though we use SSH, not serial?
Can we imagine a scenario with no power control - e.g. laptop freestanding on my desk, just connected to the network?
If we need `Power On`, we still need RTE, right?

lib/options/dcu.robot
macpijan: We can leave it like it is.

lib/options/uefi-setup-menu.robot
```diff
@@ -0,0 +1,34 @@
...
+
+Library             Collections
+Library             String
+Resource            ../bios/menus.robot
+
+
+*** Keywords ***
+Set UEFI Option
+    [Documentation]    Set an UEFI option to a value.
+    ...    TODO: Only works with options following the submenu/submenu/option
```
macpijan, line 13: Do we have an idea which limitations it brings, and how to overcome this? I would prefer if we create and link issues also in the code right away, if we intend to merge a TODO comment.

mkopec: We can probably handle this by checking the length of the path. I'll try to handle this

lib/options/dcu.robot
mkopec: a Sonoff should be enough, as long as `Power on AC` is enabled

lib/options/dcu.robot
```diff
@@ -0,0 +1,29 @@
...
+    Boot System Or From Connected Disk    ubuntu
```
macpijan, line 19: We assume Ubuntu is the first one in boot order here?
Do we have an idea for Windows tests?
In some suites we have a mixup of Ubuntu -> Windows -> Ubuntu, where tests must select which OS to boot over serial.

mkopec: Yeah on SSH this keyword is no-op, on Linux at least we can use efibootmgr to manipulate bootorder / bootnext, no idea about Windows though.

lib/options/dcu.robot
mkopec: Changed to reboot in commit 563fb1e655de14fca030709837f7edf40a603157 . As a side effect the tests take half as much time now :)

lib/options/dcu.robot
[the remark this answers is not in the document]
mkopec: changed in b039c87a3739164c8d73243438a7576bb0924fe0

lib/options/uefi-setup-menu.robot
mkopec: handled in 2b89ab672a6bcf42a4226bb58e67079cdb1bc985

lib/options/dcu.robot
mkopec: Not sure what to do about this at this point. Right now novacustom only has `TESTS_IN_UBUNTU_SUPPORT` selected so it's a valid assumption that Ubuntu will always boot automatically

macpijan: https://github.com/Dasharo/open-source-firmware-validation/issues/313

PR #307, opened 2024-06-06
dasharo-compatibility/cpu-throttling test

filipleple:


dasharo-compatibility/cpu-throttling.robot
```diff
@@ -0,0 +1,63 @@
...
+
+THR002.001 Try to enter a threshold value within the limits and verify in Ubuntu
+    [Documentation]    Verify whether a reasonable throttling threshold will
+    ...    take effect in Ubuntu
+    Skip If    not ${TESTS_IN_UBUNTU_SUPPORT}    THR002.001 not supported
+    Skip
+    Set UEFI Option    CpuThrottlingThreshold    70
+    Boot System Or From Connected Disk    ubuntu
+    Login To Linux
+    Execute Command In Terminal    stress-ng -c 16&
```
macpijan, line 58: For some platforms 16 is optimal, for some not.
I guess $(nproc) gives some more or less sane default.
Don't we have some kwd for stress test? It would be nice to have.

lib/options/dcu.robot
```diff
@@ -19,6 +19,7 @@ Set UEFI Option
     Switch To Root User
     Get Flashrom From Cloud
     Execute Command In Terminal    flashrom -p internal -r coreboot.rom --fmap -i FMAP -i SMMSTORE &> /dev/null
+    Execute Command In Terminal    chmod 666 coreboot.rom
```
macpijan, line 22: :smiling_imp:

macpijan: But why it was needed?

filipleple: @macpijan whenever 
* `coreboot.rom` doesn't exist yet
* you run sudo su && read with flashrom to `coreboot.rom`

`coreboot.rom` gets created with root only access, and there's `Permission Denied` trouble later on with modifying and sending it

macpijan (changes requested):
Why do we change submodules revisions?

dasharo-compatibility/cpu-throttling.robot
filipleple: @macpijan fixed [here](https://github.com/Dasharo/open-source-firmware-validation/pull/307/files#diff-3ceac275618cfdb4ed8088f5520173f94970e097d2bd23c6ed2ec17a5bea33eaR57) to use predefined keyword

filipleple:
removed the unnecessary submodule update

macpijan (approved):

PR #309, opened 2024-06-07
Make tests use dcu

JanPrusinowski:


dasharo-security/smm-bios-write-protection.robot
```diff
@@ -39,12 +39,7 @@ SMM001.001 SMM BIOS write protection enabling (Ubuntu 22.04)
     ...    during the attempt of firmware flashing information about the
     ...    SMM protection is returned.
     Skip If    not ${TESTS_IN_UBUNTU_SUPPORT}
-    Power On
-    ${setup_menu}=    Enter Setup Menu Tianocore And Return Construction
-    ${dasharo_menu}=    Enter Dasharo System Features    ${setup_menu}
-    ${network_menu}=    Enter Dasharo Submenu    ${dasharo_menu}    Dasharo Security Options
-    Set Option State    ${network_menu}    Enable SMM BIOS write    ${TRUE}
-    Save Changes And Reset
+    Set UEFI Option    SmmBwp    Enabled
```
mkopec, line 42: This is the one test we can't do with this tool. Enabling this option prevents DCU from working correctly, so we need to skip this test on platforms where `${OPTIONS_LIB}` is set to `dcu`

mkopec (approved):

variables.robot
```diff
@@ -140,7 +140,7 @@ ${OS_UBUNTU}=               ubuntu
 ...                         platform_vendor=protectli
 # 3mdeb Protectli vp4670_2 ---------------------------------------------------
```
mkopec, line 141: I guess if the platform config changed, so should this comment

PR #276, opened 2024-04-24
Add coreboot/EC update tests

m-iwanicki:

PR #317, opened 2024-06-19
lib: dts-lib.robot: Fix Boot Dasharo Tools Suite keyword failures on laptops

m-iwanicki:
On laptops `Read From Terminal Until    Enter an option:` should fail in 20 seconds and then try to blindly start ssh server and connect via ssh.
On other platforms keyword should succeed much faster.
Not sure if waiting 20 seconds is ok, too long or too short

### DIFF
--- a/lib/dts-lib.robot
+++ b/lib/dts-lib.robot
@@ -1,3 +1,7 @@
+*** Settings ***
+Resource    terminal.robot
+
+
 *** Keywords ***
 Boot Dasharo Tools Suite Via IPXE Shell
     Enter IPXE
@@ -33,6 +37,8 @@ Boot Dasharo Tools Suite
             ${ipxe_menu}=    Get IPXE Boot Menu Construction
             Enter Submenu From Snapshot    ${ipxe_menu}    Dasharo Tools Suite
             Set DUT Response Timeout    5m
+            Read From Terminal Until    .cpio.gz...
+            Read From Terminal Until    ok
         END
     ELSE
         FAIL    Unknown or improper connection method: ${dts_booting_method}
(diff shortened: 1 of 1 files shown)


macpijan:
Do we have some test log?
How we have verified it works correctly?

m-iwanicki:
@macpijan Only tested on MSI to check if I didn't mess anything up, couldn't test on laptop as lab testing station wasn't working. On MSI:

```
robot -v snipeit:yes -L TRACE -v rte_ip:192.168.10.199 -v device_ip:192.168.10.80 -v config:msi-pro-z690-a-wifi-ddr4 -t "DTS003.001*" dasharo-compatibility/dasharo-tools-suite.robot
Dasharo-Tools-Suite
DTS003.001 DTS option reboot DUT works correctly :: This test aims... | PASS |
Dasharo-Tools-Suite                                                   | PASS |
1 test, 1 passed, 0 failed
```

And now that I'm looking at it I thought that I managed to fix this problem in this PR https://github.com/Dasharo/open-source-firmware-validation/pull/276. It was merged 2 weeks ago at contained similar [fix](https://github.com/Dasharo/open-source-firmware-validation/pull/276/files#diff-0ca58a364903f744bc9458315dbf5ccc48c37f510ab670103bb67f789341868b), this one is just faster on non laptop versions and should work better on iPXE if downloading takes longer due to slower network connection.
But I'm not sure how, possibly by force push, that whole PR is not on `develop`. Last change to `dts-lib.robot` was 4 months ago.

macpijan:
@m-iwanicki What do you suggest? We stll need this MR, or we should close it?

m-iwanicki:
@macpijan I think this way is more robust now and less prone to errors.
Without changes in this PR, keyword waits for 10 seconds after booting before trying to blindly enable SSH.
In case of iPXE, download + boot might take longer than that.
With this PR waiting only starts after DTS finished downloading and starts booting. Additionally in case of platforms that have output from Linux, keyword won't have to wait 20 seconds but only until it reads DTS menu.

PR #318, opened 2024-06-20
Enable more tests for QEMU

krystian-hebel:


scripts/regression-qemu.sh
```diff
@@ -11,11 +11,17 @@ compatibility_tests=(
...
   "dasharo-compatibility/network-boot.robot"
+  "dasharo-compatibility/dmidecode.robot"
+  # FIXME: unsafe to enable https://github.com/Dasharo/dasharo-issues/issues/887
+  #"dasharo-compatibility/reset-to-defaults.robot"
 )
 
 security_tests=(
   "dasharo-security/network-stack.robot"
   "dasharo-security/secure-boot.robot"
+  # TODO: test if this is safe to enable
```
macpijan, line 22: I hope we can remove the regression-qemu script and use a single regression.sh for all platforms. But I will take it later: https://github.com/Dasharo/open-source-firmware-validation/issues/320

krystian-hebel: This reminded me that I've tested it and forgot to remove `TODO`.

macpijan:
If we replace URL from: https://github.com/Dasharo/open-source-firmware-validation/pull/318/files#diff-69e6355529b2609a8ddcccd04f8cd299d0fe65c19c246c853a31b9eb96fe6818R194 with the build to be tested, it should pass?

krystian-hebel:
It should, although I haven't tested it, I always had coreboot.rom in place when running `qemu-run.sh`.

krystian-hebel:
Secure Boot test resets the platform before settings are actually saved to flash in suite setup (causing first test to fail if platform hasn't SB in default state on entry) and teardown (which causes following test to fail). I'm working on fix/workaround.

scripts/ci/qemu-run.sh
```diff
@@ -176,37 +173,36 @@ case "${ACTION}" in
...
-if [ ! -f "OVMF_CODE.fd" ] || [ ! -f "OVMF_VARS.fd" ]; then
-    echo "The required files OVMF_CODE.fd and OVMF_VARS.fs are missing."
-    echo "Downloading files from the server..."
-    wget -O ./OVMF_CODE.fd https://github.com/Dasharo/edk2/releases/latest/download/OVMF_CODE_RELEASE.fd
-    wget -O ./OVMF_VARS.fd https://github.com/Dasharo/edk2/releases/latest/download/OVMF_VARS_RELEASE.fd
+# Check for the existence of coreboot.rom file
+if [ ! -f "coreboot.rom" ]; then
+    echo "The required file coreboot.rom is missing."
+    echo "Downloading from the server..."
+    wget -O ./coreboot.rom https://github.com/Dasharo/edk2/releases/latest/download/coreboot.rom
```
krystian-hebel, line 194: FIXME: this will be on coreboot repo, not edk2!

macpijan: > It should, although I haven't tested it, I always had coreboot.rom in place when running qemu-run.sh.

Then maybe drop some link with "RC" build and we will find out?

macpijan: Added here: 694812854b83670eede86338eb8822bb76f550ec

Let's see

macpijan: GH runner fails with

```
 Could not access KVM kernel module: Permission denied
qemu-system-x86_64: failed to initialize kvm: Permission denied
```

I wonder why, the kvm switch was also there before

macpijan: But just a few minutes ago it used to start OVMF binaries just fine: https://github.com/Dasharo/open-source-firmware-validation/actions/runs/9618984974/job/26534054299#step:4:126

Ideas?

krystian-hebel: Interesting. Installation of KVM probably requires a restart, and because this PR changed the way CI runs it is possible that it uses a new container, while previous ones reused existing one. I've tried restarting the CI to check if next run will pass, but it still fails.

I've moved KVM parameter from `QEMU_PARAMS_OS` to `QEMU_PARAMS_BASE`. I've also fixed an issue that dropped display mode [here](https://github.com/Dasharo/open-source-firmware-validation/pull/318/files#diff-69e6355529b2609a8ddcccd04f8cd299d0fe65c19c246c853a31b9eb96fe6818R178), this is the only difference I can think of. Not sure if `-nographic` may cause KVM permission error :man_shrugging:

krystian-hebel: Oh, wait, CI starts it as `./scripts/ci/qemu-run.sh nographic firmware &`, so it didn't use KVM earlier. This can mess up SMM operations
...

macpijan: @krystian-hebel I let myself to rebase, I suspect that is why the 2nd run fails.

krystian-hebel: Other than changing this link to non-rc release I think it is ready.

m-iwanicki:
@krystian-hebel Not sure if it's problem or not but I noticed that I can't set active PCR banks in BIOS when running QEMU from your branch. In `TCG2 Configuration` menu both `TPM2 Active PCR Hash Algorithm` and `TPM Hardware Supported Hash Algorithm` is `[Unknown]` and there is no option to choose active PCR banks.
I can set sha256 banks to be active by using `tpm2_pcrallocate` in Ubuntu and from what I can see after doing it once both SHA1 and SHA256 are active even after starting QEMU again (shouldn't all configuration reset to default between launches?).
In image below top column is QEMU from this branch and below is QEMU from `develop`
tpm-qemu

macpijan:
Related to: https://github.com/Dasharo/dasharo-issues/issues/521 ?

m-iwanicki:
> Related to: [Dasharo/dasharo-issues#521](https://github.com/Dasharo/dasharo-issues/issues/521) ?

@macpijan not exactly because I don't even have possibility to change active PCR bank. In issue you linked it was possible to change active PCR in BIOS but changes weren't applied.

SergiiDmytruk:
Looks like with rebased EDK and DasharoPayloadPkg `Tpm2GetCapabilityPcrs()` succeeds in PEI but fails in DXE with "Unsupported" error, but works fine with OvmfPkg.  Maybe coreboot leaves TPM is some weird state.

krystian-hebel:
OVMF used LAPIC for a timer, DasharoPkg uses HPET. This makes booting much longer and may also cause timeouts, because the only source of interrupts that is enabled at some points is the RTC, once a second. Using `OvmfPkg/LocalApicTimerDxe/LocalApicTimerDxe.inf` instead of HPET makes it boot faster, but it also makes the delays shorter, up to a point where tests can't traverse iPXE menu before autoboot, which should normally take 3 seconds. It may be controlled by modifying `gEfiMdePkgTokenSpaceGuid.PcdFSBClock` (MdePkg defines it as 200M, while most of OVMF platforms use 1G), however I haven't tested how it would work on real hardware.

krystian-hebel:
> after starting QEMU again

@m-iwanicki how do you start it, using `qemu-run.sh` or manually? Can you check if TPM settings are restored after removing `/tmp/osfv/tpm`?

m-iwanicki:
> how do you start it, using qemu-run.sh or manually? Can you check if TPM settings are restored after removing /tmp/osfv/tpm?

@krystian-hebel By using `qemu-run.sh`. Yes, after deleting `/tmp/osfv/tpm` settings go back to default (all PCR banks enabled).
Just checked and the same thing happens on `develop` branch, I just assumed that state should return to same default state each time QEMU is started.

SergiiDmytruk:
@krystian-hebel I've added https://github.com/Dasharo/edk2/pull/129/commits/b5eca829af8a64b9bbe739c59683b66b6b89129b that resolves the issue pointed out by @m-iwanicki above.

lib/secure-boot-lib.robot
```diff
@@ -82,6 +82,8 @@ Reset To Default Secure Boot Keys
     Enter Submenu From Snapshot    ${advanced_menu}    Reset to default Secure Boot Keys
     Read From Terminal Until    Are you sure?
     Press Enter
+    # QEMU needs a lot of time, why?
```
macpijan, line 85: @krystian-hebel Can it be the same reason as long boot time, or not really?
If not, we would need a second issue for that.

macpijan (commented):
I think we can merge it if we clarify this: https://github.com/Dasharo/open-source-firmware-validation/pull/318/files#r1653302103

If firmware link needs to be changed, we can change it in another MR. But I guess we can also update the release artifacts

lib/secure-boot-lib.robot
krystian-hebel: It is the same, it worked fine with LAPIC. I'd still leave some delay, there are cases where this keyword is immediately followed by a power-off, I'm surprised it didn't cause any issues earlier.

PR #326, opened 2024-06-24
Extend measured boot tests

m-iwanicki:


dasharo-security/measured-boot.robot
```diff
@@ -18,25 +18,375 @@ Resource            ../keys.robot
...
     Boot System Or From Connected Disk    ubuntu
     Login To Linux
     Switch To Root User
     Detect Or Install Package    tpm2-tools
-    ${out_tpm2_pcrread}=    Execute Command In Terminal    tpm2_pcrread
-    ${matching_lines}=    Get Lines Matching Regexp    ${out_tpm2_pcrread}    .*([0-3]) :.*
-    Should Not Be Empty    ${matching_lines}
-    Should Not Contain    ${matching_lines}    0x0000000000000000000000000000000000000000
-    Should Not Contain    ${matching_lines}    0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
+    Execute Command In Terminal    systemctl disable secureboot-db.service
```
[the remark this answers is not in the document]
m-iwanicki: @macpijan Added in [11a910f](https://github.com/Dasharo/open-source-firmware-validation/pull/326/commits/11a910f9d28863f0f9d3cefa380df3f17ca2767f)

dasharo-security/measured-boot.robot
```diff
@@ -18,25 +18,376 @@ Resource            ../keys.robot
...
-    ...    measurements are stored into the TPM.a.
-    Skip If    not ${MEASURED_BOOT_SUPPORT}    MBO001.001 not supported
-    Skip If    not ${TESTS_IN_UBUNTU_SUPPORT}    MBO001.001 not supported
+    ...    measurements are stored into the TPM.
+    Execute Command In Terminal    shopt -s extglob
+    ${hashes}=    Execute Command In Terminal
+    ...    grep . /sys/class/tpm/tpm0/pcr-sha@(1|256)/[0-3]
+    Should Not Contain Any    ${hashes}
+    ...    0000000000000000000000000000000000000000
+    ...    FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
```
SergiiDmytruk, line 39: Not an imminent danger of a false positive, but I'd check for sha1 and sha256 separately to not assume that no sha256 hash has a sequence of 40 0s or Fs inside of its 64-character value.

dasharo-security/measured-boot.robot
```diff
@@ -18,25 +18,376 @@ Resource            ../keys.robot
...
+    ...    No such file or directory
+    ...    error
+    ...    ignore_case=${TRUE}
+
+MBO002.001 Check if event log PCRs match actual values (Ubuntu 22.04)
```
SergiiDmytruk, line 44: Not sure that `(Ubuntu 22.04)` here and above are necessary if it's the only OS that's being checked and we don't have anything specific to this particular OS in the test case.

dasharo-security/measured-boot.robot
```diff
@@ -18,25 +18,376 @@ Resource            ../keys.robot
...
+    ...    actual PCRs values
+    ${tpm2_eventlog}=    Execute Command In Terminal
+    ...    tpm2_eventlog /sys/kernel/security/tpm0/binary_bios_measurements
+    Should Not Contain    ${tpm2_eventlog}    ERROR: Unable to run tpm2_eventlog
+    FOR    ${algo}    IN    sha1    sha256
+        ${eventlog_pcrs}=    Get PCRs From Eventlog    ${tpm2_eventlog}    ${algo}
+        FOR    ${pcr_element}    IN    @{eventlog_pcrs}
+            ${pcr}    ${hash}=    Split String    ${pcr_element}    separator=:
+            ${sha_hash}=    Execute Command In Terminal
+            ...    cat /sys/class/tpm/tpm0/pcr-${algo}/${pcr}
```
SergiiDmytruk, line 55: Commands here and in other places are specific to TPM2.  Are we intentionally not checking TPM1?

dasharo-security/measured-boot.robot
```diff
@@ -18,25 +18,376 @@ Resource            ../keys.robot
...
 *** Test Cases ***
 MBO001.001 Measured Boot support (Ubuntu 20.04)
     [Documentation]    Check whether Measured Boot is functional and
-    ...    measurements are stored into the TPM.a.
-    Skip If    not ${MEASURED_BOOT_SUPPORT}    MBO001.001 not supported
-    Skip If    not ${TESTS_IN_UBUNTU_SUPPORT}    MBO001.001 not supported
+    ...    measurements are stored into the TPM.
+    Execute Command In Terminal    shopt -s extglob
+    ${hashes}=    Execute Command In Terminal
+    ...    grep . /sys/class/tpm/tpm0/pcr-sha@(1|256)/[0-3]
```
SergiiDmytruk, line 36: A comment that `grep` is used to get file names in the output would be helpful (there is another use at the bottom).

dasharo-security/measured-boot.robot
```diff
@@ -18,25 +18,376 @@ Resource            ../keys.robot
...
+
+MBO004.001 Changing Dasharo network boot settings changes only PCR-1
+    [Documentation]    Check if changes to Dasharo security settings influence PCR-1
+    ...    value and only PCR-1
+    Skip If    not ${TESTS_IN_FIRMWARE_SUPPORT}    Tests in firmware are not supported
+    Skip If    not ${DASHARO_NETWORKING_MENU_SUPPORT}    Tests in Dasharo Networking Menu are not supported
+    Power On
+    Boot System Or From Connected Disk    ubuntu
+    Login To Linux
+    Switch To Root User
```
SergiiDmytruk, line 107: Is there a reason to reset here and in 2 tests below?  If it's to guard against dbx changes, maybe test setup keyword should check current state of the service and if it's on disable it and reset SB state.  This way no test case should worry about dbx and there won't be unnecessary reboots.

dasharo-security/measured-boot.robot
```diff
@@ -18,25 +18,376 @@ Resource            ../keys.robot
...
+    ${default_pcr_state}=    Get PCRs State From Linux
+
+    Restore Secure Boot Defaults
+    Reset To Defaults Tianocore
+    Save Changes And Reset
+
+    Boot System Or From Connected Disk    ubuntu
+    Login To Linux
+    Switch To Root User
```
SergiiDmytruk, line 218: This group of lines repeats a lot, a separate keyword for it will noticeably shorten the tests.

dasharo-security/measured-boot.robot
```diff
@@ -18,25 +18,376 @@ Resource            ../keys.robot
...
+MBO005.002 Multiple reset to defaults results in identical measurements
+    [Documentation]    Resetting Dasharo configuration twice will give the same
+    ...    PCRs measurements
+    Skip If    not ${TESTS_IN_FIRMWARE_SUPPORT}    Tests in firmware are not supported
+    Skip If    not ${RESET_TO_DEFAULTS_SUPPORT}    Tests with "Reset to defaults" are not supported
+    ${default_hashes}=    Get Default PCRs State
+
+    Restore Secure Boot Defaults
+    Reset To Defaults Tianocore
+    Save Changes And Reset
```
SergiiDmytruk, line 231: This triplet seems to appear 4 times and deserves a local keyword as well.

dasharo-security/measured-boot.robot
```diff
@@ -18,25 +18,376 @@ Resource            ../keys.robot
...
+    Boot System Or From Connected Disk    ubuntu
+    Login To Linux
+    Switch To Root User
+    FOR    ${pcr_hash}    IN    @{default_hashes}
+        ${pcr}    ${hash}=    Split String    ${pcr_hash}    separator=:
+        ${new_hash}=    Execute Command In Terminal    cat ${pcr}
+        Should Be Equal    ${hash}    ${new_hash}
+    END
+
+MBO005.003 Identical configuration results in identical measurements
```
SergiiDmytruk, line 242: Should this be `MBO006.001` and the next one `MBO006.002`.

dasharo-security/measured-boot.robot
```diff
@@ -18,25 +18,376 @@ Resource            ../keys.robot
...
+MBO004.001 Changing Dasharo network boot settings changes only PCR-1
+    [Documentation]    Check if changes to Dasharo security settings influence PCR-1
+    ...    value and only PCR-1
+    Skip If    not ${TESTS_IN_FIRMWARE_SUPPORT}    Tests in firmware are not supported
+    Skip If    not ${DASHARO_NETWORKING_MENU_SUPPORT}    Tests in Dasharo Networking Menu are not supported
+    Power On
+    Boot System Or From Connected Disk    ubuntu
+    Login To Linux
+    Switch To Root User
+    @{default_hashes}=    Get PCRs State From Linux
```
SergiiDmytruk, line 108: Here and at the start of other tests.  Is this really default state or maybe the current one?  If it must be default, I guess `Get Default PCRs State` should be used instead.

dasharo-security/measured-boot.robot
m-iwanicki: At least on QEMU `tpm2_eventlog` works with TPM 1.2 `binary_bios_measurements`. It would need to be tested on real platform with TPM 1.2 to be sure.

dasharo-security/measured-boot.robot
m-iwanicki: In this and similar tests I:
* login to linux and save current PCR values
* restart and change dasharo settings
* restart and login to linux to compare PCR values

I'm not sure If I could do this test with less restarts.

dasharo-security/measured-boot.robot
m-iwanicki: fixed in [802f7e5](https://github.com/Dasharo/open-source-firmware-validation/pull/326/commits/802f7e57553d1ff09c92255ae050c4d06e75f8b6)

dasharo-security/measured-boot.robot
m-iwanicki: Removed in [aee905e](https://github.com/Dasharo/open-source-firmware-validation/pull/326/commits/aee905e0b6173c965ad1301559190fcc657c8da8) as currently whole test suite is run in Ubuntu

dasharo-security/measured-boot.robot
m-iwanicki: Added comment in [f6fdd1e](https://github.com/Dasharo/open-source-firmware-validation/pull/326/commits/f6fdd1eca99d3349aa0be11e35aa2a58307a4141) and added `-H` flag in probably impossible case when there is only one PCR in `pcr-sha` directory

dasharo-security/measured-boot.robot
m-iwanicki: Changed in [d4ff4f0](https://github.com/Dasharo/open-source-firmware-validation/pull/326/commits/d4ff4f0868ebc8c13979ac1e53b879138aa269ad). I'm not sure if we have documentation somewhere explaining naming scheme.

dasharo-security/measured-boot.robot
m-iwanicki: Changed in [d4ff4f0](https://github.com/Dasharo/open-source-firmware-validation/pull/326/commits/d4ff4f0868ebc8c13979ac1e53b879138aa269ad) to `hashes_before_changes` as `current_hashes` wouldn't be current couple lines later.

dasharo-security/measured-boot.robot
m-iwanicki: fixed in [c458e30](https://github.com/Dasharo/open-source-firmware-validation/pull/326/commits/c458e304766e00bd12c3bca95bcac0d918229bb7), you are right there was a lot of repetition.

dasharo-security/measured-boot.robot
m-iwanicki: fixed in [c458e30](https://github.com/Dasharo/open-source-firmware-validation/pull/326/commits/c458e304766e00bd12c3bca95bcac0d918229bb7)

dasharo-security/measured-boot.robot
SergiiDmytruk: My point is that at the beginning each of these tests Linux is already running and you have the shell, so why restart it before saving PCR values?

dasharo-security/measured-boot.robot
m-iwanicki: I checked. In trenchboot:

```shell
root@tb:~# ls /sys/class/tpm/tpm0/pcr*
/sys/class/tpm/tpm0/pcrs

/sys/class/tpm/tpm0/pcr-sha1:
0   1   10  11  12  13  14  15  16  17  18  19  2   20  21  22  23  3   4   5   6   7   8   9
root@tb:~# cat /sys/class/tpm/tpm0/tpm_version_major
1
```

m-iwanicki: In ubuntu it's the same:

```shell
ubuntu@3mdeb:~$ ls /sys/class/tpm/tpm0/pcr*
/sys/class/tpm/tpm0/pcrs

/sys/class/tpm/tpm0/pcr-sha1:
0  10  12  14  16  18  2   21  23  4  6  8
1  11  13  15  17  19  20  22  3   5  7  9
```

dasharo-security/measured-boot.robot
SergiiDmytruk: > I'm not sure if we have documentation somewhere explaining naming scheme.

I couldn't find anything either.

dasharo-security/measured-boot.robot
m-iwanicki: Because if previous test fails somewhere in middle while it's in BIOS then next test would have no chance of passing.

dasharo-security/measured-boot.robot
SergiiDmytruk: My bad, for some reason I was sure there is only `.../pcrs` for TPM1.

dasharo-security/measured-boot.robot
SergiiDmytruk: Then this is kind of a workaround rather than integral part of the test.  A comment that this guards against messed up state after the previous test would be helpful.

dasharo-security/measured-boot.robot
SergiiDmytruk: One more thing though: the loop tries to use `sha256`, won't it fail for TPM1?

dasharo-security/measured-boot.robot
m-iwanicki: Does it? I always thought that it was default that tests should be as self-contained as possible, especially since I don't think we have some default way to describe dependencies between test (check/skip if test x failed).

SergiiDmytruk: Self-contained tests are great, but rebooting takes time and now tests are inconsistent: some start with a reset and some don't which made me wonder why the former need a reset.

m-iwanicki: The only one that doesn't start with reset are the first 2 that only call commands in Linux. The rest are restarted either explicitly by Power On keyword or inside other keywords.

dasharo-security/measured-boot.robot
m-iwanicki: No, you can check in logs in this comment: https://github.com/Dasharo/open-source-firmware-validation/pull/326#issuecomment-2214214232
Platform had only sha256. Keyword will return empty list if no shaX values are in event log.

dasharo-security/measured-boot.robot
```diff
@@ -18,25 +18,350 @@ Resource            ../keys.robot
...
+        ${new_hash}=    Execute Command In Terminal    cat ${pcr}
+        Should Be Equal    ${hash}    ${new_hash}
+    END
+
+MBO006.002 Identical configuration after reset results in identical measurements
+    [Documentation]    Check if same configuration state achieved by resetting
+    ...    state to default results in same PCR values
+    Skip If    not ${TESTS_IN_FIRMWARE_SUPPORT}    Tests in firmware are not supported
+    Skip If    not ${DASHARO_NETWORKING_MENU_SUPPORT} and not ${USB_MASS_STORAGE_SUPPORT}
+    ...    Platform doesn't support neither Security or USB menu tests
```
SergiiDmytruk, line 254:
```suggestion
    ...    Platform supports neither Networking nor USB menu tests
```

SergiiDmytruk:
> MBO004.002 might have something to do with the fact that test disables USB Mass Storage and platform had connected bootable USB?

PCR-8 shouldn't be touched by firmware.  But GRUB could be extending a module related to USB ([docs](https://www.gnu.org/software/grub/manual/grub/html_node/Measured-Boot.html)) or maybe it enumerates available drives.  I don't see an obvious cause in the log.  If there is no more neutral option to use, could exclude PCR-8 or check only PCR-0..7 after comparing event logs to confirm the cause of the failure.

> If yes then other test that enables/disables network boot might have similar problems?

Not necessarily, this might not apply to network boot.  Depends on what's the actual cause.

m-iwanicki:
>  If there is no more neutral option to use

Tried to find option that would be available in most platforms. Maybe lock BIOS boot medium would be better

> I don't see an obvious cause in the log

In the eventlog there is couple events with different `hd` number:

```diff
+ grub_cmd: [ -z (hd1,gpt6)/boot/grub -a -f (hd1,gpt6)/boot/grub/custom.cfg ]
- grub_cmd: [ -z (hd2,gpt6)/boot/grub -a -f (hd2,gpt6)/boot/grub/custom.cfg ]
```

So i guess I would have to change it.

m-iwanicki:
@SergiiDmytruk From what I read in [pcr-measurements.md](https://github.com/Dasharo/docs/blob/1022511ede55ff4dc935554f8c84f095ebffb3fc/docs/kb/pcr-measurements.md#pcr-measurements-1) `EV_EFI_VARIABLE_DRIVER_CONFIG` should only be measured if `Secure Boot` is on?
While this test passed I guess that's because of shim so even with SB enabled it'll still succeed but for wrong reasons?
I'm not sure how to test this otherwise.
I could boot DTS but booting via iPXE doesn't work with SB enabled (at least on QEMU) so I guess it'd have to be via USB with DTS that can be booted when SB is enabled. That would probably limit platforms on which this test can be done.

SergiiDmytruk:
Actually both old and new version of `pcr-measurements.md` seem to be wrong...  Those variables are unconditionally measured for TPM 2.0 during boot, they are also measured when they are set regardless of TPM version.  I was testing with TPM 1.2, saw unexpected event logs and confused myself.  Maybe TCG doesn't require those measurement for TPM 1.2 (https://trustedcomputinggroup.org/wp-content/uploads/TCG_PCClient_PFP_r1p05_v23_pub.pdf might apply only to TPM 2.0).  I guess adding them wouldn't hurt if it will help with tests and measurements on setting  variables might already be "unnecessary" for TPM 1.2 case.

m-iwanicki:
@SergiiDmytruk Rebased and changed USB test to use SMM write protection.
Now the only test that fails is `MBO002.001 Check if event log PCRs match actual values` tested on newest binary from https://github.com/Dasharo/coreboot/pull/517 on Protectli v2410.

```
Measured-Boot

Checking if tpm2-tools is installed...

Package tpm2-tools is not installed

Installing required package (tpm2-tools)...

Required package (tpm2-tools) installed successfully
MBO001.001 Measured Boot support :: Check whether Measured Boot is... | PASS |
MBO002.001 Check if event log PCRs match actual values :: Check wh... | FAIL |
'0x0cb9f33ccb957f8b53944617dfe746bb4788be97f576b913b67adb021ed8e072' does not contain '55a2caae3e70018dfa2d6c6571de2ad2fb98ea2099d64e0f667ab7a441618553'
MBO003.001 Changing Secure Boot certificate changes only PCR-7 :: ... | PASS |
MBO004.001 Changing Dasharo network boot settings changes only PCR... | SKIP |
Tests in Dasharo Networking Menu are not supported
MBO004.002 Changing Dasharo security settings changes only PCR-1 :... | FAIL |
67D3ECDCD7430B68FBAF990F885ABDF6D1B359F0 == 67D3ECDCD7430B68FBAF990F885ABDF6D1B359F0
MBO004.003 Changing Dasharo APU settings changes only PCR-1 :: Che... | SKIP |
Tests in Dasharo APU Menu are not supported
MBO005.001 Flashing firmware and reset to defaults results in same... | PASS |
MBO005.002 Multiple reset to defaults results in identical measure... | PASS |
MBO006.001 Identical configuration results in identical measuremen... | PASS |
MBO006.002 Identical configuration after reset results in identica... | PASS |
Measured-Boot                                                         | FAIL |
10 tests, 6 passed, 2 failed, 2 skipped
```

[log.tar.gz](https://github.com/user-attachments/files/16425951/log.tar.gz)

`MBO004.002` was tested on current fw that's why it failed, after retesting on your version:

```
MBO002.001 Check if event log PCRs match actual values :: Check wh... | FAIL |
'0x5398290bdf51282424884831e17e1f730ee6c8f5' does not contain 'b2a83b0ebf2f8374299a5b2bdfc31ea955ad7236'
MBO004.002 Changing Dasharo security settings changes only PCR-1 :... | PASS |
Measured-Boot                                                         | FAIL |
2 tests, 1 passed, 1 failed
```

`MBO002.001` fails when comparing `pcr-sha1/2` pcr.

SergiiDmytruk:
> `MBO002.001` fails when comparing `pcr-sha1/2` pcr.

That's because `tpm2_eventlog` doesn't know that it should skip `0100...` log entries.  I referred to them initially as "fake digests" and later commented in https://github.com/Dasharo/docs/pull/838#discussion_r1679795457 that we might want to not add such entries in EDK so tests wouldn't need to take them into account, but the conclusion is that they stay.  I see two ways to address this:
 1. Allow some PCR banks to not match if `0100...` is present in event log (can happen only for TPM2).
 2. Implement computing expected PCR values that skips such fake digests.  Iterating over YAML (output of `tpm2_eventlog`) in Python and performing PCR extend operation (concatenation of current PCR value (all zeros initially) and a digest followed by hashing of the result) sounds doable.  I don't think there is anything already available to do this (like an option for `tpm2_eventlog`).

pietrushnic:
@SergiiDmytruk, shouldn't we bring such issues to the attention of people who have 10+ years of experience with TCG?

SergiiDmytruk:
@pietrushnic, the issue is caused by coreboot extending only a single PCR bank and EDK having to deal with it somehow, TCG isn't at fault here.  Initial version of upstreamed coreboot patches that add TCG logs allowed extending multiple banks, but one of the reviewers didn't want to "complicate the code", so that part got dropped.

macpijan:
As author of [this comment](https://github.com/Dasharo/docs/pull/838#discussion_r1681148102), do you have a suggestion on how the test scenario should look like @krystian-hebel ?

I think we are discussing the `MBO002.001` scenario right now.

krystian-hebel:
I would have to see the event log to be sure, but to me it looks that scenario is good, it is coreboot that doesn't abide by the [TCG PC Client Platform Firmware Profile Specification](https://trustedcomputinggroup.org/wp-content/uploads/TCG_PCClient_PFP_r1p05_v23_pub.pdf), namely `3.3.1 Collection and Reporting of Measurements`:

> 4. Unless TPM is hidden, integrity measurements made by the Platform Firmware that are extended MUST be
extended into all allocated banks for a PCR, see Section 10.1 Introduction.

Section 10.1 is about the event log (so it applies to edk2 rather than coreboot), which includes:

> 8. There SHALL be a Hash algorithm in the TCG PCClientPCREvent structure for all allocated PCR banks.

So edk2 does the sane thing here, and mismatch between PCR and event log shows that something fishy is going on.

There are two possible solutions I can think of, neither of them involves changes to the test:
- make coreboot work with multiple allocated banks (makes upstream unhappy), or
- enforce that only one PCR bank can be allocated at any given time (decreases security against hash collisions).

pietrushnic:
> make coreboot work with multiple allocated banks (makes upstream unhappy)

When did we start to worry about that? Is the whole upstream happy about Dasharo's existence?

krystian-hebel:
> > make coreboot work with multiple allocated banks (makes upstream unhappy)
> 
> When did we start to worry about that? Is the whole upstream happy about Dasharo's existence?

https://docs.dasharo.com/newcomers/#dasharo-contribution

> Since Dasharo is based on coreboot and edk2, it's best to contribute directly in the upstream if possible.

Or in our internal documentation: http://wiki.3mdeb.com/products/EX018-dasharo-release-process/#3-implement

pietrushnic:
@krystian-hebel I have no idea what precisely you refer to. Can you be more specific? Ideally, you would quote a sentence, which means we have to make upstream happy because I can't see such a sentence in the document. So maybe we should change the newcomer documentation you are sticking to so it reflects the correct approach because not first time perfect is the enemy of good.

SergiiDmytruk:
After a discussion with @macpijan I've created https://github.com/Dasharo/dasharo-issues/issues/982 to move this forward.  The plan is to limit PCR banks to only one for now (to not postpone landing MeasuredBoot improvements) and make coreboot extend all banks at some later point (the issues gives pointers how to do that).

PR #335, opened 2024-06-27
Network test case

JanPrusinowski:


dasharo-compatibility/network-speed.robot
```diff
@@ -0,0 +1,134 @@
...
+...                 ${BACKSPACE}\#echo $1${ENTER}
+...                 ${BACKSPACE}\#echo $2${ENTER}
+...                 ${BACKSPACE}${ENTER}
+...                 ${BACKSPACE}\# Use ip -d link list to see maxmtu value of interface${ENTER}
+...                 ${BACKSPACE}\# maximum transmission unit (MTU) is a measurement representing the
+...                 ${BACKSPACE}largest data packet that a network-connected device will accept.${ENTER}
+...                 ${BACKSPACE}\#ip link set $1 mtu 9000${ENTER}
+...                 ${BACKSPACE}\#ip link set $2 mtu 9000${ENTER}
+...                 ${BACKSPACE}\# ----${ENTER}
+...                 ${BACKSPACE}${ENTER}
```
miczyg1, line 41:
```suggestion
```

These are comments. Do not really have to be here.

dasharo-compatibility/network-speed.robot
```diff
@@ -0,0 +1,134 @@
...
+...                 ${BACKSPACE}killall iperf3${ENTER}
+...                 ${BACKSPACE}${ENTER}
+...                 ${BACKSPACE}ip netns exec ns_client iperf3 -s &${ENTER}
+...                 ${BACKSPACE}ip netns exec ns_server iperf3 -c 10.1.1.2${ENTER}
+...                 ${BACKSPACE}${ENTER}
+...                 ${BACKSPACE}killall iperf3${ENTER}
+...                 ${BACKSPACE}${ENTER}
+...                 ${BACKSPACE}ip netns del ns_server${ENTER}
+...                 ${BACKSPACE}ip netns del ns_client
```
miczyg1, line 63: It would really be better to use `Execute Command In Terminal`. Otherwise we may have less information in the test log if something goes wrong

platform-configs/include/default.robot
```diff
@@ -24,6 +24,7 @@ ${FLASH_VERIFY_METHOD}=                             ${EMPTY}
 ${WIFI_CARD}=                                       ${EMPTY}
 ${MAX_CPU_TEMP}=                                    ${EMPTY}
 ${INTERNAL_PROGRAMMER_CHIPNAME}=                    Opaque flash chip
+${MIN_ACCEPTED_NET_SPEED}=                          2.32
```
miczyg1, line 27: It shouldn't be here. Some platform have only 1Gb Ethernet ports and some have 2.5Gb Ethernet. So this value will never be reached by some platforms...

mkopec: This value is just the max link speed of a device minus TCP overhead, so we can get this programatically:

```
echo $(($(cat /sys/class/net/enp0s13f0u1u2c2/speed) * 0.93 / 1000))
```

dasharo-compatibility/network-speed.robot
JanPrusinowski: I was thinking that if for some reason this script would disconnect the DUT from the internet - this script would bring it back online. I think that running such test should be executed on the DUT side.

platform-configs/include/default.robot
JanPrusinowski: Didn't use this way because contents of this file for disabled nic is just "-1". But I have used the ```ethtool```. You can check my solution at: fca76a3e13ec001b787f9a90a25930eda2547f2d

macpijan:
4650:

```
Ethernet                                                                      
ETHPERF001.001 Check Performance of 2.5G Wired Network Interface (... .....
Checking if iperf3 is installed...

Package iperf3 is installed
ETHPERF001.001 Check Performance of 2.5G Wired Network Interface (... | PASS |
ETHPERF002.001 Check Performance of 10G Wired Network Interface (U... | SKIP |
ETHPERF002.001 not supported: None
Ethernet                                                              | PASS |
2 tests, 1 passed, 0 failed, 1 skipped
```

6670:

```
ETHPERF001.001 Check Performance of 2.5G Wired Network Interface (... .....
Checking if iperf3 is installed...

Package iperf3 is installed
ETHPERF001.001 Check Performance of 2.5G Wired Network Interface (... | PASS |
ETHPERF001.001 Check Performance of 10G Wired Network Interface (U... .....
Checking if iperf3 is installed...

Package iperf3 is installed
ETHPERF001.001 Check Performance of 10G Wired Network Interface (U... | PASS |
Network-Speed                                                         | PASS |
2 tests, 2 passed, 0 failed
```

dasharo-compatibility/network-speed.robot
macpijan: Reworked as per suggestion from @miczyg1

miczyg1: You have to make sure this test is not run on network interfaces providing connectivity with external world...

If it breaks the connection, then something is either wrong in the test or in the ionfrastracture

PR #340, opened 2024-07-01
Pid32 h automate novacustom tests v1

philipanda:


dasharo-compatibility/efibootmgr.robot
```diff
@@ -0,0 +1,86 @@
...
+# - go threough them and make sure they are doing what the name suggest (not
+# exactly the case right now)
+Suite Setup         Run Keyword
+...                     Prepare Test Suite
+Suite Teardown      Run Keyword
+...                     Log Out And Close Connection
+
+
+*** Test Cases ***
+EBM001.001 Network Boot enable, disable
```
mkopec, line 29: Typically we have separate tests for enable and disable

philipandag: I figured it may not make much sense to test enabling something if there is a chance it is already enabled because we may think that the enabling feature worked when in reality it just didn't change anything. That may cause tests to work incorrectly when done seperately or in wrong order.

Should I in this case test if NetworkBoot is disabled before enabling it and say the test failed if starting conditions were not right?

Or would it make more sense to trust that the disabling feature works and use it to make sure NetworkBoot is disabled? But then why not test if it actually worked in one go? 

Or should I just assume the starting conditions are right? I recon that's the case in other tests.

mkopec: For preparing platform state before tests, you can add keywords to configure the platform correctly to the  `Suite Setup` keyword ([here](https://github.com/Dasharo/open-source-firmware-validation/pull/340/files#diff-0cf6e1c840f878221aa7bed0f49ae2c8363bf2ce1b4ccb2f61b53de8383a2cafR22). That's what we did in https://github.com/Dasharo/open-source-firmware-validation/blob/f5bead6c91692e1759ed62d2de1ce913db442975/dasharo-security/me-neuter.robot#L27 for example.

> I figured it may not make much sense to test enabling something if there is a chance it is already enabled because we may think that the enabling feature worked when in reality it just didn't change anything. 

In this case either the enable or disable test will fail, and I think it makes sense to have these tested separately, so that we can see on first glance of the results which state is not working correctly. I think that was the idea when the other tests were written, anyway.

What you're proposing also makes sense, but in this case I think I'd follow the established convention. Maybe @macpijan can chime in :)

philipandag: > in this case I think I'd follow the established convention.

Sure, I am not trying to argue with your conventions.  I can't easily understand them yet as I am still very much unfamiliar with the codebase and the robot framework overall. Thanks for explaining that to me.

mkopec:
@philipandag should this still be a draft?

philipandag:
> @philipandag should this still be a draft?
@mkopec 
Not all tests are working, so I suppose not. I have explained it on Jira.

macpijan:
@BeataZdunczyk @mkopec are we going to use it for the upcoming release cycle? What is the desired state of this PR?

Issue #281, opened 2024-04-26
Temperature detection and reporting should be reworked


pkubaj:
### Device

Any

### RTE version

1.10

### Brief summary

Temperature detection and reporting should be reworked

### How to reproduce

Try running CPT tests.

### Expected behavior

CPT tests should not return failure when the temperature is almost all the time ok.

### Actual behavior

CPT tests detect and report the current temperature and return fail when it's too high even when there are only temporary spikes.

### Link to screenshots or logs

I don't have logs.

tym2k1:
Already addressed by https://github.com/Dasharo/open-source-firmware-validation/commit/9b50c95d93ac9b0f8501080044c5a4930cb8f2fa

PR #341, opened 2024-07-01
Temperature detection and reporting improvements

tym2k1:
Fixes for https://github.com/Dasharo/open-source-firmware-validation/issues/281

### DIFF
--- a/dasharo-performance/cpu-temperature.robot
+++ b/dasharo-performance/cpu-temperature.robot
@@ -123,19 +123,37 @@ CPU Temperature Without Load (Ubuntu)
     Execute Command In Terminal    sensors-detect --auto
     ${timer}=    Convert To Integer    0
     @{temperature_list}=    Create List
+    ${max_temperature}=    Set Variable    0
+    ${min_temperature}=    Set Variable    999
     ${sum}=    Convert To Integer    0
-    FOR    ${i}    IN RANGE    (${TEMPERATURE_TEST_DURATION} / ${TEMPERATURE_TEST_MEASURE_INTERVAL}) + 1
-        Log To Console    \n ----------------------------------------------------------------
-        Log To Console    ${timer}s
+    ${sum_previous}=    Convert To Integer    0
+    ${total_intervals}=    Evaluate    (${TEMPERATURE_TEST_DURATION} / ${TEMPERATURE_TEST_MEASURE_INTERVAL}) + 1
+    ${minute_counter}=    Set Variable    0
+    Log To Console    \nStarting Test...
+    FOR    ${i}    IN RANGE    ${total_intervals}
         ${temperature}=    Get CPU Temperature CURRENT
         Append To List    ${temperature_list}    ${temperature}
+        ${max_temperature}=    Evaluate    max(${max_temperature}, ${temperature})
+        ${min_temperature}=    Evaluate    min(${min_temperature}, ${temperature})
         ${sum}=    Evaluate    ${sum} + ${temperature}
-        Log To Console    Current temperature: ${temperature}°C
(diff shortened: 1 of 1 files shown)


tym2k1:
[Log to Console keyword shouln't be used here](https://github.com/Dasharo/open-source-firmware-validation/blob/main/README.md?plain=1#L299-L307)

dasharo-performance/cpu-temperature.robot
```diff
@@ -124,15 +124,24 @@ CPU Temperature Without Load (Ubuntu 22.04)
...
         Log To Console    Current temperature: ${temperature}°C
-        Should Be True    ${temperature} < ${MAX_CPU_TEMP}
+        ${temperature}=    Evaluate    float(${temperature})
+        Append To List    ${temperatures}    ${temperature}
+        ${max_temperature}=    Evaluate    max(${max_temperature}, ${temperature})
         Sleep    ${TEMPERATURE_TEST_MEASURE_INTERVAL}m
         ${timer}=    Evaluate    ${timer} + ${TEMPERATURE_TEST_MEASURE_INTERVAL}
     END
+    ${mean_temperature}=    Evaluate    sum(${temperatures}) / len(${temperatures})
+    Log To Console    Mean temperature over test duration: ${mean_temperature}°C
```
tym2k1, line 141: [`Log To Console` keyword shouln't be used to write to terminal](https://github.com/Dasharo/open-source-firmware-validation/blob/main/README.md?plain=1#L299-L307). Applies to all instances

tym2k1:
It would also be cool to graph the CPU temperature instead of printing the raw numbers. Tools like [termgraph](https://github.com/mkaz/termgraph) or [matplotlib](https://matplotlib.org/) can be used here. TBD

dasharo-performance/cpu-temperature.robot
tym2k1: My bad. Log to console is used correctly here. The mentioned keywords are used for writing to the platform's terminal remotely

tym2k1:
Ready for review. I've created an [issue](https://github.com/Dasharo/open-source-firmware-validation/issues/366) related to graphing the CPU temperature for it to be handled at a later time.

tym2k1:
While rebasing on top of `develop` i've noticed that the [functionality has been already added some time ago](https://github.com/Dasharo/open-source-firmware-validation/commit/9b50c95d93ac9b0f8501080044c5a4930cb8f2fa). Thus I only retained the logging improvements so console isn't flooded with log messages every second

dasharo-performance/cpu-temperature.robot
```diff
@@ -125,19 +125,37 @@ CPU Temperature Without Load (Ubuntu)
     Execute Command In Terminal    sensors-detect --auto
     ${timer}=    Convert To Integer    0
     @{temperature_list}=    Create List
+    ${max_temperature}=    Set Variable    0
+    ${min_temperature}=    Set Variable    999
```
macpijan, line 129: @tym2k1 is there any particular logic here to define max at 0 and min at 999?

tym2k1: The `min_temperature` later uses the python [min()](https://www.w3schools.com/python/ref_func_min.asp) function to set the value to the lesser one between the current temperature and the currently stored `min_temperature`. The value parsed to the `min()` function can't be undefined so by setting it to a high number we ensure that it will get overwritten. I suppose this logic might skew the results if the CPU temperature reading will surpass 999°C but such reading is unlikely to happen.

The reverse happens for `max_temperature`.

[Relevant Code](https://github.com/3mdeb/meta-rte/pull/70/commits/c69d3ed9d8cf24211c8424fc6592d6f7adf63401)

PR #344, opened 2024-07-02
32H automate device detection tests

philipandag:
Moved all related commits from #340 to a new pull request


lib/options/dcu.robot
```diff
@@ -48,3 +48,29 @@ Get UEFI Option
     ...    cd dcu && ./dcu v coreboot.rom --get "${option_name}"
     ...    shell=True
     RETURN    ${out.stdout}
+
+Execute Warmboot Command
```
mkopec, line 52: Please move these out of `lib/options/dcu.robot` . I think `lib/sleep-lib.robot` fits these better

dasharo-compatibility/device-detection.robot
```diff
@@ -0,0 +1,105 @@
...
+
+#     Set DUT Response Timeout    30s
+#     Login To Linux With Root Privileges
+#     ${out}=    Execute Command In Terminal
+#     ...    cbmem -1 | grep "UsbBusStart:"
+
+#     Should Contain    ${out}     usb bus started
+#     ...    ignore_case=True
+
+# DDET002.001 USB Stack enable
```
mkopec, line 51: Should these remain commented out?

lib/options/dcu.robot
```diff
@@ -48,3 +48,29 @@ Get UEFI Option
...
+Execute Suspend And Wake Command
+    [Documentation]    Suspends and then wakes up the device after some time
+
+    # Using "Execute Command In Terimal" will cause the test to wait
+    # for command prompt to appear before continuing but the prompt
+    # will not appear again until we Login after reboot, so the test
+    # would hang here and fail.
+    # Sometimes it may take long to shutdown all systemd services,
+    # so the waiting times have to be excessive to avoid false negatives.
+    Write Into Terminal    rtcwake -m disk -s 20
```
mkopec, line 74:
```suggestion
    Write Into Terminal    rtcwake -m disk -s 20
```

Suspend to disk is hibernation, the suspend tests typically test suspend to RAM (S3) and / or suspend to idle (S0ix). I'd use `mem` here

dasharo-compatibility/device-detection.robot
philipandag: Yes, they won't work because of the issues with a python library. Trying to write anything via ssh after disabling usb stack will cause an exception and end the whole test suite.

lib/options/dcu.robot
philipandag: I wrote those in `lib/options/dcu.robot` because I assumed the test suite may be also used to test devices via serial port. In that case any differences in performing those operations could be addressed by redefining the keywords in `lib/options/uefi-setup-menu.robot` .
Nevertheless its moved to `lib/sleep-lib.robot` now.

lib/options/dcu.robot
philipandag: Done

dasharo-compatibility/device-detection.robot
mkopec: Right, @macpijan any ideas what might be going wrong in the paramiko library?

lib/options/dcu.robot
mkopec: Thanks. Please remember to link to commits that address each review comment

philipandag: Oh, sure, it would be this one [50dd4f4](https://github.com/Dasharo/open-source-firmware-validation/pull/344/commits/50dd4f4a09dc87ef61ac4b4684e45300c2e0d23d). This commit addressed two comments. Next time i will address every comment in a seperate commit.

dasharo-compatibility/device-detection.robot
```diff
@@ -0,0 +1,104 @@
...
+
+   Execute Reboot Command
+   Sleep    10s
+
+   Set DUT Response Timeout    30s
+   Login To Linux With Root Privileges
+   ${out}=    Execute Command In Terminal
+   ...    cbmem -1 | grep "UsbBusStart:"
+
+   Should Contain    ${out}    usb bus started
```
mkopec, line 48: I think this should use the `Should Not Contain` keyword instead? Right now it's checking the same thing as `DDET002`

dasharo-compatibility/device-detection.robot
```diff
@@ -0,0 +1,104 @@
...
+
+DDET004.001 Usb Devices Detected In Firmware Suspension
+    [Documentation]    Test if USB devices are detected after a warmboot
+
+    Login To Linux With Root Privileges
+    Perform Suspend And Wake Using Rtcwake
+    Login To Linux With Root Privileges
+
+    ${out}=    Execute Command In Terminal
+    ...    cbmem -1 | grep -i 'UsbEnumeratePort'
```
mkopec, line 88: This log message is not generated on wake from suspend. This entire test case can be removed, because on the resume path from S3 we're skipping the entire UEFI payload which includes USB detection.

On wake from S0ix firmware is not actually involved at all

dasharo-compatibility/device-detection.robot
philipandag: That's right, thanks, fixed: https://github.com/Dasharo/open-source-firmware-validation/pull/344/commits/c96f25afaf0ff876c0b7b421a777974f30d117cb

dasharo-compatibility/device-detection.robot
philipandag: Right, so I will just remove it as you said. https://github.com/Dasharo/open-source-firmware-validation/pull/344/commits/c45c746329bb97d1a22dd262b03002b6723f4961

philipandag (commented):
Applied fix suggestions.

mkopec (approved):

PR #345, opened 2024-07-02
32H automate efi boot manager tests

philipandag:
Moved all relevant commits from #340 to a new pull request

### DIFF
--- a/lib/options/uefi-setup-menu.robot
+++ b/lib/options/uefi-setup-menu.robot
@@ -65,4 +65,6 @@ Get UEFI Boot Manager Entries
     [Documentation]    Read list of UEFI boot manager
 
     Power On
+    
     ${boot_menu}=    Enter Boot Menu Tianocore And Return Construction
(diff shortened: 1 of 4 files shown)


lib/options/uefi-setup-menu.robot
```diff
@@ -59,3 +59,11 @@ Get UEFI Option
 
     ${state}=    Get Option State    ${menu}    ${option_path[${path_len}-1]}    ${VALUE}
     RETURN    ${state}
+
+Get UEFI Boot Manager Entries
+    [Documentation]    Read list of UEFI boot manager
+
+    Power On
+
+    ${boot_menu}=    Enter Boot Menu Tianocore And Return Construction
```
mkopec, line 68: I'm a bit worried that the two `Get UEFI Boot Manager Entries` don't return the boot list in the same format, but it's probably fine for now.

mkopec (approved):

PR #348, opened 2024-07-03
Remove regression-qemu.sh

### LINKED ISSUE #320: Remove regression-qemu.sh script
All platforms should use a single regression.sh script

PLangowski:
closes https://github.com/Dasharo/open-source-firmware-validation/issues/320


scripts/regression.sh
```diff
@@ -3,12 +3,42 @@
...
 source "${SCRIPT_DIR}/lib/robot.sh"
 
-# FW_FILE and DEVICE_IP are required for full regression
-check_env_variable "FW_FILE"
-check_env_variable "DEVICE_IP"
+if [ $CONFIG == "qemu" ]; then
+  export RTE_IP=127.0.0.1
+  export SNIPEIT_NO="no"
+  compatibility_tests=(
+  "dasharo-compatibility/custom-boot-menu-key.robot"
```
macpijan, line 10: Ok, I am not a huge fan of such solution.
Ideally, we simply execute the ./rgression.sh as we do for any other platform, and only the required tests are executed, thanks to the correctly set test flags.

PLangowski: Isn't this what happens in this script? If we set `CONFIG` to `qemu`, only some of the tests are executed (the ones that were listed in `regression-qemu.sh`, so I assume those can run on QEMU)

macpijan: @PLangowski  But this should be resolved by properly setting support flags in platform-configs/qemu.robot, not be the if-else here.

macpijan:
@PLangowski You pushed some WIP commit in July, do you intend to finalize this at some point? Do you know how?

PLangowski:
@macpijan I got stuck here. For some reason RF fails to establish a serial connection over telnet when running `regression.sh` on qemu. The problem does not occur when I run the tests separately.
[dasharo-compatibility_log.txt](https://github.com/user-attachments/files/17989506/dasharo-compatibility_log.txt)

macpijan:
@PLangowski Well, if we look in the log, the obvious difference is the addition of

image

when running the whole regression.

This comes from the `dasharo-compatibility/__init__.robot`. You should find answer there.

PLangowski:
Yes, but I have modified it to only run `Prepare Test Suite`, which works when running single test files.

PR #312, opened 2024-06-10
remove  "boot from usb" kw

PLangowski:

Issue #279, opened 2024-04-25
No keyword with name 'Boot From USB' found.


pkubaj:
### Device

VP66?0

### RTE version

1.1.0

### Brief summary

UBT* and UDT* tests fail with No keyword with name 'Boot From USB' found.

### How to reproduce

`robot -b command_log.txt -v snipeit:no -L TRACE -v rte_ip:192.168.10.110 -v device_ip:192.168.10.114 -v fw_file:protectli_vp66xx_v0.9.0-rc2.rom -v config:protectli-vp6670 -t UBT001.001* dasharo-compatibility/usb-boot.robot`

### Expected behavior

Test should finish.

### Actual behavior

Logs indicate that in Try block, Boot From USB fails:
`No keyword with name 'Boot From USB' found.`

### Link to screenshots or logs

[log.html.gz](https://github.com/Dasharo/open-source-firmware-validation/files/15111282/log.html.gz)

macpijan:
This is still up to date and might need more rework of the USB tests.

macpijan:
It falls under USB problems: https://github.com/Dasharo/open-source-firmware-validation/issues/216

m-iwanicki:
Should be fixed by https://github.com/Dasharo/open-source-firmware-validation/pull/312
Some other errors that would also make UDT* tests fail were fixed in https://github.com/Dasharo/open-source-firmware-validation/pull/350

Checked on Protectli VP2410

logs:

```
robot -v snipeit:no -L TRACE -v rte_ip:192.168.10.233 -v config:protectli-vp2410 dasharo-compatibility/usb-detect.robot
==============================================================================
Usb-Detect
==============================================================================
UDT001.001 USB detection after coldboot :: Check whether the DUT d... | PASS |
------------------------------------------------------------------------------
UDT002.001 USB detection after warmboot :: Check whether the DUT d... | PASS |
------------------------------------------------------------------------------
UDT003.001 USB detection after system reboot :: Check whether the ... | PASS |
------------------------------------------------------------------------------
Usb-Detect                                                            | PASS |
3 tests, 3 passed, 0 failed
==============================================================================

robot -v snipeit:no -L TRACE -v rte_ip:192.168.10.233 -v config:protectli-vp2410 dasharo-compatibility/usb-boot.robot
==============================================================================
Usb-Boot
==============================================================================
UBT001.001 USB detect and boot after coldboot :: Check whether the... | PASS |
------------------------------------------------------------------------------
[...]

PR #350, opened 2024-07-03
Fix usb-detect tests

m-iwanicki:
PR removes non existing keywords and fixes `Enter Tianocore And Return Menu` keyword.
Also test to check if test should fail was moved inside try/except block to Fail test immediately when number of failed tries goes above some threshold (no need to wait for 5 fails if test would fail anyway after first one)
Related to https://github.com/Dasharo/open-source-firmware-validation/pull/312


macpijan:
@m-iwanicki Any logs and / or details on how it was tested?

m-iwanicki:
@macpijan Yes, in the issue: https://github.com/Dasharo/open-source-firmware-validation/issues/279#issuecomment-2206059601

Issue #256, opened 2024-03-29
Cannot assign label: INTERMED : (127) when generating VBOOT key


pkubaj:
### Device

APU2

### RTE version

1.1.0

### Brief summary

Cannot assign label: INTERMED : (127) when generating VBOOT key

### How to reproduce

`debian@debian:~/open-source-firmware-validation/scripts/secure-boot/generate-images$ ./sb-img-wrapper.sh `

### Expected behavior

Key should be generated.

### Actual behavior

```
Creating image for INTERMEDIATE 

Installing dependencies
Detected Package Manager: apt
Detected Distro: Debian
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
The following additional packages will be installed:
  libfaketime
The following NEW packages will be installed:
  faketime libfaketime sbsigntool
0 upgraded, 3 newly installed, 0 to remove and 0 not upgraded.
Need to get 133 kB of archives.
After this operation, 617 kB of additional disk space will be used.
Do you want to continue? [Y/n] 
Get:1 http://debian-archive.trafficmanager.net/debian bookworm/main amd64 libfaketime amd64 0.9.10-2.1 [47.3 kB]
Get:2 http://debian-archive.trafficmanager.net/debian bookworm/main amd64 faketime amd64 0.9.10-2.1 [17.6 kB]
Get:3 http://debian-archive.trafficmanager.net/debian bookworm/main amd64 sbsigntool amd64 0.9.4-3.1 [68.4 kB]
Fetched 133 kB in 0s (436 kB/s)       
Selecting previously unselected package libfaketime:amd64.
(Reading database ... 46719 files and directories currently installed.)
Preparing to unpack .../libfaketime_0.9.10-2.1_amd64.deb ...
Unpacking libfaketime:amd64 (0.9.10-2.1) ...
Selecting previously unselected package faketime.
Preparing to unpack .../faketime_0.9.10-2.1_amd64.deb ...
Unpacking faketime (0.9.10-2.1) ...
Selecting previously unselected package sbsigntool.
Preparing to unpack .../sbsigntool_0.9.4-3.1_amd64.deb ...
Unpacking sbsigntool (0.9.4-3.1) ...
Setting up sbsigntool (0.9.4-3.1) ...
Setting up libfaketime:amd64 (0.9.10-2.1) ...
Setting up faketime (0.9.10-2.1) ...
Processing triggers for man-db (2.11.2-2) ...
[...]

Issue #255, opened 2024-03-29
Unable to locate package sbsigntools


pkubaj:
### Device

APU2

### RTE version

1.1.0

### Brief summary

Unable to locate package sbsigntools when running sb-img-wrapper.sh

### How to reproduce

`./sb-img-wrapper.sh`

### Expected behavior

Tools should be installed and script should proceed.

### Actual behavior

```
Creating image for INTERMEDIATE 

Installing dependencies
Detected Package Manager: apt
Detected Distro: Debian
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
E: Unable to locate package sbsigntools
Cannot install sbsigntools faketime : (100)
```

### Link to screenshots or logs

There ane no logs.

pkubaj:
This issue happens when I force using apt in `generate-image.sh` as otherwise Debian is not detected.

pkubaj:
Package name is changed to `sbsigntool`.

BeataZdunczyk:
> This issue happens when I force using apt in generate-image.sh as otherwise Debian is not detected.

_Debian_? Do you mean Ubuntu @pkubaj?

pkubaj:
No, Debian, since that's what's installed on the VM.

macpijan:
Related to: https://github.com/Dasharo/open-source-firmware-validation/issues/256

Issue #254, opened 2024-03-29
Debian not detected in scripts/secure-boot/generate-images/sb-img-wrapper.sh


pkubaj:
### Device

APU2

### RTE version

1.1.0

### Brief summary

Debian not detected in scripts/secure-boot/generate-images/sb-img-wrapper.sh

### How to reproduce

```
debian@debian:~/open-source-firmware-validation/scripts/secure-boot/generate-images$ ./sb-img-wrapper.sh 
```

### Expected behavior

Debian should be detected.

### Actual behavior

```
Creating image for INTERMEDIATE 

Installing dependencies
Unknown Package Manager
debian@debian:~/open-source-firmware-validation/scripts/secure-boot/generate-images$ lsb_release -is
No LSB modules are available.
Debian
```

### Link to screenshots or logs

There are no logs.

macpijan:
Related to: https://github.com/Dasharo/open-source-firmware-validation/issues/256

artur-rs:
PR with fixes has been merged, closing

PR #212, opened 2024-03-04
Encrypted rootfs release

tomzy-0:
Gather all tests that were implemented while working on Encrypted rootfs based on TPM

PR #368, opened 2024-07-11
Fix issues 254 255 256

DaniilKl:
This PR solves bugs reported in in issues https://github.com/Dasharo/open-source-firmware-validation/issues/256, https://github.com/Dasharo/open-source-firmware-validation/issues/254 and https://github.com/Dasharo/open-source-firmware-validation/issues/255.

IMHO, if we want the script to be supported on several distros - it is better to use smth like Docker container, so to ommit problems with packages naming and versions.

### DIFF
--- a/scripts/secure-boot/generate-images/generate-image.sh
+++ b/scripts/secure-boot/generate-images/generate-image.sh
@@ -31,14 +31,11 @@ check_using_pkg_mng() {
         debian|ubuntu)
             PKG_MNG="apt"
             ;;
-        redhat|centos|fedora)
+        fedora)
             PKG_MNG="dnf"
             ;;
         arch)
             PKG_MNG="pacman"
-            ;;
-        openbsd)
-            PKG_MNG="pkg"
             ;;
         gentoo)
             PKG_MNG="emerge"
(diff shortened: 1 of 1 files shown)


macpijan:
@BeataZdunczyk Review / integration of this can be scheduled together with the other SB MRs.

BeataZdunczyk:
> Review / integration of this can be scheduled together with the other SB MRs.

@macpijan Could you provide more context? I don't understand why we should schedule a separate review/integration from another team for this PR. Shouldn't it be tested on QEMU (if possible) and hardware within the task fixing those issues scope?

macpijan:
I am not saying anything about teams.

This MR clearly targets the big and old one which we want to integrate into main branch at some point: https://github.com/Dasharo/open-source-firmware-validation/pull/212

So it seems sane that this can be reviewed by whoever will be finalizing this integration work.

scripts/secure-boot/generate-images/generate-image.sh
```diff
@@ -140,29 +143,35 @@ create_expired_cert() {
...
     for file in *; do
 
         if [ "$file" != "image.img" ]; then
-            cp $file "/run/media/$(whoami)/$IMAGELABEL/"
+            cp $file "$MOUNT_POINT/"
         fi
 
     done
-    umount "/run/media/$(whoami)/$IMAGELABEL"
+    umount "$MOUNT_POINT"
```
SebastianCzapla, line 174: We should probably remove the loop device after it is used, so that we don't leave it hanging after it is done, something like this works just fine:

```
umount "$MOUNT_POINT"
sudo losetup -d /dev/$_object_to_mount
```

Besides that LGTM and works as expected.

SebastianCzapla (changes requested):

scripts/secure-boot/generate-images/generate-image.sh
DaniilKl: Addressed here 48eaae863f110335d81b2818fa14472826990423.

SebastianCzapla:
Typo  in cannot + precommit error
  ```error_check "Connot delete /dev/$_object_to_mount."```

SebastianCzapla (approved):

DaniilKl:
This PR does not change KWs, no need to wait for these CIs, merging.

scripts/secure-boot/generate-images/generate-image.sh
m-iwanicki: Loop-device and mount points will remain if script fails somewhere earlier and exits. `trap` would be probably better.

PR #375, opened 2024-07-12
ME-neuter tests fix suggestion

wiktormowinski:
-Changed the method test cases : MES001.001, MNE002.001, MNE003.001, MNE004.001 
  are performed.
-Created a keyword to perform this operation


dasharo-security/me-keyword.resource
```diff
@@ -0,0 +1,9 @@
+*** Settings ***
+Library     BuiltIn
+Library     Process
+
+
+*** Keywords ***
+Check If ME Is Disabled
```
macpijan, line 7: This file should go into `lib/`, see how other keywords are distributed

macpijan: Please add documentation section, explaining what this keyword does, what it returns, what it means

dasharo-security/me-keyword.resource
```diff
@@ -0,0 +1,9 @@
...
+    ${out}=    Run Process    /usr/sbin/setpci    -s    16.0    40.L
+    RETURN    Should Not Be Equal As Strings    ${out.stdout}[3]    0
```
macpijan, line 9: Adding some comment here would be useful - e.g. we check for bit number X which should inidcate something

dasharo-security/me-keyword.resource
```diff
@@ -0,0 +1,9 @@
+*** Settings ***
+Library     BuiltIn
+Library     Process
+
+
+*** Keywords ***
+Check If ME Is Disabled
+    ${out}=    Run Process    /usr/sbin/setpci    -s    16.0    40.L
```
macpijan, line 8: Run Process executes program on your computer (where the `robot`) process is started, you should use `Execute Command In Terminal` instead, to run on the DUT

dasharo-security/me-keyword.resource
macpijan: Either add to some existing library, or create a new one, and include it in the keywords.robot

dasharo-security/me-keyword.resource
macpijan: Is this return value useful ? Do we rely on it somehow? I see the kwd is called, but we do not assign return value anywhere.

macpijan (changes requested):

dasharo-security/me-keyword.resource
wiktormowinski: now it has a use, everything is explained in comment which i also added

dasharo-security/me-keyword.resource
wiktormowinski: fixed

lib/me-keyword.resource
```diff
@@ -0,0 +1,22 @@
...
+
+*** Keywords ***
+Check ME State
+    [Documentation]    Keyword Checks The ME PCI Register Offset 0x40 To Find Current ME State, Returns:
+    ...    -Disabled(soft) For 3
+    ...    -Disabled(HAP) for 2
+    ...    -Enabled for 0
+
+    ${out}=    Execute Command In Terminal    setpci -s 16.0 40.L
+    ${char}=    Evaluate    ${out}[3]
```
mkopec, line 14: We're converting a 32bit integer to a string, and then getting the 3rd character from the left, a rather roundabout way to read bits 16:19 of a register :). Let's add a comment explaining what's going on here, something like:

```
# Current Operation Mode: bits 16:19 of register HFSTS1 (0x40)
```

lib/me-keyword.resource
```diff
@@ -0,0 +1,22 @@
...
+    IF    '${char}' == '3'
+        ${result}=    Evaluate    'Disabled(soft)'
```
mkopec, line 16:
```suggestion
        ${result}=    Evaluate    'Disabled (Soft)'
```

Let's make this consistent with how the option is called in the UEFI setup menu, and implemented in lib/options

lib/me-keyword.resource
```diff
@@ -0,0 +1,22 @@
...
+    ELSE IF    '${char}' == '2'
+        ${result}=    Evaluate    'Disabled(HAP)'
```
mkopec, line 18: a/a
```suggestion
        ${result}=    Evaluate    'Disabled (HAP)'
```

wiktormowinski: changes included in: https://github.com/Dasharo/open-source-firmware-validation/pull/375/commits/e8738cef4b6e89953cb45972e3df68a60965bd21

lib/me-keyword.resource
wiktormowinski: will the comment be sufficient then? if so here's the update: https://github.com/Dasharo/open-source-firmware-validation/pull/375/commits/e8738cef4b6e89953cb45972e3df68a60965bd21

lib/me-keyword.resource
wiktormowinski: done: https://github.com/Dasharo/open-source-firmware-validation/pull/375/commits/e8738cef4b6e89953cb45972e3df68a60965bd21

macpijan:
@wiktormowinski I rebased and squashed it.

On which devices it has been tested?

Protectli VP6670 returns:

```
setpci -s 16.0 40.L
setpci: Warning: No devices selected for operation group 1.
```

Because of that, your keyword returns:

image

Which means this solution is not perfect.

I guess, we should first check if the PCI device exists. And only if it exists, we can try to probe. If it does not exist, we should assume it is disabled?

wiktormowinski:
> @wiktormowinski I rebased and squashed it.
> 
> On which devices it has been tested?
> 
> Protectli VP6670 returns:
> 
> ```
> setpci -s 16.0 40.L
> setpci: Warning: No devices selected for operation group 1.
> ```
> 
> Because of that, your keyword returns:
> 
> image
> 
> Which means this solution is not perfect.
> 
> I guess, we should first check if the PCI device exists. And only if it exists, we can try to probe. If it does not exist, we should assume it is disabled?

@macpijan only VP2420 like the scope stated.
 I will look into it in the near future.

wiktormowinski:
https://github.com/Dasharo/open-source-firmware-validation/pull/403

PR #384, opened 2024-07-16
secure-boot-lib: Fix 'Enter UEFI Shell' documentation

m-iwanicki:
Maybe kwd should also be moved to another file, e.g. `menus.robot` as it doesn't really have anything to do with `Secure Boot`?


macpijan:
> Maybe kwd should also be moved to another file, e.g. menus.robot as it doesn't really have anything to do with Secure Boot?

makes sense

PR #387, opened 2024-07-18
cpuid_QEMU_TalosII_fix

wiktormowinski:


m-iwanicki:
@wiktormowinski you should probably target branch that has those caf414360c062ec19adc26fbf750bc5efae10c6e changes as without them keyword would fail completely for talos platform.
And if cpuid is really not needed then you could remove it from every config, but you should check if it isn't used in any test/keyword.

wiktormowinski:
you are right, i used develop instead of osfv-integraton-regression

PR #389, opened 2024-07-19
30B add tags to tests for minimum regression test scope

philipandag:
Add tags for minimum regression and make it possible to use tags and any other options of robot from the `scripts/run.sh` wrapper

### DIFF
--- a/scripts/run.sh
+++ b/scripts/run.sh
@@ -27,4 +27,4 @@ if [ "$#" -eq 0 ] || [ ! -e "$1" ]; then
   exit 1
 fi
 
-execute_robot "$1"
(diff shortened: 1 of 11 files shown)


scripts/run.sh
```diff
@@ -27,4 +27,4 @@ if [ "$#" -eq 0 ] || [ ! -e "$1" ]; then
   exit 1
 fi
 
-execute_robot "$1"
+execute_robot "$1" "${@:2}"
```
m-iwanicki, line 30:
```suggestion
execute_robot "$@"
```

Unless I'm missing something

philipandag: Yes, it should be equivalent

PR #394, opened 2024-07-19
Custom fan curve test fails

philipandag:
https://github.com/Dasharo/open-source-firmware-validation/issues/336
NovaCustom laptops should not support it - change configs
The test may be faulty - look into that


philipandag:
The issue is on commit 95f5deee1afdb1327b811c38489bdb63095492ce

macpijan:
@mkopec some old PR, is it possible still to review it? Seems not so invasive

PR #395, opened 2024-07-19
lib/dts-lib.robot: change flash method

wiktormowinski:


lib/dts-lib.robot
```diff
@@ -92,22 +92,22 @@ Run EC Transition
...
-    ...    command=flashrom --ifd -i bios -p internal -w /tmp/coreboot.rom --noverify-all
-    ...    timeout=320s
-    ${verified}=    Run Keyword And Return Status
-    ...    Should Contain    ${out}    VERIFIED
-    IF    ${verified} == ${FALSE}
-        ${out}=    Execute Command In Terminal
-        ...    command=flashrom --ifd -i bios -p internal -w /tmp/coreboot.rom --noverify-all
-        ...    timeout=320s
-        Should Contain    ${out}    identical
+    [Arguments]    ${fw_file_path}    ${region}=${EMPTY}
```
macpijan, line 95: @wiktormowinski Do I see correctly, or you simply copied over the body of the `Flash Via Internal Programmer` kwd from the `lib/flash.robot`?

This does not sound great.

You should probably remove `Flash Firmware in DTS` from here, and replace it's occurrences with the one from `lib/flash.robot`.

wiktormowinski: that is excatly what I did, and now it is fixed

PR #403, opened 2024-07-23
A branch devoted to fix the ,,me neuter tests"

wiktormowinski:


wiktormowinski:
@macpijan so, what do you think?

lib/me.robot
```diff
@@ -0,0 +1,30 @@
...
+Check ME State
+    [Documentation]    Keyword Checks The ME PCI Register Offset 0x40 To Find Current ME State, Returns:
+    ...    -Disabled (Soft) For 3
+    ...    -Disabled (HAP) for 2
+    ...    -Enabled for 0
+    TRY
+        ${out}=    Execute Command In Terminal    setpci -s 16.0 40.L
+        ${char}=    Evaluate    ${out}[3]
+        # Current Operation Mode: bits 16:19 of register HFSTS1 (0x40)
+    EXCEPT    Evaluating expression 'p' failed: NameError: name 'p' is not defined nor importable as module
```
macpijan, line 20: What exactly is this message?

lib/me.robot
```diff
@@ -0,0 +1,30 @@
...
+        ${char}=    Evaluate    2
```
macpijan, line 21: I understand that if this command fails (presumably due to no pci device), we assume it is HAP-disabled?

This will not always be the case, but I have no better idea to determine it easily. We can leave this logic like this, but it would be nice to leave some comments, e.g. that we cannot find ME PCI HECI device, so we assume it's HAP-disabled, but it might be soft disabled as well.

However, we rely here on the `setpci` to fail here. There are more cases why it may fail, not only when device does not exist.

So I would suggest, to modfiy this logic as follows:
- check via lspci if ME device exists<
- if not - return it is disabled
- if it does, use the setpci method to determine the actual state

macpijan:
@wiktormowinski I provided some suggestion how it can be improved

lib/me.robot
wiktormowinski: the updated keyword https://github.com/Dasharo/open-source-firmware-validation/pull/403/commits/3c4f4589ede0faa0dee81eacb5937e654ea2a784 and the output decoder https://github.com/Dasharo/open-source-firmware-validation/pull/403/commits/dd0ba0cdf1574aac122695d924e03bdf9453ba59

lib/me.robot
wiktormowinski: now i see making 3 different branches was a mistake .. here is the context: https://github.com/Dasharo/open-source-firmware-validation/pull/375#discussion_r1679311540

lib/me.robot
```diff
@@ -0,0 +1,42 @@
...
+
+*** Variables ***
+${EXPECTED}=    00:16.0
+
+
+*** Keywords ***
+Check ME Out
+    [Documentation]    Keyword Runs Lspci To Check Whether The ME Device
+    ...    Exists Or Not
+    ${out}=    List Devices In Linux    pci
```
JanPrusinowski, line 18: Just a thought/suggestion... This keyword is redundant. as it only glues together 'ls' with whatever we send to it. I think it is much more clear if we will use 'Execute Linux Command    lspci' instead. And in future maybe we should think about refactoring whole code to remove this keyword.

wiktormowinski: true https://github.com/Dasharo/open-source-firmware-validation/pull/403/commits/cfba1a7ddeaae34ada8fb9ddf05d9319543e5516

JanPrusinowski (approved):

PR #419, opened 2024-07-30
419 minimal regression regression fix

philipanda:


macpijan:
I can see these changes being already part of the develop branch.

PR #422, opened 2024-07-31
Add info how to enable serial in fw for pikvm on novacustom

JanPrusinowski:


docs/troubleshooting.md
```diff
@@ -0,0 +1,12 @@
...
+
+If DUT is configured to use PiKVM it is important to make sure that serial
+output is enabled in firmware. Default setting for serial output is off.
+This is why tests that flash firmware multiple times need to have
+fw file modified to have serial enabled. To prepare modified fw file use:
+
+```./dcu variable ./test/data/protectli_vault_cml_v1.2.0-rc1_vp46xx.rom --set
+SerialRedirection --value Enabled```
+
+* This is an example for Protectli VP46xx platform. Adjust the command to
```
macpijan, line 11: All of the Protectli boards have serial enabled by default, I suggest we use MSI here, as in practice that is the only board right now with this problem.

docs/troubleshooting.md
```diff
@@ -0,0 +1,12 @@
+# Tests fail on 'Suite Setup' - 'To Boot Directly' not found
+
+If DUT is configured to use PiKVM it is important to make sure that serial
```
macpijan, line 3: The actual problem is the default settings of Serial Redirection option in firmware. On some platforms (laptops, MSI, it is disabled by default). It may be worth mentioning, that this is not directly cause by the PiKVM itself, but it just happens that if we use PiKVM, we typically disable serial by default, so users can have shorter boot time by default, and only enable serial console if really needed.

docs/troubleshooting.md
```diff
@@ -0,0 +1,12 @@
+# Tests fail on 'Suite Setup' - 'To Boot Directly' not found
```
macpijan, line 1: `#` should happen only once in document, so use it as a name of the document, and use `##` for sections.

JanPrusinowski:
@macpijan I have added changes with regards to your comments

docs/troubleshooting.md
macpijan: We still use Protectli as an example @JanPrusinowski

docs/troubleshooting.md
```diff
@@ -0,0 +1,19 @@
...
+output is enabled in the firmware and this setting is turned on by default.
+This issue is not related to PiKVM itself. It will occur on any DUT that uses
+serial connection. However PiKVM may make it look like everything is working
+correctly (video output etc). On some platforms like example on the MSI
+laptops, the default setting for the Serial Redirection is off. This is why
+tests that flash firmware need to have the fw file modified to have the serial
+enabled. Otherwise every time firmware is flashed by the test the DUT setup
+will be broken. To prepare a modified fw file use:
+
+```./dcu variable ./test/data/protectli_vault_cml_v1.2.0-rc1_vp46xx.rom --set
```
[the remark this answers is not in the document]
JanPrusinowski: edited

JanPrusinowski:
fixed precommit issues at: 72d963c922e3e878a7e182ee485e3c8521f7cf46

macpijan (approved):
LGTM, thanks @JanPrusinowski

PR #424, opened 2024-08-01
platform-configs/optiplex-*: add

filipleple:


platform-configs/optiplex-7010.robot
```diff
@@ -0,0 +1,217 @@
...
+${DTS_SUPPORT}=                                 ${TRUE}
+${UPLOAD_ON_USB_SUPPORT}=                       ${TRUE}
+${CPU_TESTS_SUPPORT}=                           ${TRUE}
+${L2_CACHE_SUPPORT}=                            ${TRUE}
+${L3_CACHE_SUPPORT}=                            ${TRUE}
+${MEMORY_PROFILE_SUPPORT}=                      ${TRUE}
+${ESP_SCANNING_SUPPORT}=                        ${FALSE}
+${DCU_UUID_SUPPORT}=                            ${TRUE}
+${DCU_SERIAL_SUPPORT}=                          ${TRUE}
+
```
JanPrusinowski, line 90: Do those values differ from default? If no then there is no need to include them in those files.
Same goes for 9010.
It is very hard to extract specific data for this machine if all of the variables are included.

platform-configs/optiplex-7010.robot
```diff
@@ -0,0 +1,217 @@
...
+${DEVICE_AUDIO1}=                               DA Intel PCH
+
+
+*** Keywords ***
+Power On
+    Sonoff Power Cycle On
+    Sleep    2s
+    Rte Power On
+
+Flash Firmware Optiplex
```
JanPrusinowski, line 170: Is this Keyword in anyway different for 7010 and 9010? If not maybe it would be a good idea to create a common config file for Optiplex machines? Same goes for ```Read Firmware Optiplex```

variables.robot
```diff
@@ -462,6 +474,8 @@ ${OS_UBUNTU}=               ubuntu
 @{CONFIG45}=                &{RTE53}    &{USB11}    &{SSD15}
 @{CONFIG46}=                &{RTE54}    &{USB11}    &{SSD07}
 @{CONFIG47}=                &{RTE63}    &{EMMC01}
+@{CONFIG48}=                &{RTE64}    &{SSD16}
+@{CONFIG49}=                &{RTE65}
```
JanPrusinowski, line 478: No ssd in this one?

platform-configs/optiplex-7010.robot
macpijan: Does it even work? The RteCtrl is alerady removed from this repo. We already use osfv_cli libraries. This should be enough as long as the machine is correctly supported in osfv_cli: https://github.com/Dasharo/open-source-firmware-validation/blob/develop/lib/flash.robot#L86

platform-configs/optiplex-7010.robot
filipleple: Right, removed all the overlap in [this commit](https://github.com/Dasharo/open-source-firmware-validation/pull/424/commits/b65bed093f6a66656720c48b4da15e841c026b89)

variables.robot
filipleple: Updated in [this commit](https://github.com/Dasharo/open-source-firmware-validation/pull/424/commits/a5400a1df5c50c438d138565924e5ab2a6d9c75f)

platform-configs/include/optiplex-common.robot
JanPrusinowski, line 1: Do those values differ from the ones in default config? We want only the ones which differ so that for example we can turn on and off suites to run for all platforms. If those keys will be duplicated then this platform wouldnt be tested...

filipleple: @JanPrusinowski yeah I realize, I'm fairly convinced that I've only left in ones that do differ from the included default

JanPrusinowski (approved):

platform-configs/include/optiplex-common.robot
JanPrusinowski: @filipleple Approved! :)

PR #425, opened 2024-08-01
no mst dock display detection

wiktormowinski:


lib/docks.robot
```diff
@@ -111,3 +122,26 @@ Detect Docking Station In Linux
...
+
+Check Docking Connection
+    [Documentation]    Returns True if a monitor is connected to a docking station, False otherwise.
+    [Arguments]    ${content}    ${target}
+    ${lines}=    Split To Lines    ${content}
+    ${dock_connected}=    False
+    FOR    ${line}    IN    @{lines}
+        IF    ${line}    Contains    status: connected
+        ${connector}=    Get Connector Name    ${line}
+        IF    "${connector}" starts with "${target} and "${connector}" not in "eDP"
```
mkopec, line 134: this will also pass if a non eDP monitor is connected directly to the laptop, right? E.g. a HDMI laptop

mkopec: If there isn't a better solution for this, let's at least print a log message or add a comment explaining that this false positive may happen

mkopec (approved):

PR #448, opened 2024-08-12
menus.robot: added check if DUT reset succeded

JanPrusinowski:


lib/bios/menus.robot
```diff
@@ -535,13 +536,36 @@ Remove Disk Password
     END
     Press Key N Times    1    ${SETUP_MENU_KEY}
 
+Send Reboot Command
```
macpijan, line 539: Why do we define new keyword and not use it?

lib/bios/menus.robot
```diff
@@ -535,13 +536,36 @@ Remove Disk Password
...
+    Telnet.Write Bare    \x1bR\x1br\x1bR
+
 Tianocore Reset System
-    # EDK2 interprets Alt + Ctrl + Del on USB keyboards as reset combination.
-    # On serial console it is ESC R ESC r ESC R.
     IF    '${DUT_CONNECTION_METHOD}' == 'SSH'
         FAIL    SSH not supported for interfacing with TianoCore
     ELSE IF    '${DUT_CONNECTION_METHOD}' == 'Telnet'
-        Telnet.Write Bare    \x1bR\x1br\x1bR
+        FOR    ${i}    IN RANGE    0    5
```
macpijan, line 546: I do not like that we change it for Telnet only. It means now we will have two different methods of saving changes. One - more manual - for Telnet, the other one - for PiKVM. This sounds like asking for even more problems?=.

lib/bios/menus.robot
```diff
@@ -535,13 +536,36 @@ Remove Disk Password
...
+            ${out}=    Read From Terminal
+
+            # Clean up the string
```
macpijan, line 549: Do we need to implement it once again?

In the history of this function, we can see the "manual" method has been replaced with the CTRL+ALT+DELETE here: ee72aadba96d247500fb31c22d878c0113255e2c

If we can prove that:
- the "new" method was less reliable
- the "old" method was more reliable

We can simply bring back the old method.

On top of this, we can add some improvements, if necessary, but I would start with simply bringing the old way.

lib/bios/menus.robot
```diff
@@ -535,13 +536,36 @@ Remove Disk Password
...
+            ${out}=    Replace String    ${out}    \n    ${EMPTY}
+            ${out}=    Replace String    ${out}    \t    ${EMPTY}
+            ${out}=    Strip String    ${out}
+
+            Set Test Variable    ${CHECK}    ${FALSE}
```
macpijan, line 554: I would appreciate some comments to help us understand the thought process more. and have easier time following the concepts in the code

lib/bios/menus.robot
JanPrusinowski: The new method breaks from time to time: 
[Uploading log.zip…]()
image

Ill restore the "old" method and check if it will resolve the issue.

JanPrusinowski: [log2.zip](https://github.com/user-attachments/files/17631741/log2.zip)
Issue still persists... I'll restore previous version and see if it will make any difference

lib/bios/menus.robot
JanPrusinowski: I have tested 3 options for this keyword.
Current one - Which is the least "stable" There is an issue where sending CTRL+ALT+DEL in dasharo just does not reset the platform. I have tried sending this shortcut even 2 or 3 times - and the reboot was not always guaranteed. Actually about 1 in 5 times this would fail.

The one I wrote and submitted in this PR. I know the solution isnt perfect as it is - but it is only a proof of concept - and this seemed to work much better. It still had its hickups and hung from time to time. I wrote some tests which rebooted the DUT 50 times using this method and sometimes the test has passed. Sometimes it hung on 40+th try... But more testing should be done and this method should be refined.

3rd method was the 'old' implementation of the kwd (which stored how many times you should press enter and up arrow to get to the default cursor position in dasharo) <- i have run this kwd only once (50 times) and it looked pretty stable - i had to stop the test after some 40th try but as it actually works similarly to the method I tried to write IMO reverting the change to this kwd would greatly improve tests stability.

lib/bios/menus.robot
JakubRed: I think that this keyword has no application in the current revision of OSFV.

lib/bios/menus.robot
JakubRed: Checked for times on current `develop` and there were no issues. Thus I think that this keyword is not needed anymore.

lib/bios/menus.robot
```diff
@@ -551,6 +575,26 @@ Tianocore Reset System
         FAIL    Unknown connection method for config: ${CONFIG}
     END
 
+Go To Reset Option
```
JakubRed, line 578: If rework of the `Tianocore Reset System` keyword is not implemented, then this keyword is not needed.

self-tests/setup-and-boot-menus.robot
```diff
@@ -239,3 +239,15 @@ Get Menu Construction Stress Test
             Run Keyword And Continue On Failure    Should Not Contain    ${line}    disabled.
         END
     END
+
+Tianocore Reset System Stress Test
```
JakubRed, line 243: I see no use for this keyword in the current revision the OSFV.

PR #449, opened 2024-08-12
339 stop logging to terminal KW

philipandag:


keys-and-keywords/ubuntu-keywords.robot
```diff
@@ -40,3 +40,22 @@ Install Package
     Write Into Terminal    apt-get install --assume-yes ${package}
     Read From Terminal Until Prompt
     Set DUT Response Timeout    30s
+
+Stop Logging To Terminal
```
macpijan, line 44: Is it used anywhere? Since we add it in this PR?

keys-and-keywords/ubuntu-keywords.robot
```diff
@@ -40,3 +40,22 @@ Install Package
...
+    [Documentation]    This keyword returns TRUE if logging is disabled and
+    ...    FALSE if it is not.
+    ${ret}=    Execute Linux Command    cat /proc/sys/kernel/printk | cut -f1
+
+    RETURN    ${ret}
+
+Set Logging Level
+    [Documentation]    This keyword sets the logging level to given value [0; 7]
+    [Arguments]    ${level}
+    Execute Linux Command    echo "kernel.printk = ${level} 4 1 7" > /etc/sysctl/d/10-console-messages.conf
```
macpijan, line 59:
```suggestion
    Execute Linux Command    echo "kernel.printk = ${level} 4 1 7" > /etc/sysctl.d/10-console-messages.conf
```

keys-and-keywords/ubuntu-keywords.robot
```diff
@@ -40,3 +40,22 @@ Install Package
...
+
+Stop Logging To Terminal
+    [Documentation]    This keyword stops all unwanted logging to the terminal
+    ...    we are using for the shell. This prevents garbage from being logged.
+    Set Logging Level    0
+
+Get Logging Level
+    [Documentation]    This keyword returns TRUE if logging is disabled and
+    ...    FALSE if it is not.
+    ${ret}=    Execute Linux Command    cat /proc/sys/kernel/printk | cut -f1
```
macpijan, line 52: `cut -f 1 /proc/sys/kernel/printk` will also do

cat is redundant in many cases like this.

keys-and-keywords/ubuntu-keywords.robot
philipandag: It is not used anywhere, but the task was to add this keyword. It could jut as well be removed and `Set Logging Level` could be used instead. That is just a matter of the level of abstraction we want to have here.

keys-and-keywords/ubuntu-keywords.robot
```diff
@@ -40,3 +40,22 @@ Install Package
...
+
+    RETURN    ${ret}
+
+Set Logging Level
+    [Documentation]    This keyword sets the logging level to given value [0; 7]
+    [Arguments]    ${level}
+    Execute Linux Command    echo "kernel.printk = ${level} 4 1 7" > /etc/sysctl.d/10-console-messages.conf
+    Execute Linux Command    echo "kernel.printk = ${level} 4 1 7" >> /etc/sysctl.conf
+    Execute Linux Command    sysctl --system
```
JanPrusinowski, line 61: this could be changed to oneliner too: 
```echo "kernel.printk = ${level} 4 1 7" | tee /etc/sysctl.d/10-console-messages.conf >> /etc/sysctl.conf && sysctl --system```

however I leave it for your consideration.

Comment (philipandag) [reply keys-and-keywords/ubuntu-keywords.robot:61]: Thanks. I am not sure which version would be better though. If performance is considered then the oneliner should be preferred. If not then separating this into three calls may be a bit more readable in the code and in the logs?

File: keys-and-keywords/ubuntu-keywords.robot (hunk 3 above)
Comment (philipandag) [reply keys-and-keywords/ubuntu-keywords.robot]: https://github.com/Dasharo/open-source-firmware-validation/compare/4f351632f30b20656555653ad0d6b40a726bb3be..c2910d941d4d01840e0e8479d6c1dbaa8eb8f728#diff-b7afbf080f275d1995a5212ce98c26d58fb1004705c786cc31496e1043ac04b6L52

**Comment by philipandag** [comment]
pre-commit.ci is complying about something being deprecated. Locally pre-commit accepts these changes.

File: keys-and-keywords/ubuntu-keywords.robot (hunk 4 above)
Comment (philipandag) [reply keys-and-keywords/ubuntu-keywords.robot:62]: Turns out 4 actions need to be performed here, this could, again, be joined into one `Execute Linux Command` call, but in my opinion keeping these actions separate would be better.
 
https://github.com/Dasharo/open-source-firmware-validation/pull/449/commits/c2910d941d4d01840e0e8479d6c1dbaa8eb8f728#diff-b7afbf080f275d1995a5212ce98c26d58fb1004705c786cc31496e1043ac04b6R59-R62

**Comment by macpijan** [comment]
> pre-commit.ci is complaining about something being deprecated. Locally pre-commit accepts these changes.

No, the major problem is this: 

```
robotidy.................................................................Failed
- hook id: robotidy
- files were modified by this hook

```

The warnings about deprecation is another story, not related here.

philipandag:
> Maybe something broken slipped in to the develop branch.

That must have been the case. It is now fixed.

keys-and-keywords/ubuntu-keywords.robot
JanPrusinowski: Adding unused code is a bad practice. Nothing links this code to anything else in the project. If forgotten it may cause problems in future when someone will want to use a keyword with the same name.

"Best part is no part at all"

As we are not writing a library with some declared specification...
In my opinion it should be removed.
Unless it canbe used in the code somewhere - then it should be used in this PR.

philipandag: That is true. 
The issue did not include details on where this KW could be used, but because of this PR https://github.com/Dasharo/preseeds/pull/19 the keyword should not even be needed as long as the Ubuntu installation was performed using our auto-installer in the newest version.
There is also the question if `Get/Set Logging Level` keywords should be kept. I don't know at this moment where they could be needed, if at all.

philipandag: https://github.com/Dasharo/open-source-firmware-validation/compare/0b887780e8f42a1edd83a483217840694e9f4cce..cf93c30f34bed18e15389c9c5adc3043a8589bb8#diff-b7afbf080f275d1995a5212ce98c26d58fb1004705c786cc31496e1043ac04b6L44-L48

JanPrusinowski:
How about adding a check to: https://github.com/Dasharo/open-source-firmware-validation/blob/develop/util/basic-platform-setup.robot#L129 to verify if the log level is set correctly? This is a test run on "fresh" install of OS and can be run only once.

Is there a way to edit those settings after OS installation? If so maybe it would be a good idea to make the test perform required changes?

I think that editing installer is a good idea anyway. This way we make sure that this issue is not an issue... And test in this case may be redundant or it can just verify that everything is in order.

philipandag:
> How about adding a check to: https://github.com/Dasharo/open-source-firmware-validation/blob/develop/util/basic-platform-setup.robot#L129 to verify if the log level is set correctly? This is a test run on "fresh" install of OS and can be run only once.

It has already been added here. https://github.com/Dasharo/open-source-firmware-validation/pull/449/commits/5eb36045aa8f98a13d2cd9f709e51a78383c0b3f#diff-2fa360ccac1eb282bd2596d6c1b9f96038ea0a7e90c17ad80181bc5593766d8aR136-R137

> Is there a way to edit those settings after OS installation? If so maybe it would be a good idea to make the test perform required changes?

Yes, that's what the keywords I added do. https://github.com/Dasharo/open-source-firmware-validation/pull/449/commits/f503bc58c4d91ac2d6c52ac3b963e28e6deb8781

> I think that editing installer is a good idea anyway. This way we make sure that this issue is not an issue... And test in this case may be redundant or it can just verify that everything is in order.

The change to the installer has already been merged. If someone used it the logging level should be equal to 0. https://github.com/Dasharo/preseeds/commit/aa172c489c26aa2dbc31b86c5fed6064e3f65fbf

JanPrusinowski (approved):

PR #452, opened 2024-08-12
Skip Telnet-reliant USB tests on NVC laptops 

filipleple:
Gives false FAILs in minimal regression


dasharo-compatibility/usb-hid-and-msc-support.robot
```diff
@@ -64,6 +64,7 @@ USB002.001 USB keyboard detected in FW
     ...    correctly by the firmware and all basic keys work
     ...    according to their labels.
     [Tags]    minimal-regression
+    Skip If    not ${USB_KEYBOARD_DETECTION_SUPPORT}    USB002.001 not supported
```
mkopec, line 67: wouldn't `not ${TESTS_IN_FIRMWARE_SUPPORT}` fit better here?

mkopec (approved):

Issue #431, opened 2024-08-05
V5xTNx laptops need a powernap after executing `efibootmgr` in CNB001.001


philipandag:
### Device

V540TND, V560TNE

### RTE version

v0.9.1-rc1

### OSFV version

fix-novacustom-v5xtnx-configs-powerctrl

### Affected component(s) or functionality

CNB001.001 test

### Brief summary

The test fails and then the device is bricked for about 30 minutes

### How reproducible

seems to appear every time

### How to reproduce

Run CNB001.001

### Expected behavior

The test should pass and the device should not become bricked

### Actual behavior

The device enters a weir state where it seems it is bricked. The keyboard and control leds work fine but it never boots. Pressing power button causes the device to power off immediately suggesting it did not leave firmware

### Link to screenshots or logs

V54 log: https://cloud.3mdeb.com/index.php/s/ba2tz85pH3XyTEC
V56 log: https://cloud.3mdeb.com/index.php/s/TRYzeBPXc3LEXcd

V54 dasharo-security: https://cloud.3mdeb.com/index.php/s/GD9Gmni6XBLFg7Q
V56 dasharo-security: https://cloud.3mdeb.com/index.php/s/WPtcyE6772dLXKG

### Additional context

It is hard to reproduce because the laptops become unusable for about 30 minutes after the test. Both eventually managed to boot and work normally after performing multiple cmos resets and then leaving them on for about 30 minutes.

The described behavior also occured when performing dasharo-security.measured-boot, so it may not be connected to osfv. The V54 took the 'power nap' after MNE004.001 PASSed, the V56 survived for longer and become unresponsible after  MNE002.001 test.

It may be worth mentioning that during this bricked-like state the laptops get pretty hot.
### Solutions you've tried

Still haven't tried doing anything about it. Just finding out if it is reproducible takes a long time.

philipandag:
@macpijan that is an interesting behavior that you might want to know about

mkopec:
The cause has been narrowed down to commit https://github.com/Dasharo/open-source-firmware-validation/pull/455/commits/a539822bb98268ffd552f1dcc83709fd5d15d0f4 , not sure how to fix it correctly yet

philipandag:
@mkopec This commit a539822 does not appear on the branch on which I had the issue. Are you sure it is the cause of the problems?

philipandag:
[...]

PR #455, opened 2024-08-12
Mtl nvidia fixes

filipleple:


filipleple:
@mkopec so I think it's safe to assume this change wasn't causing the bricks, and it is necessary for full regressions to pass

philipandag:
I am not a reviewer but can confirm that this change fixes some issues with executing commands via ssh.

mkopec:
@philipandag unfortunately this commit is also possibly the cause of https://github.com/Dasharo/open-source-firmware-validation/issues/431 , when I tested on a rev before this commit it stopped bricking laptops

mkopec (approved):

PR #458, opened 2024-08-14
memory-profile.robot: type errors

philipandag:
"Get Option State" Keyword returns True/False if the value is a boolean value you "check" or "uncheck" and String if there is some text inside. Relying on the type of returned value may be dangerous and it caused the memory-profile test suite to fail because of type errors. 
But the memory profile is a list of values, not a checkbox! Now it turns out if an option value starts with an "X" it is treated as a boolean because the letter "X" is used as a checkmark. That is the case with options like "XMP#1...".

### DIFF
--- a/lib/bios/menus.robot
+++ b/lib/bios/menus.robot
@@ -284,12 +284,15 @@ Get Option State
     [Arguments]    ${menu}    ${option}
     ${index}=    Get Index Of Matching Option In Menu    ${menu}    ${option}
     ${value}=    Get Value From Brackets    ${menu}[${index}]
-    IF    '${value}[0]' == 'X'
-        ${state}=    Set Variable    ${TRUE}
-    ELSE IF    '${value}[0]' == ' '
-        ${state}=    Set Variable    ${FALSE}
-    ELSE
-        ${state}=    Set Variable    ${value}
+    ${len}=    Get Length    ${value}
+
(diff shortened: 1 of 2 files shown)


macpijan:
@philipandag Please fix pre-commit. Do you have it installed locally?

lib/bios/menus.robot
```diff
@@ -284,12 +284,15 @@ Get Option State
...
+    ${len}=    Get Length    ${value}
+
+    ${state}=    Set Variable    ${value}
+    IF    ${len} == 1
+        IF    '${value}[0]' == 'X'
+            ${state}=    Set Variable    ${TRUE}
+        ELSE IF    '${value}[0]' == ' '
+            ${state}=    Set Variable    ${FALSE}
+        END
     END
```
JanPrusinowski, line 296: Does this actually introduce any changes in how the code works?
It seems more complicated but does the same thing. Or is there some issue with comparing strings which didn't work in the original solution?

JanPrusinowski (approved):

macpijan:
@philipandag I love your detailed explanation in the comment.

Next time, can you please try to include it in the commit description instead? Please note that GH will automatically copy it over to MR comment as well. But this way, we store this knowledge in independent git repo, not only in the GH UX.

PR #461, opened 2024-08-16
180 performance tests fail on msi zx90

philipanda:
This PR does not strictly solve the issue, which seems to be already solved before, but the boot time measure tests did not work beause of other reasons which are fixed in this PR.

### DIFF
--- a/keywords.robot
+++ b/keywords.robot
@@ -1480,11 +1480,12 @@ Get Cbmem From Cloud
     [Documentation]    Download cbmem from the cloud.
     ${cbmem_path}=    Set Variable    /usr/local/bin/cbmem
     ${out_sha256sum}=    Execute Command In Terminal    sha256sum ${cbmem_path}
-    ${sha256}=    Set Variable    null
     IF    len(${out_sha256sum.split()}) != 0
         ${sha256}=    Set Variable    ${out_sha256sum.split()}[0]
+    ELSE
+        ${sha256}=    Set Variable    null
     END
-    IF        '${sha256}' != '169c5a5a63699cb37cf08d1eff83e59f146ffa98cf283145f27adecc081ac3f6'
+    IF    '${sha256}' != '169c5a5a63699cb37cf08d1eff83e59f146ffa98cf283145f27adecc081ac3f6'
         Download File    https://cloud.3mdeb.com/index.php/s/C6LJMi4bWz3wzR9/download    ${cbmem_path}
         Execute Command In Terminal    chmod 777 ${cbmem_path}
     END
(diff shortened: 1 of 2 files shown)


lib/terminal.robot
```diff
@@ -153,8 +153,10 @@ Execute Command In Terminal
     Set DUT Response Timeout    ${timeout}
     IF    '${DUT_CONNECTION_METHOD}' == 'Telnet'
         ${output}=    Telnet.Execute Command    ${command}    strip_prompt=True
-    ELSE IF    '${DUT_CONNECTION_METHOD}' == 'SSH'
-        ${output}=    SSHLibrary.Execute Command    ${command}
+        # With this executing the command via ssh to not return any
+        # output
+        # ELSE IF    '${DUT_CONNECTION_METHOD}' == 'SSH'
```
JanPrusinowski, line 158: Won't this break multiple tests which use SSH?

JanPrusinowski:
> This PR does not strictly solve the issue, which seems to be already solved before, but the boot time measure tests did not work beause of other reasons which are fixed in this PR.

```Get Cbmem From Cloud``` is not used in any of boot time measure tests. It is used in ```early-boot-dma-protection``` and ```tpm-support``` so how does it fix those tests?

File: keywords.robot [hunk 2]
```diff
@@ -1480,7 +1480,10 @@ Get Cbmem From Cloud
     [Documentation]    Download cbmem from the cloud.
     ${cbmem_path}=    Set Variable    /usr/local/bin/cbmem
     ${out_sha256sum}=    Execute Command In Terminal    sha256sum ${cbmem_path}
-    ${sha256}=    Set Variable    ${out_sha256sum.split()}[0]
+    ${sha256}=    Set Variable    null
```
JanPrusinowski, line 1483: I think it would be more clear if this line was moved to ```ELSE``` part of the ```IF``` below.

lib/terminal.robot
philipandag: This change is something like a cherry-pick-hot-fix of this change https://github.com/Dasharo/open-source-firmware-validation/pull/455/files which is now merged into develop. I will remove it so it merges well.

keywords.robot
philipandag: https://github.com/Dasharo/open-source-firmware-validation/compare/47fba484de7fdf64c98e2f4f7f383a2ea86b400b..623c8aad979a5274ae777c8a1f27038655e3d869

philipandag:
> Get Cbmem From Cloud is not used in any of boot time measure tests. It is used in early-boot-dma-protection and tpm-support so how does it fix those tests?

It is used by dasharo-performance/boot-time-measure.robot -> Measure Cold/Warm/Reboot Time -> Get Boot Time From Cbmem. 

The core of the issue was the ssh connection failing without the cherry-pick from Filip L's fix. This change is not necessary with the fix. It is helpful only in case the `Execute Command In Terminal` would return an empty string which should not happen with the fix. If it would happen though then with this check the issue would be easier to understand. If this is desirable then all the `Get X From Cloud` keywords should be updated accordingly. If not then this change is unnecesary and can be removed.

PR #465, opened 2024-08-20
lib/robot.sh: Fix splitting args containing whitespace

philipandag:
This solution uses a couple for loops which may not be optimal, that's the first thing that i came up with.


macpijan:
@philipandag pre-commit fails. You can try rebasing if it passes locally on your side.

philipandag:
@macpijan 
A change not passing pre-commit was merged into develop. Here is a quick fix so that everyone can rebase and have it fixed: https://github.com/Dasharo/open-source-firmware-validation/pull/486

macpijan (approved):

scripts/lib/robot.sh
```diff
@@ -33,15 +33,29 @@ execute_robot() {
...
-    _robot_args=""
-  fi
-  local _test_name
-  _test_name="$(echo "$_test_path" | sed -e 's/\ /_/' | awk '{$1=$1};$1')"
+  local _args=("$@")
+  local _test_path=()
+  local _seperator_idx=-1
+  local _args_len=${#_args[@]}
+  _test_path=()
+
```
macpijan, line 41: @philipandag Would it make sense to add some comments with these blocks, to explain what your goal is?

philipandag: Yes, it most definitely would make sense. I will explain it a bit.

philipandag: https://github.com/Dasharo/open-source-firmware-validation/pull/465/commits/6b75d6c1af325478c85c72c5386c347e346aef8a

macpijan (approved):

PR #477, opened 2024-08-21
v560tne-v0.9.1-rc3-testing-fixes

philipandag:
fix usb-hid-and-msc-support tests missing skip
if tests in firmware not supported

### DIFF
--- a/lib/options/dcu.robot
+++ b/lib/options/dcu.robot
@@ -23,7 +23,8 @@ Set UEFI Option
     Login To Linux
     Switch To Root User
     Get Flashrom From Cloud
-    Execute Command In Terminal    flashrom -p internal -r coreboot.rom --fmap -i FMAP -i SMMSTORE &> /dev/null
+    ${result}=    Execute Command In Terminal    flashrom -p internal -r coreboot.rom --fmap -i FMAP -i SMMSTORE
+    Should Not Contain    ${result}    read-only
     Execute Command In Terminal    chmod 666 coreboot.rom
     SSHLibrary.Get File    coreboot.rom    dcu/coreboot.rom
     ${result}=    Run Process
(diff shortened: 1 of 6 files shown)


lib/terminal.robot
```diff
@@ -153,8 +153,6 @@ Execute Command In Terminal
     Set DUT Response Timeout    ${timeout}
     IF    '${DUT_CONNECTION_METHOD}' == 'Telnet'
         ${output}=    Telnet.Execute Command    ${command}    strip_prompt=True
-    ELSE IF    '${DUT_CONNECTION_METHOD}' == 'SSH'
-        ${output}=    SSHLibrary.Execute Command    ${command}
```
JanPrusinowski, line 157: Wont removing this part breake tests for other machines?

lib/options/dcu.robot
```diff
@@ -33,7 +34,7 @@ Set UEFI Option
     Should Contain    ${result.stdout}    Success
     SSHLibrary.Put File    dcu/coreboot.rom    coreboot.rom
     ${result}=    Execute Command In Terminal
-    ...    flashrom -p internal -w coreboot.rom --fmap -i SMMSTORE --noverify-all &> /dev/null
+    ...    flashrom -p internal -w coreboot.rom --fmap -i SMMSTORE    300s
```
JanPrusinowski, line 37: wont doping the `noverify-all` flag cause more tear in the chip? Or is there a reason to always flash whole die?

lib/terminal.robot
philipandag: It might, I suppose.  It is required for testing the laptops though. I am not sure why it helps yet, I was instructed to use this change and it did help.
I thought this commit was already merged into develop. Either I imagined it or it was removed.

lib/options/dcu.robot
philipandag: Verifying will certainly wear the chip more, but the test will not continue if the flashing failed in an obvious way. If the wear is a real concern, then this change can be omitted at the cost of clearer logs in case of failed flashing.

mkopec: Verifying is reading so the added wear, if any, is miniscule. Even if it was writing the entire chip I don't think we need to concern ourselves with flash wear-out

lib/terminal.robot
mkopec: Let's check if it does indeed break other machines, or maybe if this change is already in develop

philipandag: This change is already on the develop branch https://github.com/Dasharo/open-source-firmware-validation/blob/develop/lib/terminal.robot#L148-L163

philipandag: After rebasing the file does not change in this PR.

philipandag:
@mkopec rebased

mkopec:
CI checks are failing though :(

mkopec (approved):
Feel free to merge when you get CI sorted out

PR #493, opened 2024-09-06
250 vbo009001 fails on apu

JanPrusinowski:


platform-configs/novacustom-v560tnd.robot
```diff
@@ -16,11 +16,3 @@ ${CPU_MAX_FREQUENCY}=               4800
 ${CPU_MIN_FREQUENCY}=               200
 
 ${NVIDIA_GRAPHICS_CARD_SUPPORT}=    ${TRUE}
-
-${POWER_CTRL}=                      none
```
macpijan, line 20: @JanPrusinowski I wonder why do we need such a change in this MR?

JanPrusinowski: No we dont. I have cleaned the Commits - rebased to current develop. I noticed that some of my fixes from this PR were merged already with other issues.

platform-configs/pcengines-apu4.robot
```diff
@@ -3,4 +3,5 @@ Resource    include/pcengines.robot
 
 
 *** Variables ***
-${DMIDECODE_PRODUCT_NAME}=      apu4
+${DMIDECODE_PRODUCT_NAME}=          apu4
+${INTERNAL_PROGRAMMER_CHIPNAME}=    W25Q64BV/W25Q64CV/W25Q64FV
```
krystian-hebel, line 7: Is this required only for apu4 or for all apus? IIRC flash is the same on all platforms.

In any case, this should be in separate commit, it has nothing to do with the keyword added in the other file.

macpijan:
It's actually easier to create new MRs rather than fixing this one.

1. https://github.com/Dasharo/open-source-firmware-validation/pull/585/files

PR #501, opened 2024-09-13
capsule-update.robot: Persistence of User Data in Capsule Updates

JanPrusinowski:


dasharo-stability/capsule-update.robot
```diff
@@ -70,6 +70,92 @@ CUP002.001 Capsule Update With Wrong GUID
...
+    ${unplugged}=    Run Keyword And Return Status    Should Not Contain    ${out}    No such file
+
+CUPXX6.001 Verifying BIOS Settings Persistence After Update
+    [Documentation]    Check if BIOS settings didn't change after Capsule Update.
+    Power On
+    ${setup_menu}=    Enter Setup Menu Tianocore And Return Construction
+    ${dasharo_menu}=    Enter Dasharo System Features    ${setup_menu}
+    ${net_menu}=    Enter Dasharo Submenu    ${dasharo_menu}    Networking Options
+
+    Set Option State    ${net_menu}    Enable network boot    ${TRUE}
```
krystian-hebel, line 141: This option is modified by many other tests, I don't think each of them restores it properly. Additionally, this is a boolean value, which means that with fully random value you have 50% chance of success. Perhaps setting [Auto Boot Time-out](https://docs.dasharo.com/dasharo-menu-docs/boot-maintenance-mgr/) to unreasonably high value and testing that would be better. It is also always present.

dasharo-stability/capsule-update.robot
```diff
@@ -278,3 +364,31 @@ Display Preparation Instructions
     Log To Console    Without it, a successful flash of DUT will prevent tests from working
     Log To Console    correctly.
     Log To Console    \n******************************************************************************
+
+Prepare DUT For Logo Persistence Test
+    IF    '${MANUFACTURER}' != 'QEMU'
```
krystian-hebel, line 369: Is there a reason why it wouldn't work on QEMU, or just not tested?

mkopec: Yeah this is not a good test for whether persistent bootsplash is supported.

We could add another flag, or have the test check if it's supported:

```
$ cbfstool build/coreboot.rom layout -w | grep BOOTSPLASH > /dev/null && echo "Custom logo supported" || echo "Custom logo not supported"
Custom logo supported
```

dasharo-stability/capsule-update.robot
```diff
@@ -70,6 +70,92 @@ CUP002.001 Capsule Update With Wrong GUID
...
+    Switch To Root User
+    ${updated_uuid}=    Get Firmware UUID
+    Log To Console    \n[after update] ${updated_uuid}
+
+    Should Be Equal    ${original_uuid}    ${updated_uuid}
+
+    Restore DUT Initial State
+
+CUPXX4.001 Verifying Serial
+    [Documentation]    Check if UUID didn't change after Capsule Update.
```
mkopec, line 96:
```suggestion
    [Documentation]    Check if serial number didn't change after Capsule Update.
```

dasharo-stability/capsule-update.robot
mkopec: If the goal is to test SMMSTORE preservation, I'd just add a generic EFI variable from Ubuntu that doesn't impact firmware functionality and isn't subject to resetting to a default. Use `efivar` from Ubuntu to create a new variable, update, and check if it's still there.

dasharo-stability/capsule-update.robot
JanPrusinowski: There is no way to flash the dut in qemu to prepare it fot the test. It can be done differently by preparing original qemu.rom file with bootsplash logo and then preparing capsule update.
This is only because you have to manualy change fw in qemu to original one after each test.

dasharo-stability/capsule-update.robot
```diff
@@ -70,6 +70,92 @@ CUP002.001 Capsule Update With Wrong GUID
     ${out}=    Execute UEFI Shell Command    CapsuleApp.efi -S
     Should Contain    ${out}    Capsule Status: Not Ready
 
+CUPXX3.001 Verifying UUID
```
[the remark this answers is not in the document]
JanPrusinowski: So how do I set custom values of UUID etc? By DCU? I agree that there is a lot of flashing going on. I'd prefer flashing once only and check if data in BIOS has changed.

dasharo-stability/capsule-update.robot
JanPrusinowski: We can test both of those cases. @mkopec could you suggest how to set or maybe what would be a good pick to use in such case?

dasharo-stability/capsule-update.robot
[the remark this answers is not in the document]
SergiiDmytruk: The tool that generates ROMHOLE data is in `util/msi/romholetool` of coreboot's repo.  You could update the code and run:
```
romholetool -i build.h -a 0xff7c0000 -s 0x20000 -o romhole
```
to generate `romhole` file and use `cbfstool` to put it into a firmware image.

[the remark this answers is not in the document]
JanPrusinowski: @krystian-hebel I have reworked those tests. Now the original keys are read in suite prepare and then checked if they changed in tests performed after capsule update. This should be much aster and reduce the wear on the chip.
6a367c9e760c23abc55ceaf20785cb0932d6773d

dasharo-stability/capsule-update.robot
JanPrusinowski: @krystian-hebel I have reworked this test to utilize Auto Boot Time-out: cdd63ce30ebcfe4732c50c4b8cfe834ed2923591

It is a 2 part test - again to update FW only once.

dasharo-stability/capsule-update.robot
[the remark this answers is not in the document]
krystian-hebel: > The tool that generates ROMHOLE data is in `util/msi/romholetool` of coreboot's repo. You could update the code and run:
> 
> ```
> romholetool -i build.h -a 0xff7c0000 -s 0x20000 -o romhole
> ```
> 
> to generate `romhole` file and use `cbfstool` to put it into a firmware image.

We could also generate it once and keep in `scripts/capsules`, as we did with keys.

dasharo-stability/capsule-update.robot
krystian-hebel: This will be good for now. We may have to add test for cases like https://github.com/Dasharo/DasharoModulePkg/commit/1a488a4da5643010d32004ee41bba525bf84b44a in the future, but those will be added when such incompatibility appears.

dasharo-stability/capsule-update.robot
```diff
@@ -25,10 +25,14 @@ Suite Setup         Run Keywords
 ...                     AND
 ...                     Check If CAPSULE FW FILE Is Present
 ...                     AND
-...                     Flash Firmware If Not QEMU
+...                     Prepare For Logo Persistence Test
 ...                     AND
 ...                     Upload Required Files
```
krystian-hebel, line 30: This should happen after firmware is flashed, in case previous firmware didn't boot, unless of course one of the earlier setup keywords does it.

JanPrusinowski: Thing is that we want to flash the FW with prepared logo to check if the capsule update changes it. So it needs to be added before flashing.

dasharo-stability/capsule-update.robot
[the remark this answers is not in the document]
JanPrusinowski: > The tool that generates ROMHOLE data is in `util/msi/romholetool` of coreboot's repo. You could update the code and run:
> 
> ```
> romholetool -i build.h -a 0xff7c0000 -s 0x20000 -o romhole
> ```
> 
> to generate `romhole` file and use `cbfstool` to put it into a firmware image.

I cant compile this tool: 
image
How can I create ROMHOLE? Do i need to compile some other pack of tools?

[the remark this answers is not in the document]
SergiiDmytruk: Its build instructions are integrated into coreboot's Makefiles, so need to configure for an MSI board and either build coreboot or use:
```bash
make build/util/msi/romholetool
```

dasharo-stability/capsule-update.robot
BeataZdunczyk: @krystian-hebel, please take a look ^

krystian-hebel: Yes, `Prepare For Logo Persistence Test` must happen before flashing, but `Upload Required Files` must take place after flashing.

dasharo-stability/capsule-update.robot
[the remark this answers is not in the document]
JanPrusinowski: [log rom and cap.zip](https://github.com/user-attachments/files/17129982/log.rom.and.cap.zip)
I have reworked logo test. Everything looks ok. Im starting on ROMHOLE test.

dasharo-stability/capsule-update.robot
JanPrusinowski: @krystian-hebel Changed the order in: a6d1871fd413293000eb2c8994413e33e55b3654
Actually it simplifies suite prepare as there is no need to boot to ubuntu twice to upload files and get system data.

dasharo-stability/capsule-update.robot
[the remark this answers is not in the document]
JanPrusinowski: @SergiiDmytruk I'm still having problems with using this tool...
image
I would apreciate a step by step instruction of how to add this data to a rom file.

> romholetool -i build.h -a 0xff7c0000 -s 0x20000 -o romhole

Does not work in any of the folders that I try to run it.
And I still dont know how to check this data from OS or Shell.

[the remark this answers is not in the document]
SergiiDmytruk: You should have called `build/util/msi/romholetool` or just extracted `ROMHOLE` from an image after building it.  Here, I've made ROMHOLE region that changes UUID to `00112233-4455-6677-8899-aabbccddeeff`: [romhole.zip](https://github.com/user-attachments/files/17132716/romhole.zip)

You can insert it into an image with:

```bash
build/cbfstool build/coreboot.rom write -r ROMHOLE -f romhole
```

Serial number requires constructing a valid `struct smbios_type2_v3_4` which `romholetool` doesn't do, so I wouldn't bother with checking its value, UUID should be enough to demonstrate that ROMHOLE is preserved.

> And I still dont know how to check this data from OS or Shell.

Use `smbiosview -t 1` in UEFI shell (shows FFs or UUID as a hexdump) or `dmidecode -t 1` in GNU/Linux (shows either "not present" or `00112233-4455-6677-8899-aabbccddeeff`.

dasharo-stability/capsule-update.robot
```diff
@@ -89,11 +103,54 @@ CUP999.001 Capsule Update
     Should Contain    ${out}    CapsuleMax
     Should Not Contain    ${out}    CapsuleLast
 
+CUP160.001 Verifying UUID
+    [Documentation]    Check if UUID didn't change after Capsule Update.
+    Go To Ubuntu Prompt
```
krystian-hebel, line 108: This is common to all of post-update tests except for BIOS settings persistence. This keyword takes 2:45 each time, which means additional 5:30 to the test suite execution time.

My proposition:
- make `Get System Values` take arguments which would specify the name of suite variables to write to (remember to **not** use escaped versions, as not recommended in https://robotframework.org/robotframework/latest/libraries/BuiltIn.html#Using%20variables%20with%20keywords%20creating%20or%20accessing%20variables)
- use `Get System Values    '$ORIGINAL_UUID'    '$ORIGINAL_SERIAL'    '$CUSTOM_LOGO_SHA'` in suite setup (I'm not sure about how exactly those names should be escaped)
- in these 3 tests, change first two lines (more lines in case of logo) to something like:
    ```
    Run Keyword If    $UPDATED_UUID is None    Get System Values    '$UPDATED_UUID'    '$UPDATED_SERIAL'    '$UPDATED_LOGO_SHA'
    ```
    Of course, change `$UPDATED_UUID is None` to whatever current test checks.

This should make it boot only once, while still allowing to skip some tests when not needed.

dasharo-stability/capsule-update.robot
```diff
@@ -278,3 +339,59 @@ Display Preparation Instructions
...
+    Log To Console    [UUID Before Update] ${ORIGINAL_UUID}\n
+
+    ${out}=    Execute Command In Terminal
+    ...    sha256sum /sys/firmware/acpi/bgrt/image
+    ${unplugged}=    Run Keyword And Return Status
+    ...    Should Contain    ${out}    No such file
+    IF    ${unplugged} == ${TRUE}
+        Fail    Please make sure that a display device is connected to the DUT
+    END
+    Set Suite Variable    ${CUSTOM_LOGO_SHA}    ${out}
```
krystian-hebel, line 397:
```suggestion
    IF    $CUSTOM_LOGO_SUPPORT == $TRUE
        ${out}=    Execute Command In Terminal
        ...    sha256sum /sys/firmware/acpi/bgrt/image
        ${unplugged}=    Run Keyword And Return Status
        ...    Should Contain    ${out}    No such file
        IF    ${unplugged} == ${TRUE}
            Fail    Please make sure that a display device is connected to the DUT
        END
        Set Suite Variable    ${CUSTOM_LOGO_SHA}    ${out}
    END
```

krystian-hebel (changes requested):

dasharo-stability/capsule-update.robot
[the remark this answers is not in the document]
JanPrusinowski: Ok. Have created an additional step in preparation which prepares FW before flashing in regards to Sergiis suggestions. Also added generated romhole and cbfstool to files required by test.
Logs and FW which I used: [logs and fw.zip](https://github.com/user-attachments/files/17145579/logs.and.fw.zip)

dasharo-stability/capsule-update-files/cbfstool
SergiiDmytruk, line 1: The binary might stop working at some point and might not work on all distributions now either.  I think ROMHOLE should be at the same location in every image so `dd` can be used to write it (with `seek=... conv=notrunc`).

dasharo-stability/capsule-update.robot
```diff
@@ -89,11 +103,54 @@ CUP999.001 Capsule Update
...
+    Should Be Equal    ${updated_state}    32123
+
+CUP190.001 Verifying If Custom Logo Persists Across updates
+    [Documentation]    Check if Logo didn't change after Capsule Update.
+    Skip If    not ${CUSTOM_LOGO_SUPPORT}    not supported
+
+    Power On
+    Boot System Or From Connected Disk    ubuntu
+    Login To Linux
+    Switch To Root User
```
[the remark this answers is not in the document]
JanPrusinowski: Edited in: f6b555c689a14dddf3de2e169562bc3af2ccc2e8

dasharo-stability/capsule-update.robot
```diff
@@ -89,11 +103,54 @@ CUP999.001 Capsule Update
...
+
+CUP170.001 Verifying Serial Number
+    [Documentation]    Check if serial number didn't change after Capsule Update.
+    Go To Ubuntu Prompt
+    ${updated_serial}=    Get Firmware Serial Number
+    Log To Console    \n[Serial Number After Update] ${updated_serial}
+
+    Should Be Equal    ${ORIGINAL_SERIAL}    ${updated_serial}
+
+CUP180.001 Verifying BIOS Settings Persistence After Update - PART 2
```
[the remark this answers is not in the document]
JanPrusinowski: changed order in: f6b555c689a14dddf3de2e169562bc3af2ccc2e8

dasharo-stability/capsule-update-files/cbfstool
JanPrusinowski: What is the solution to this? Download whole coreboot - build it and use the tool to create binary each time?

mkopec: you can build cbfstool pretty quickly

```
git clone --depth=1 https://review.coreboot.org/coreboot.git
cd coreboot
make -C util/cbfstool
```

dasharo-stability/capsule-update.robot
krystian-hebel: > lambda function (like in C)

C doesn't have lambda functions.

What I'm asking is to add arguments that are **names** of variables, that would be expanded as per linked documentation. If it is mentioned in the official documentation, chances are that RF has such capability :slightly_smiling_face: 

> Or I could merge those 3 tests to 1.

No, that way we wouldn't know which one failed without looking in the logs. Or worse, logic for skipping a part of the test would skip something that should be there.

dasharo-stability/capsule-update-files/cbfstool
JanPrusinowski: cbfstool is not the issue here... rather generation of a romhole file for each rom.
Ill add changes to clone coreboot etc - but romhole file will need to be generated manually.

SergiiDmytruk: No need for `cbfstool` at all, this does the same:
```
dd if=romhole of=build/coreboot.rom seek=$((0x17c0000)) bs=1 conv=notrunc
```

No need to create a romhole file per platform, `romholetool` produces identical one for all of them.

krystian-hebel: @JanPrusinowski remember to adjust paths in `of` and `if` to structure of directories in this repo.

dasharo-stability/capsule-update.robot
```diff
@@ -278,3 +342,71 @@ Display Preparation Instructions
...
+
+Prepare For Logo Persistence Test
+    Log To Console    PREPARE: Logo Persistence Test
+
+    Run    rm -rf dcu
+    Run    git clone https://github.com/Dasharo/dcu
+    Run    cp ${FW_FILE} dcu/coreboot.rom
+
+    ${out}=    Run
+    ...    cbfstool ./dcu/coreboot.rom layout -w | grep BOOTSPLASH > /dev/null && echo "Custom logo supported" || echo "Custom logo not supported"
```
krystian-hebel, line 354: This also depends on `cbfstool`, but I think we can skip it. The same test is done by `dcu`: https://github.com/Dasharo/dcu/blob/main/src/logo_command.sh#L20, it should be enough to test for `Customizable logo not supported` in the output.

krystian-hebel (changes requested):

dasharo-stability/capsule-update.robot
krystian-hebel: @JanPrusinowski I've tested the variable handling and discovered that:

- in the keyword invocation, variable names must be passed without any quotation and curly braces (e.g. `$ORIGINAL_UUID`), no other form worked for me
- in the `[Arguments]`, I used `${var_uuid}` form, haven't checked other forms but this one works
- in the body of keyword, `${var_uuid}` form must be used when calling `Set Suite Variable` to have the name properly expanded.

I've updated my comment above accordingly, let me know if you still have any problems with it.

JanPrusinowski: I have made changes to the keyword in: 2e122f2df5630f9b82de24c8073696d619eb36cf
Let me know it it is what you are looking for. Anyway I think that this overcomplicates things...

dasharo-stability/capsule-update-files/cbfstool
JanPrusinowski: I have changed the code acordingly in: 2710787ff4875ef1155a678094e31c76e173b8ae

dasharo-stability/capsule-update.robot
JanPrusinowski: yes but we have to check if the logo persists across the capsule update... so it is not actually the same.

dasharo-stability/capsule-update.robot
krystian-hebel: Starting from https://github.com/Dasharo/open-source-firmware-validation/commit/e7ab746fb5c8ed7ab99ba02fd985844182bd588e:

```diff
diff --git a/dasharo-stability/capsule-update.robot b/dasharo-stability/capsule-update.robot
index fca5bbe51e72..fd15402650a2 100644
--- a/dasharo-stability/capsule-update.robot
+++ b/dasharo-stability/capsule-update.robot
@@ -33,7 +33,7 @@ Suite Setup         Run Keywords
 ...                     AND
 ...                     Upload Required Files
 ...                     AND
-...                     Get System Values
+...                     Get System Values    $ORIGINAL_SERIAL    $ORIGINAL_UUID    $ORIGINAL_LOGO_SHA256
 ...                     AND
 ...                     Turn Off Active ME
 Suite Teardown      Run Keywords
@@ -115,38 +115,31 @@ CUP160.001 Verifying BIOS Settings Persistence After Update - PART 2
 
 CUP170.001 Verifying UUID
     [Documentation]    Check if UUID didn't change after Capsule Update.
-    Go To Ubuntu Prompt
+    Run Keyword If    $UPDATED_UUID is None    Get System Values    $UPDATED_UUID    $UPDATED_SERIAL    $UPDATED_LOGO_SHA256
+
     Log To Console    \n[Before Update] ${ORIGINAL_UUID}
-    ${updated_uuid}=    Get Firmware UUID
-    Log To Console    \n[After Update] ${updated_uuid}
+    Log To Console    \n[After Update] ${UPDATED_UUID}
 
-    Should Be Equal    ${ORIGINAL_UUID}    ${updated_uuid}
+    Should Be Equal    ${ORIGINAL_UUID}    ${UPDATED_UUID}
     IF    ${ROMHOLE_SUPPORT} == ${TRUE}
-        Should Be Equal    ${ORIGINAL_UUID}    00112233-4455-6677-8899-aabbccddeeff
-        Should Be Equal    ${updated_uuid}    00112233-4455-6677-8899-aabbccddeeff
+        Should Be Equal    ${UPDATED_UUID}    00112233-4455-6677-8899-aabbccddeeff
     END
 
 CUP180.001 Verifying Serial Number
     [Documentation]    Check if serial number didn't change after Capsule Update.
-    Go To Ubuntu Prompt
+    Run Keyword If    $UPDATED_SERIAL is None    Get System Values    $UPDATED_UUID    $UPDATED_SERIAL    $UPDATED_LOGO_SHA256
+
     Log To Console    \n[Before Update] ${ORIGINAL_SERIAL}
-    ${updated_serial}=    Get Firmware Serial Number
-    Log To Console    \n[After Update] ${updated_serial}
+    Log To Console    \n[After Update] ${UPDATED_SERIAL}
 
-    Should Be Equal    ${ORIGINAL_SERIAL}    ${updated_serial}
+    Should Be Equal    ${ORIGINAL_SERIAL}    ${UPDATED_SERIAL}
 
 CUP190.001 Verifying If Custom Logo Persists Across updates
     [Documentation]    Check if Logo didn't change after Capsule Update.
     Skip If    not ${CUSTOM_LOGO_SUPPORT}    not supported
-    Go To Ubuntu Prompt
-    ${out}=    Execute Command In Terminal
-    ...    sha256sum /sys/firmware/acpi/bgrt/image
-    ${unplugged}=    Run Keyword And Return Status
-    ...    Should Contain    ${out}    No such file
-    IF    ${unplugged} == ${TRUE}
-        Fail    Please make sure that a display device is connected to the DUT
-    END
-    Should Contain    ${out}    ${CUSTOM_LOGO_SHA}
+    Run Keyword If    $UPDATED_LOGO_SHA256 is None    Get System Values    $UPDATED_UUID    $UPDATED_SERIAL    $UPDATED_LOGO_SHA256
+
+    Should Be Equal    ${ORIGINAL_LOGO_SHA256}    ${UPDATED_LOGO_SHA256}
 
 
 *** Keywords ***
@@ -380,15 +373,11 @@ Go To Ubuntu Prompt
     Switch To Root User
 
 Get System Values
-    Log To Console    PREPARE: Get System Values
+    [Arguments]    ${var_serial}    ${var_uuid}    ${var_logo_sha256}
 
-    ${temp_serial}=    Get Firmware Serial Number
-    Set Suite Variable    ${ORIGINAL_SERIAL}    ${temp_serial}
-    Log To Console    \n[Serial Number Before Update] ${ORIGINAL_SERIAL}
+    Set Suite Variable    ${var_serial}    Get Firmware Serial Number
 
-    ${temp_uuid}=    Get Firmware UUID
-    Set Suite Variable    ${ORIGINAL_UUID}    ${temp_uuid}
-    Log To Console    [UUID Before Update] ${ORIGINAL_UUID}\n
+    Set Suite Variable    ${var_uuid}    Get Firmware UUID
 
     IF    $CUSTOM_LOGO_SUPPORT == $TRUE
         ${out}=    Execute Command In Terminal
@@ -398,7 +387,7 @@ Get System Values
         IF    ${unplugged} == ${TRUE}
             Fail    Please make sure that a display device is connected to the DUT
         END
-        Set Suite Variable    ${CUSTOM_LOGO_SHA}    ${out}
+        Set Suite Variable    ${var_logo_sha256}    ${out}
     END
 
 Prepare For ROMHOLE Persistence Test

```

krystian-hebel: Oh, now `Get System Values` doesn't boot to Linux, this would have to be added. Instead of all three `Run Keyword If` you would have to do something like:

```python
IF    $UPDATED_UUID is None
    Go To Ubuntu Prompt
    Get System Values    $UPDATED_UUID    $UPDATED_SERIAL    $UPDATED_LOGO_SHA256
END
```

krystian-hebel: Updated code in https://github.com/Dasharo/open-source-firmware-validation/pull/501#discussion_r1778344724, `$UPDATED_UUID` isn't `None` without `Get Variable Value`. I also messed up the order of arguments, this was also fixed in that comment.

SergiiDmytruk: I see serial number, but remember that custom romhole contents doesn't set it.  It's part of this structure which isn't trivial to construct and looks like it should be board-specific, so I left it filled with `0xff`.  At least for 
```c
struct smbios_type2_v3_4 {
	struct smbios_header header;
	u8 manufacturer;
	u8 product_name;
	u8 version;
	u8 serial_number;
	u8 asset_tag;
	u8 feature_flags;
	u8 location_in_chassis;
	u16 chassis_handle;
	u8 board_type;
	u8 contained_objects;
	u8 eos[2];
} __packed;
```
So for MSI serial number check should be skipped or we need to make this struct with somewhat valid data.

JanPrusinowski: I have made some changes acording to @krystian-hebel suggestions.
Also noticed that there is no boot to ubuntu - but I have changed the order of functions in prepare and the keyword should work now.
6a465ff4c5e22311bb4b39ea6c07d9fb1c1b8b67

krystian-hebel: @JanPrusinowski this is from previous version, I updated the comment later. UUID and serial have switched places.

dasharo-stability/capsule-update.robot
krystian-hebel: OK, then what is the difference?

JanPrusinowski: We are checking if the logo was not changed by the update.

SergiiDmytruk: Here `cbfstool` is used to check whether custom logo is supported.  Seems like you can just use `${CUSTOM_LOGO_SUPPORT}` for that.

JanPrusinowski: Removed this check in: 94c3fd7c4f176ca7c53085393f1b9e90b19b8a50

dasharo-stability/capsule-update.robot
JanPrusinowski: Also removed not needed resource file: 695a31bbc62ddc420a10a59b526859277c943535

JanPrusinowski:
After fixes at: 96ed8abdeda4d5913fc03502115b1f6a92ef209e : [log.zip](https://github.com/user-attachments/files/17203568/log.zip)

dasharo-stability/capsule-update.robot
```diff
@@ -140,19 +203,20 @@ Get BIOS Version
...
     Check If Capsule File Exists    ./dl-cache/edk2/${file_name}_max_fw_ver.cap
     Check If Capsule File Exists    ./dl-cache/edk2/${file_name}_wrong_cert.cap
     Check If Capsule File Exists    ./dl-cache/edk2/${file_name}_invalid_guid.cap
 
+    Set DUT Response Timeout    5m
     Power On
     Boot System Or From Connected Disk    ubuntu
-
     IF    '${DUT_CONNECTION_METHOD}' == 'pikvm'
         Set Global Variable    ${DUT_CONNECTION_METHOD}    SSH
```
krystian-hebel, line 218: Does it have to be global? `Go To Ubuntu Prompt` uses suite variable with the same name, and according to https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#variable-scopes only test variables can't be used in suite setup. If suite variable works here, please change this set of keywords to `Go To Ubuntu Prompt`, if not, please provide a log with failure.

krystian-hebel (changes requested):

dasharo-stability/capsule-update.robot
krystian-hebel: But now it doesn't change the logo, it just clones the repo and copies the file without any modification.

dasharo-stability/capsule-update.robot
JanPrusinowski: @krystian-hebel changed in: 3417ae2f8e267a5ad154a2654a4a5c84b563ed22

dasharo-stability/capsule-update.robot
```diff
@@ -278,3 +343,50 @@ Display Preparation Instructions
...
+        Set Suite Variable    ${DUT_CONNECTION_METHOD}    SSH
+    END
+    Login To Linux
+    Switch To Root User
+
+Get System Values
+    [Arguments]    ${var_serial}    ${var_uuid}    ${var_logo_sha256}
+
+    ${serial}=    Get Firmware Serial Number
+    Set Suite Variable    ${VAR_SERIAL}    ${serial}
```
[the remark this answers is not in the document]
JanPrusinowski: Precommit changes the size of letters in those variables

dasharo-stability/capsule-update.robot
JanPrusinowski: Ive cut too much of the code. Fixed in: 3e278bdb6a3fbe17c6ebfb35e06ad28c899525dc

krystian-hebel:
I've added few commits to deal with pre-commit and make it launch with QEMU. There are still some issues to resolve, not necessarily here and now:

- no "flashing", which causes some tests to not actually check anything
  - not sure is it is safe to copy FW file over while QEMU is running and then reset the machine
- unexpected passes
  - serial number - not actually changed for non-MSI, so default `123456789` is used
  - UUID (`Should Be Equal    'Not Settable'    'Not Settable'`), similar to above, both could be modified with `dcu`
  - logo isn't changed, yet the test passes, perhaps it should be skipped for QEMU until flashing works

dasharo-stability/capsule-update.robot
```diff
@@ -116,7 +115,7 @@ CUP160.001 Verifying BIOS Settings Persistence After Update - PART 2
 CUP170.001 Verifying UUID (Ubuntu)
     [Documentation]    Check if UUID didn't change after Capsule Update.
     ${tmp}=    Get Variable Value    $UPDATED_UUID
-    IF    $tmp is None
+    IF    '${tmp}' == 'None'
```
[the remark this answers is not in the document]
JanPrusinowski: `$tmp is None` doesnt work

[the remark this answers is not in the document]
krystian-hebel: Indeed, `Evaluating IF condition failed: Evaluating expression '$tmp is None' failed: SyntaxError: invalid syntax (<string>, line 1)`. Interestingly, this exact syntax is described in https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#special-variable-syntax:

> For example, if the variable value can be either a string or Python `None`, quoting like `'${var}'` is needed because otherwise strings do not work, but then `None` is interpreted to be a string as well. Luckily there is an easy solution to these problems discussed in this section.

`${tmp} is None` works fine for UUID and serial, but not for logo:
`Evaluating IF condition failed: Evaluating expression 'bf6cd62f794e18db7c18816ba38401204a9a2bf2532f201f3bfaee67608b8511  /sys/firmware/acpi/bgrt/image is None' failed: NameError: name 'bf6cd62f794e18db7c18816ba38401204a9a2bf2532f201f3bfaee67608b8511' is not defined nor importable as module`

Using `'${tmp}' is None` also won't work, because `'None'` isn't `None`, same for `'$tmp' is None`.

Doing comparison in inline Python (`${{$tmp is None}}` etc.) doesn't change anything, neither does passing it through `Evaluate`.

I guess this really is the best we can get from RF, our expectations exceeded the tool capabilities. It doesn't help that the documentation is wrong on the original version being a solution to this exact problem.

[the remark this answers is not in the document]
BeataZdunczyk: @SergiiDmytruk This thread is blocking the merge (specifically, rebase -> testing on MSI -> and then merging). Is this explanation enough for you?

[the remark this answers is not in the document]
SergiiDmytruk: Sure, it was just a question because the docs claim it should work.

krystian-hebel:
I've rebased and squashed smaller commits were possible. Some of them still have changes to multiple unrelated parts. I think the PR can be merged in this form, I'll just run it one more time to make sure nothing broke during rebase.

Created two issues for problems mentioned in previous comment: https://github.com/Dasharo/open-source-firmware-validation/issues/522 and https://github.com/Dasharo/open-source-firmware-validation/issues/523.

krystian-hebel (approved):

dasharo-stability/capsule-update.robot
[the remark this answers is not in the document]
pekkaklarck: @krystian-hebel, do you have a concrete example where `$tmp is None` doesn't work?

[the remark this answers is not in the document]
krystian-hebel: @pekkaklarck thanks for showing interest, it was unexpected. Out of curiosity, how did you find the mention of this problem here? :slightly_smiling_face: 

Simplest example:

```
*** Test Cases ***
TEST001 Tests suite variable
    Set Suite Variable    ${MY_VAR}    Some value
    IF    $MY_VAR is None
        Log To Console    MY_VAR is None
    ELSE
        Log To Console    MY_VAR is $MY_VAR
    END
```

Result:

```
$ robot --version
Robot Framework 5.0 (Python 3.12.7 on linux)
$ robot test.robot 
Test                                                                          
TEST001 Tests suite variable                                          | FAIL |
Evaluating IF condition failed: Evaluating expression '$MY_VAR is None' failed: SyntaxError: invalid syntax (<string>, line 1)
Test                                                                  | FAIL |
1 test, 0 passed, 1 failed
```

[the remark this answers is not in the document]
macpijan: I am afraid it might be fixed in newer versions:

```
python3 -m robot -h
Robot Framework -- A generic automation framework

Version:  7.1.1 (Python 3.12.7 on linux)
```

```
(venv) macpijan in /tmp/pp λ python3 -m robot ./test.robot 
Test                                                                          
TEST001 Tests suite variable                                          .MY_VAR is $MY_VAR
TEST001 Tests suite variable                                          | PASS |
Test                                                                  | PASS |
1 test, 1 passed, 0 failed
Output:  /tmp/pp/output.xml
Log:     /tmp/pp/log.html
Report:  /tmp/pp/report.html
```

We're still on the v5.0, could potentially upgrade (with the necessity to rebase some patches). The final goal is to use the upstream version, once we resolve the migration of  telnet lib https://github.com/MarketSquare/robotframework-telnet-library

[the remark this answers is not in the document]
pekkaklarck: The problem with `$MY_VAR is None` with RF 5.0 is that this release isn't fully compatible with Python 3.12 that you used. Subtle changes to Python's tokenizer broke the `$var` syntax as you can read from this issue:
https://github.com/robotframework/robotframework/issues/4771

Your options include upgrading to RF 6.1 or newer or downgrading to Python 3.11.

PR #511, opened 2024-09-20
automate test matrix documentation creation

philipanda:
A script that should automate or at least make it much faster to create `test-matrix` pages in docs.dasharo based on the test coverage for a platform.

### DIFF
--- a/scripts/create-dut-test-matrix.py
+++ b/scripts/create-dut-test-matrix.py
@@ -1,0 +1,138 @@
+#!/usr/bin/env python3
+
+import os
+import os.path
+import sys
+import pandas as pd
+import re
+import numpy as np
+
+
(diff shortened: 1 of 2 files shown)


philipandag:
Right now all supported tests are written from the test matrix to a markdown file in a copy-pastable format. 

The test-matrix pages on docs.dasharo feature links to the test specification for every test suite. It will require creating a mapping file and it's not implemented yet.

macpijan:
@BeataZdunczyk @philipanda was it connected to some task/issue?

Did we want to finalize that?

philipanda:
If I recall correctly I was supposed to create a test matrix and just assumed it would be faster to create a script like this to make a base for it than to write it manually.

PR #512, opened 2024-09-20
Increase rtcwake warmboot wait time

philipandag:
If the device takes longer than the time given to the command it will not wake up ever.

This change will increase the time it takes to perform this time every time. Even if the shutdown took only 1s the device will power on only after the specified time.

I don't have a good idea how to solve this unreliability other than increasing the delay...


mkopec (approved):
On platforms with an RTE we could hijack the power LED signal and wait for it to turn off, then we know the platform is off. On platforms with SSH access only, I have no idea how to do it better than this either :\

PR #515, opened 2024-09-23
dasharo-security/sort-pci-semiauto.robot: new test suite

wiktormowinski:


pietrushnic:
@wiktormowinski I'm very interested in what is going on here. Could you give me some background on this task? Why do we enable Dasharo (coreboot+SeaBIOS) here? It would be helpful to release: https://github.com/Dasharo/dasharo-issues/milestone/39

wiktormowinski:
oh ok now i see, the filename can be misleading, it's not sortboot related (not now at least) i got the old sortboot file from the old gitlab repo and just stared working on it, because there were some useful lines of code to put in this project. I' ll fix that right away to prevent further confusion.
Sorry to disappoint, it was not the milestone

pietrushnic:
Excellent, I thought it was something about PC Engines :) Thanks for the clarification.

dasharo-security/sort-pci-semiauto.robot
```diff
@@ -0,0 +1,79 @@
...
+
+Suite Setup         Run Keywords
+...                     Prepare Test Suite
+Suite Teardown      Run Keyword
+...                     Log Out And Close Connection
+
+
+*** Variables ***
+${LSPCI_CMD}=       lspci -vv
+${IP_LINK_CMD}=     ip link
```
miczyg1, line 22: It doesn't make sense to put those commands into variables. It only obfuscates what is really done by the test. It would be justified if it was some really long command.

dasharo-security/sort-pci-semiauto.robot
```diff
@@ -0,0 +1,79 @@
...
+    ${pci_devices}=    Create Dictionary
+    ${current_device}=    Set Variable    None
+    FOR    ${line}    IN    @{lines}
+        ${is_device_line}=    Evaluate    'Ethernet controller' in '${line}'
+        IF    ${is_device_line}
+            ${current_device}=    Get Device Bus Info    ${line}
+        END
+        ${is_mac_line}=    Evaluate    'Device Serial Number' in '${line}'
+        IF    ${is_mac_line}
+            Append PCIe MAC    ${line}    ${pci_devices}    ${current_device}
```
miczyg1, line 52: We don't actually need to store the exact bus number. All we need is the ascending order of the devices/MACs sorted by the bus number. As we parse the lspci output line by line, we will get the ascending order already.

dasharo-security/sort-pci-semiauto.robot
```diff
@@ -0,0 +1,79 @@
...
+
+*** Keywords ***
+Get PCIe Bus Info With MACs
+    [Documentation]    Extract PCIe bus numbers and MAC addresses from lspci output.
+    ${lspci_output}=    Execute Command In Terminal    ${LSPCI_CMD}
+    ${lines}=    Split String    ${lspci_output}    \n
+    ${pci_devices}=    Create Dictionary
+    ${current_device}=    Set Variable    None
+    FOR    ${line}    IN    @{lines}
+        ${is_device_line}=    Evaluate    'Ethernet controller' in '${line}'
```
miczyg1, line 46: There might be multiple Ethernet controllers in the system of different type. It is the case for VP66xx at least. To be specific, we may check for a specific controller, which could be defined by some variable. If the variable is not found, then we may look for `Ethernet controller` as a fallback.

platform-configs/protectli-v1210.robot
```diff
@@ -10,3 +10,7 @@ ${CPU_MIN_FREQUENCY}=           800
 ${PLATFORM_CPU_SPEED}=          2.00
 ${PLATFORM_RAM_SPEED}=          2933
 ${PLATFORM_RAM_SIZE}=           4096
+
+# List of ethernet interfaces
+&{ETH_PORTS}=                   eno0:=64-62-66-2f-00-12
+...                             eno1:=64-62-66-2f-00-13
```
miczyg1, line 16: We should only care about MACs. MACs are immutable, but the interface names are and the test aims to detect this. We cannot bind MACs to enoX. We just need the order of MACs to appear in a specific order dictated by the labels on the chassis.
```suggestion
&{ETH_PORTS}=                   64-62-66-2f-00-12
...                             64-62-66-2f-00-13
```

Apply it to the other configs as well and modify test accordingly.

miczyg1 (changes requested):

platform-configs/protectli-v1210.robot
wiktormowinski: makes sense https://github.com/Dasharo/open-source-firmware-validation/pull/515/commits/82760ad9e19d8348e16b3bda7cab450ad73a8418

dasharo-security/sort-pci-semiauto.robot
wiktormowinski: updated test https://github.com/Dasharo/open-source-firmware-validation/pull/515/commits/b8b54f5ba57824d5099809577c355ff6c8daf6cb

wiktormowinski:
@miczyg1 i addressed the requests mentioned above by changing:
- removed the enoX-es form configs as well as the operation of reading and storing them in functions
- moved commands from variables to keywords directly

dasharo-security/sort-pci-semiauto.robot
wiktormowinski: the test checks the MACs and config list: https://github.com/Dasharo/open-source-firmware-validation/pull/515/commits/306e90f5e3d77337b3488d4b1832240b4b97bb86 is it enough?

miczyg1: We don't store bus numbers now, however the keyword names still have the word bus  in their names. Will open new comments for that.

dasharo-security/sort-pci-semiauto.robot
miczyg1: I think you don't understand... Let me illustrate that for you. This is what I do for an lspci output from VP66:

```
cat lspci.log |grep 'Ethernet controller'
02:00.0 Ethernet controller [0200]: Intel Corporation Ethernet Controller X710 for 10GbE SFP+ [8086:1572] (rev 02)
02:00.1 Ethernet controller [0200]: Intel Corporation Ethernet Controller X710 for 10GbE SFP+ [8086:1572] (rev 02)
03:00.0 Ethernet controller [0200]: Intel Corporation Device [8086:125c] (rev 04)
04:00.0 Ethernet controller [0200]: Intel Corporation Device [8086:125c] (rev 04)
05:00.0 Ethernet controller [0200]: Intel Corporation Device [8086:125c] (rev 04)
06:00.0 Ethernet controller [0200]: Intel Corporation Device [8086:125c] (rev 04)
```

As you can see there are two SFP ports, which are labeled separately from the ethernet port on chassis. We only want to check the ethernet ones, so you have to actually check for specific devices.

As a last resort you may add the SFP MACs as well to the config,  but then can we say that the test does what it is supposed to do? No, because it checks ports which are not supposed to be checked. There may be machines with different ethernet chips, which may not be internal (externally connected cards).

So what the test needs to do is t additionally check for the device name or ID which is of interest to us. In the example above it would be `8086:125c`

dasharo-security/sort-pci-semiauto.robot
```diff
@@ -0,0 +1,60 @@
...
+    Boot System Or From Connected Disk    ubuntu
+    Login To Linux
+    Switch To Root User
+    ${pci_devices}=    Get PCIe Bus Info With MACs
+    Log    PCIe devices and their MACs: ${pci_devices}
+    Compare Interface And PCIe Bus Order    ${pci_devices}
+
+
+*** Keywords ***
+Get PCIe Bus Info With MACs
```
miczyg1, line 34: The keyword no longer collects PCIe Buses, so its name has to change.

dasharo-security/sort-pci-semiauto.robot
```diff
@@ -0,0 +1,60 @@
...
+
+Append PCIe MAC
+    [Arguments]    ${line}    ${pci_devices}
+    ${serial}=    Evaluate    '${line}'.split()[-1]
+    ${mac}=    Replace String    ${serial}    ff-ff-    ${EMPTY}
+    IF    '00-00-00-00-00-00' not in '${mac}'
+        Append To List    ${pci_devices}    ${mac}
+    END
+
+Compare Interface And PCIe Bus Order
```
miczyg1, line 53: The keyword does not compare PCIe bus order, just MAC order. Name has to change

dasharo-security/sort-pci-semiauto.robot
wiktormowinski: I just added the SFP because they and regular ethernet looked alike, no wonder i got confused. Thanks for your extra time to explain this. I guess I'll just skip the SFP ones and this should be it: https://github.com/Dasharo/open-source-firmware-validation/pull/515/commits/8cdab97d10bb5f5312cdfc8926f3d8607ca704ae

dasharo-security/sort-pci-semiauto.robot
wiktormowinski: https://github.com/Dasharo/open-source-firmware-validation/pull/515/commits/b319dd02a7c842be46442a66134fee681ae30d77

dasharo-security/sort-pci-semiauto.robot
wiktormowinski: https://github.com/Dasharo/open-source-firmware-validation/pull/515/commits/b319dd02a7c842be46442a66134fee681ae30d77

dasharo-security/sort-pci-semiauto.robot
miczyg1: > I just added the SFP because they and regular ethernet looked alike, no wonder i got confused.

Yes, there are SFP to Ethernet adapters plugged in there, thus it looks the same... We do not have many means to test fibre connection, thus as a workaround we use the Ethernet plugs.

miczyg1: > I guess I'll just skip the SFP ones and this should be it

Yes and no... Liek I have mentioned, you may find different network cards in the system, e.g. some external Ethernet Controllers which may not be SFP. So the test may not always catch this. To be sure, I would like the test to check for the device ID we are interested in, like in the comment above.

wiktormowinski: ok i see, all protectli v1000 and VP6670 has the same ethernet controller id (8086:125c), this test looks for it and ignores any other line: https://github.com/Dasharo/open-source-firmware-validation/pull/515/commits/aef36b8ddc07b10943d7e4b55cc4f1d9c7f11984

dasharo-compatibility/check-ethernet-ports-order.robot
```diff
@@ -0,0 +1,71 @@
...
+*** Keywords ***
+Get MACs
+    [Documentation]    Extract MAC addresses from lspci output.
+    ${lspci_output}=    Execute Command In Terminal    lspci -vvnn
+    ${lines}=    Split String    ${lspci_output}    \n
+    ${pci_devices}=    Create List
+
+    ${sfp_flag}=    Set Variable    ${FALSE}
+    FOR    ${line}    IN    @{lines}
+        IF    '[8086:125c]' in '${line}'
```
miczyg1, line 42: This ID should be a variable settable from platform config. note that the controller IDs may change in the future (sometimes happens with new revisions of HW)

dasharo-compatibility/check-ethernet-ports-order.robot
```diff
@@ -0,0 +1,71 @@
...
+            ${sfp_flag}=    Set Variable    ${TRUE}
```
miczyg1, line 43: Why this SFP flag is even used? When we match by ID, we don't need to check for SFP at all...

miczyg1 (changes requested):

dasharo-compatibility/check-ethernet-ports-order.robot
wiktormowinski: the naming is off but i believe the functionality of this flag should remain the same: https://github.com/Dasharo/open-source-firmware-validation/pull/515/commits/8b37fc9d8fc2d98ce6614a15bc9af8d905024f70

dasharo-compatibility/check-ethernet-ports-order.robot
wiktormowinski: moved to configs: https://github.com/Dasharo/open-source-firmware-validation/pull/515/commits/8b37fc9d8fc2d98ce6614a15bc9af8d905024f70

dasharo-compatibility/check-ethernet-ports-order.robot
miczyg1: We do not need that flag at all, that's what I mean, Please remove it.

platform-configs/protectli-v1210.robot
```diff
@@ -10,3 +10,8 @@ ${CPU_MIN_FREQUENCY}=           800
 ${PLATFORM_CPU_SPEED}=          2.00
 ${PLATFORM_RAM_SPEED}=          2933
 ${PLATFORM_RAM_SIZE}=           4096
+
+# List of ethernet interfaces
+@{ETH_PORTS}=                   64-62-66-2f-00-12
+...                             64-62-66-2f-00-13
+${ETHERNET_ID}=                 [8086:125c]
```
miczyg1, line 17:
```suggestion
${ETHERNET_ID}=                 8086:125c
```

And everywhere else.

miczyg1 (changes requested):

dasharo-security/sort-pci-semiauto.robot
miczyg1: Yes, we can drop it

dasharo-compatibility/check-ethernet-ports-order.robot
wiktormowinski: id is not within the same line as serial number though, that way we read only the ethernet ones

platform-configs/protectli-v1210.robot
wiktormowinski: https://github.com/Dasharo/open-source-firmware-validation/pull/515/commits/efd955f8de238bd905c88dd7c8d242ea1beb15dd

dasharo-compatibility/check-ethernet-ports-order.robot
wiktormowinski: https://github.com/Dasharo/open-source-firmware-validation/pull/515/commits/51db74b9011234584aaff0fe83bb124365766b35

dasharo-compatibility/check-ethernet-ports-order.robot
```diff
@@ -0,0 +1,69 @@
...
+    ${serial}=    Evaluate    '${line}'.split()[-1]
+    ${mac}=    Replace String    ${serial}    ff-ff-    ${EMPTY}
+    IF    '00-00-00-00-00-00' not in '${mac}'
+        Append To List    ${pci_devices}    ${mac}
+    END
+
+Compare Interfaces
+    [Arguments]    ${pci_devices}
+    Log    Sorted interfaces: ${ETH_PORTS} \n
+    Log    Sorted PCIe devices: ${pci_devices} \n
```
miczyg1, line 65:
```suggestion
```

No need to log it. The lists will be visible anyways in the `Should Be Equal As Strings` keyword in the log.

platform-configs/include/protectli-common.robot
```diff
@@ -1,5 +1,6 @@
 *** Settings ***
 Resource    default.robot
+Resource    ../../keys.robot
```
miczyg1, line 3: Is this needed? I don't see any of the keys being used in the new test

miczyg1 (changes requested):

platform-configs/include/protectli-common.robot
wiktormowinski:
```
Check-Ethernet-Ports-Order                                                    
[ ERROR ] Error in file '/home/wmowinski/open-source-firmware-validation/platform-configs/include/protectli-common.robot' on line 12: Setting variable '${BOOT_MENU_KEY}' failed: Variable '${F11}' not found.
SPS001.001 Ethernet ports are in order :: This test automates the ... | FAIL |
Variable '${BOOT_MENU_KEY}' not found.
Check-Ethernet-Ports-Order                                            | FAIL |
1 test, 0 passed, 1 failed
```
apparently, it is

dasharo-compatibility/check-ethernet-ports-order.robot
wiktormowinski: ok, https://github.com/Dasharo/open-source-firmware-validation/pull/515/commits/35fd5151cb4fae97bcf4f53aaacdbdea03a74325

platform-configs/include/protectli-common.robot
miczyg1: `keys.robot` does not contain `BOOT_MENU_KEY` variable. `BOOT_MENU_KEY` should be defiend i nthe paltform config.

miczyg1: `keys.robot` is only included as a resource in the tests, not platform configs currently. Please add it to `dasharo-security/sort-pci-semiauto.robot` to keep consistency with other tests.

wiktormowinski: https://github.com/Dasharo/open-source-firmware-validation/pull/515/commits/18d1c5f8639fae1cddb4cd28aaa9e25382e1d8ff

PR #518, opened 2024-09-29
Some MSI fixes

SergiiDmytruk:
Some commits here should be useful, but they don't even remotely let regression tests pass, although part of it is probably setup (like missing Windows on the platform I used).  Actual release binary can't be tested because serial output is reset to its default (off) state in multiple places.

I've probably never tried to run so many tests and can say they are extremely slow even when skipped.  When suite setup skips all tests, teardown still does things like flashing or starting the machine and changing EDK settings.

See commit messages.

### DIFF
--- a/lib/flash.robot
+++ b/lib/flash.robot
@@ -73,7 +73,7 @@ Flash Firmware
     ...    chip size.
     [Arguments]    ${fw_file}
     ${file_size}=    Run    ls -l ${fw_file} | awk '{print $5}'
-    IF    '${file_size}'!='${FLASH_SIZE}'
+    IF    '''${file_size}''' != '''${FLASH_SIZE}'''
         FAIL    Image size doesn't match the flash chip's size!
     END
 
--- a/dasharo-compatibility/apu-configuration-menu.robot
+++ b/dasharo-compatibility/apu-configuration-menu.robot
@@ -18,9 +18,9 @@ Suite Setup         Run Keywords
 ...                     Prepare Test Suite    AND
 ...                     Skip If    not ${APU_CONFIGURATION_MENU_SUPPORT}    APU configuration tests not supported.
 Suite Teardown      Run Keywords
-...                     Flash Firmware    ${FW_FILE}
+...                     Skip If    '${SUITE_STATUS}' == 'SKIP'    Skipping Teardown since Suite was skipped as well
 ...                     AND
-...                     Turn On Power Supply
+...                     Flash Firmware    ${FW_FILE}
 ...                     AND
 ...                     Log Out And Close Connection
 
--- a/dasharo-compatibility/apu-configuration-menu.robot
+++ b/dasharo-compatibility/apu-configuration-menu.robot
@@ -20,6 +20,8 @@ Suite Setup         Run Keywords
 Suite Teardown      Run Keywords
 ...                     Flash Firmware    ${FW_FILE}
 ...                     AND
+...                     Turn On Power Supply
+...                     AND
 ...                     Log Out And Close Connection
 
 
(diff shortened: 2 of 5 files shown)


lib/flash.robot
```diff
@@ -92,6 +92,9 @@ Flash Firmware
         Fail    Flash firmware not implemented for platform config ${CONFIG}
     END
 
+    # Don't leave the platform in a state in which it won't react to tests
+    Turn On Power Supply
```
JanPrusinowski, line 96: If you want DUT to power on after flashing it would be better to use: `Power Cycle On` kwd. As `Turn On Power Suply` does only turn the power on and does not switch the power button by RTE.

Also is there a need for adding this? As I think all of tests which use this kwd and do something to the DUT actually have 'Power On' or similar kwd after flashing?

SergiiDmytruk: > If you want DUT to power on after flashing it would be better to use: `Power Cycle On` kwd. As `Turn On Power Suply` does only turn the power on and does not switch the power button by RTE.

That was the intent.
 
> Also is there a need for adding this? As I think all of tests which use this kwd and do something to the DUT actually have 'Power On' or similar kwd after flashing?

I don't  know if this is really necessary to have, but as commit message says, this was a problem with running regression tests because `apu-configuration-menu.robot` turned of power supply in teardown for flashing firmware and the tests following it timed out.  Updating `apu-configuration-menu.robot` might be a better idea, I just wanted to avoid this stopping tests multiple times.

JanPrusinowski: So this does fix only one test/suite? Its better to edit the test then. This kwd is used in various other tests and adding power on potentially might break those tests.

lib/flash.robot
```diff
@@ -73,7 +73,7 @@ Flash Firmware
     ...    chip size.
     [Arguments]    ${fw_file}
     ${file_size}=    Run    ls -l ${fw_file} | awk '{print $5}'
-    IF    '${file_size}'!='${FLASH_SIZE}'
+    IF    int($file_size) != $FLASH_SIZE
```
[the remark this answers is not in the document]
SergiiDmytruk: Can be changed to use `'''` instead, but what's your `robot` version?  I'd expect it to match mine in venv:
```
$ robot --version
Robot Framework 5.0 (Python 3.11.10 on linux)
```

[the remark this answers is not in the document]
JanPrusinowski: Should work the same:
image

lib/flash.robot
JanPrusinowski: I think that option 1 is the correct one. As it only involves the tests that you had an issue with.

lib/flash.robot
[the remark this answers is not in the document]
SergiiDmytruk: Just changed it to `'''` in https://github.com/Dasharo/open-source-firmware-validation/pull/518/commits/90482b6bda1a4d1665c359443c10b56c632a3c2a

lib/flash.robot
SergiiDmytruk: Updated that commit: https://github.com/Dasharo/open-source-firmware-validation/pull/518/commits/6d10119365616c62c408a36caa65259e2c8610f9

> I'm thinking that it would be a good idea to add some skip mechanism in suite teardown if whole suite is skipped... To avoid situations like this. And avoid unnecesary flashing of fw when running regression script.

RF seems to be doing it intentionally for cleanup opportunity and there is no way to check that the test was skipped, so it either need to be a flag or a check for whether to skip the tests should be moved into a separate keyword.

macpijan (approved):

macpijan:
Picked to develop. For some reason GH has not marked it as merged this time.

image

SergiiDmytruk:
Commit hashes here and on `develop` differ, so it wasn't a fast-forward.

PR #527, opened 2024-10-08
Added sanity check for empty string in Cache Tests

JanPrusinowski:


dasharo-compatibility/cpu-status.robot
```diff
@@ -23,7 +23,7 @@ Suite Setup         Run Keywords
 ...                     AND
 ...                     Power On    AND
 ...                     Enter Setup Menu Tianocore    AND
-...                     Reset To Defaults Tianocore    AND
+# ...               Reset To Defaults Tianocore    AND
```
miczyg1, line 26: Rather not intended change?

miczyg1 (changes requested):

miczyg1:
@JanPrusinowski would be great to have a test log from a bad case (before fix) and good case (after fix)

dasharo-compatibility/cpu-status.robot
JanPrusinowski: Yes. Commented only for testing purposes. Fixed: 0eb2859d045c806e608c7da819b8302de0e9775c

JanPrusinowski:
> @JanPrusinowski would be great to have a test log from a bad case (before fix) and good case (after fix)

Logs for before and after.
[log_negative_path_after_sanity_check.zip](https://github.com/user-attachments/files/17327014/log_negative_path_after_sanity_check.zip)
[log_negative_path_before_sanity_check.zip](https://github.com/user-attachments/files/17327016/log_negative_path_before_sanity_check.zip)

miczyg1:
Ok so the problem here is different. Cache size can be zero if given cache level is not supported or not present in HW. The proper fix would be to set L4_CACHE_SUPPORT variable properly for all platforms where L4 cache is expected to be present...

JanPrusinowski:
> Ok so the problem here is different. Cache size can be zero if given cache level is not supported or not present in HW. The proper fix would be to set L4_CACHE_SUPPORT variable properly for all platforms where L4 cache is expected to be present...

Do you know of any such platforms which have this variable set improperly?

miczyg1:
> > Ok so the problem here is different. Cache size can be zero if given cache level is not supported or not present in HW. The proper fix would be to set L4_CACHE_SUPPORT variable properly for all platforms where L4 cache is expected to be present...
> 
> Do you know of any such platforms which have this variable set improperly?

No. But as far as I know, there is no platform in our lab that has L4 cache. Max is L3.

Only Talos II might have L4 via Centaurs, but not sure. @krystian-hebel might know

krystian-hebel:
> Only Talos II might have L4 via Centaurs

Talos II has Nimbus chips, so directly attached memory only, no Centaurs.

miczyg1:
> > Only Talos II might have L4 via Centaurs
> 
> Talos II has Nimbus chips, so directly attached memory only, no Centaurs.

So no platforms in our lab should have L4. Max is L3, but note that not all platforms have L3 too.

platform-configs/include/default.robot
```diff
@@ -143,6 +143,7 @@ ${CPU_TESTS_SUPPORT}=                               ${FALSE}
 ${HYPER_THREADING_SUPPORT}=                         ${FALSE}
 ${INTEL_HYBRID_ARCH_SUPPORT}=                       ${FALSE}
 ${RESET_TO_DEFAULTS_SUPPORT}=                       ${FALSE}
+${L2_CACHE_SUPPORT}=                                ${FALSE}
```
miczyg1, line 146: It was supposed to be TRUE?

platform-configs/raptor-cs_talos2.robot
```diff
@@ -105,6 +105,8 @@ ${RESET_TO_DEFAULTS_SUPPORT}=                       ${FALSE}
 ${MEMORY_PROFILE_SUPPORT}=                          ${FALSE}
 ${DEFAULT_POWER_STATE_AFTER_FAIL}=                  Powered Off
 ${ESP_SCANNING_SUPPORT}=                            ${FALSE}
+${L2_CACHE_SUPPORT}=                                ${TRUE}
```
miczyg1, line 108: If the default becomes true, then it is not needed here.

miczyg1 (changes requested):
Many more platforms should have L3 cache support. YOu should check them too.

L3 should be also on Protectli (vp46xx, VP66xx) platforms, MSI board and probably others too. Please check them

platform-configs/include/default.robot
JanPrusinowski: removed in: 2cde25bac3dd60533fbc565a0ca6010772a8d1b6

platform-configs/raptor-cs_talos2.robot
JanPrusinowski: removed in: 2cde25bac3dd60533fbc565a0ca6010772a8d1b6 as it is no longer needed if the default is set to true

miczyg1 (approved):

PR #532, opened 2024-10-11
trenchboot: remove sonoff-api and rtectrl inclusion

m-iwanicki:
@EduKav1813 As you are running these tests currently please say if those changes worked. And maybe post console output after running tests.


EduKav1813 (approved):
This solves the problem of missing files, but more fixes are needed to fully fix those tests.

EduKav1813:
[log.zip](https://github.com/user-attachments/files/17342777/log.zip)

Logs from attempted execution

m-iwanicki:
Looks like `root@tb:~#` was interpreted as

```text
r
oot@tb:~#
```

Not sure why

image

macpijan (approved):

Issue #535, opened 2024-09-13
V540TND v0.9.1-rc5 relatively long boot times


philipandag:
### Component

Dasharo firmware

### Device

NovaCustom V54 14th Gen

### Dasharo version

v0.9.1-rc5

### Dasharo Tools Suite version

develop

### Test case ID

CBMEM002, CBMEM003

### Brief summary

Very long boot times, exceeding the limits given in platform configs

### How reproducible

I have only run the suite once.

### How to reproduce

Run CBMEM tests via osfv

### Expected behavior

Average boot times should be < 10s.

### Actual behavior

 In this case most are ~7-8s but because of a large spike on the first boot time the average warmboot time exceeded 10s.

### Screenshots

[dasharo-performance_log.zip](https://github.com/user-attachments/files/16989613/dasharo-performance_log.zip)

### Additional context

I have seen large boot time spikes on the first measurement before. Maybe it's an error in measurements?

wiktormowinski:
https://github.com/Dasharo/open-source-firmware-validation/pull/538 done, the new version skips the first measurement

PR #538, opened 2024-10-17
dasharo-performance/boot-time-measure.robot:ignore first element

wiktormowinski:
This PR will fix the problem appearing with some NovaCustom MTL machines where the first boot time is much longer (as expected) and the test takes it as fail 

Issue: https://github.com/Dasharo/open-source-firmware-validation/issues/535

### DIFF
--- a/dasharo-performance/boot-time-measure.robot
+++ b/dasharo-performance/boot-time-measure.robot
@@ -26,7 +26,7 @@ Suite Teardown      Run Keyword
 
 
 *** Variables ***
-${ITERATIONS}=      6
+${ITERATIONS}=      5
 
 
 *** Test Cases ***
--- a/dasharo-performance/boot-time-measure.robot
+++ b/dasharo-performance/boot-time-measure.robot
@@ -26,7 +26,7 @@ Suite Teardown      Run Keyword
 
 
 *** Variables ***
-${ITERATIONS}=      5
+${ITERATIONS}=      6
 
 
 *** Test Cases ***
--- a/lib/cbmem.robot
+++ b/lib/cbmem.robot
@@ -39,7 +39,7 @@ Calculate Boot Time Statistics
     ${max}=    Set Variable    0
     ${average}=    Set Variable    0
 
-    FOR    ${index}    IN RANGE    1    ${iterations}
+    FOR    ${index}    IN RANGE    0    ${iterations}
         ${duration}=    Get From List    ${samples}    ${index}
         ${min}=    Evaluate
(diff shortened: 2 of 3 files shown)


macpijan:
@wiktormowinski Can you please provide reasoning behind this change in the long commit description?

dasharo-performance/boot-time-measure.robot
```diff
@@ -26,7 +26,7 @@ Suite Teardown      Run Keyword
 
 
 *** Variables ***
-${ITERATIONS}=      5
+${ITERATIONS}=      6
```
miczyg1, line 29: I would leave it at 5.  Why? Because it is the desired amount of iterations you want to measure. It should be independent of test workarounds.

miczyg1:
Instead of modifyign the `Calculate Boot Time Statistics` I would change the
Measure Coldboot Time/Measure Warmboot Time/Measure Reboot Time in the following way:

```
Measure Coldboot Time
    [Documentation]    Performs a measurement of coldboot
    ...    boot time
    [Arguments]    ${iterations}
    ${durations}=    Create List
    Log To Console    \n
        Power Cycle On    power_button=${TRUE}
        Login To Linux
        Switch To Root User
        ${boot_time}=    Get Boot Time From Cbmem
        Log To Console    (${index}) Boot time: ${boot_time} s
        # Skip appending first result, as it may be the first boot which is longer
        IF    ${index} > 0
            Append To List    ${durations}    ${boot_time}
        END
    END
    ${min}    ${max}    ${average}    ${stddev}=
    ...    Calculate Boot Time Statistics    ${durations}
    RETURN    ${min}    ${max}    ${average}    ${stddev}
```

miczyg1 (changes requested):

dasharo-performance/boot-time-measure.robot
wiktormowinski: let's keep it tight then: https://github.com/Dasharo/open-source-firmware-validation/pull/538/commits/0c0b1a1df098656fab25619c4ef6a3d4f4cbec20

miczyg1:
I would add the comments:
`# Do one more iteration than requested, as we may hit first boot which is always longer.`
`# Skip appending first result, as it may be the first boot which is longer`

To all keywords doing the measurement fo consistency and to avoid any confusion.

wiktormowinski:
alright @miczyg1 , comments added for consistency https://github.com/Dasharo/open-source-firmware-validation/pull/538/commits/2ee0f841239c8211a9f9214943f64e1f056ac567

miczyg1 (approved):

PR #550, opened 2024-10-23
Regression msi

JanPrusinowski:


JanPrusinowski:
> @JanPrusinowski Do you expect that any more fixes are needed after reviewing my results from yesterday? Or after your full regression run.

@macpijan I think that this is ok. I have run few tests which failed in your run due to connection issues or similar cases to verify if they run correctly - and they do.

keywords.robot
```diff
@@ -1733,14 +1733,10 @@ Check That USB Devices Are Detected
...
 
-    IF    '${DUT_CONNECTION_METHOD}' == 'pikvm'
-        Should Contain Match    ${boot_menu}    *PiKVM*
-    ELSE
-        @{attached_usb_list}=    Get Current CONFIG List Param    USB_Storage    name
-        FOR    ${stick}    IN    @{attached_usb_list}
-            # ${stick} should match with one element of ${boot_menu}
-            Should Contain Match    ${boot_menu}    *${stick}*
-        END
+    @{attached_usb_list}=    Get Current CONFIG List Param    USB_Storage    name
```
macpijan, line 1736: @JanPrusinowski As @krystian-hebel has noticed before, when removing this pikvm branch (which we agreed to on the call), we must update the documentation string as well.

JanPrusinowski: Removed in: 7cde5cbb39b651db442620c425fe257f341b2e39

PR #551, opened 2024-10-23
scripts/secure-boot/generate-images/generate-image.sh: Unify name evals

SebastianCzapla:
On Fedora systems, script fails, as it expects "fedora" but it gets "Fedora" instead.


SebastianCzapla:
This was the quick solution, adding additional pipe, instead of echo + pipe in separate line fails to change the capitalization of letters

SebastianCzapla:
Wider scope of changes, that includes the same fix, available here https://github.com/Dasharo/open-source-firmware-validation/pull/368

PR #555, opened 2024-10-28
TrenchBoot updates

SergiiDmytruk:
Previously only non-DRTM case was tested, now DRTM one works as well.  It requires SeaBIOS which is harder to automate, see README file for how to use it.

See commit messages for more info.

### DIFF
--- a/lib/bios/menus.robot
+++ b/lib/bios/menus.robot
@@ -572,6 +572,15 @@ Boot System Or From Connected Disk
     ...    to boot from connected disk set up in config
     [Arguments]    ${system_name}
     IF    '${DUT_CONNECTION_METHOD}' == 'SSH'    RETURN
+
+    IF    '''${SEABIOS_BOOT_DEVICE}''' != ''
+        Read From Terminal Until    Press F10 key now for boot menu
+        Write Bare Into Terminal    ${F10}
+        Read From Terminal Until    Select boot device
+        Write Bare Into Terminal    ${SEABIOS_BOOT_DEVICE}
+        RETURN
+    END
+
     ${menu_construction}=    Enter Boot Menu Tianocore And Return Construction
     # When ESP scanning feature is there, boot entries are named differently than
     # they used to
--- a/trenchboot/02-with-drtm.robot
+++ b/trenchboot/02-with-drtm.robot
@@ -100,7 +100,54 @@ WTD005.001 DRTM event log has DRTM entries (AMD)
 WTD006.001 DRTM log aligns with PCR values
     [Documentation]    Verify that all DRTM measurements are correctly reflected
     ...    in DRTM event log
-    Validate PCRs Against Event Log    /sys/kernel/security/slaunch/eventlog
+    Execute Command In Terminal
+    ...    tpm2_eventlog /sys/kernel/security/slaunch/eventlog > /tmp/event-log
+    # This extracts hashes for a combination of a PCR (first parameter) and an
+    # algorithm (second parameter). The printed line have the form
+    #    {event data length}-{digest}
+    # to make it possible to ignore some events based on a combination of these
+    # fields.
+    #
+    # Backslash in front of some dollar signs is escaping AWK fields while
+    # unescaped ones are parameters of this function that get embedded into
+    # AWK script. The escaping is doubled because Robot Framework also
+    # interprets escapes.
+    ${hashes_func}=    Catenate    SEPARATOR=${SPACE}    function hashes() {
+    ...    awk "/PCRIndex:/{ pcr = \\$2; alg = \\"\\"; digest = \\"\\" }
+    ...    /- AlgorithmId:/{ alg = \\$3 }
+    ...    /Digest:/{ if (alg == \\"$2\\") digest = substr(\\$2, 2, length(\\$2) - 2) }
+    ...    /EventSize:/{ if (pcr == $1 && length(digest) != 0) print \\$2 \\"-\\" digest }"
+    ...    /tmp/event-log;
+    ...    }
+    Execute Command In Terminal    ${hashes_func}
+
+    Execute Command In Terminal
+    ...    function extend() { pcr=$(echo -n "$pcr$2" | xxd -r -p | $1sum | cut -f1 -d' '); }
+    FOR    ${algo}    IN    sha1    sha256
+        ${algo_present}=    Execute Command In Terminal
+        ...    test -d /sys/class/tpm/tpm0/pcr-${algo} && echo y
+        IF    '''${algo_present}''' != 'y'    CONTINUE
+
+        FOR    ${pcr}    IN    17    18
+            # Replaying the log skips digests of no data. These correspond to
+            # informational markers which don't extend PCRs.
+            ${replay_pcr}=    Catenate    SEPARATOR=${SPACE}
+            ...    empty=$(${algo}sum < /dev/null | cut -f1 -d' ');
+            ...    pcr=\${empty//?/0};
+            ...    for line in $(hashes ${pcr} ${algo}); do
+            ...    len=\${line%-*};
+            ...    hash=\${line#*-};
+            ...    if [ "\${hash^^*}" != "\${empty^^*}" -o "$len" -ne 0 ]; then extend ${algo} "$hash"; fi;
+            ...    done;
+            ...    echo "$pcr"
+            ${expected}=    Execute Command In Terminal    ${replay_pcr}
+            ${actual}=    Execute Command In Terminal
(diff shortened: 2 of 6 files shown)


pietrushnic:
@SergiiDmytruk, I have to automate Dasharo (coreboot+SeaBIOS) on a large scale for the upcoming PC Engine releases. @macpijan, I'm looking for your input on whether you see `Boot SeaBIOS` as a valid approach or if you have some ideas about more sophisticated infrastructure.

SergiiDmytruk:
@pietrushnic Changed it in https://github.com/Dasharo/open-source-firmware-validation/pull/555/commits/40c9d5f7ea9bf3e0d7d2d47586b9181ef6ece262, which is more reasonable and can be at least a temporary solution.

lib/bios/menus.robot
```diff
@@ -572,6 +572,15 @@ Boot System Or From Connected Disk
     ...    to boot from connected disk set up in config
     [Arguments]    ${system_name}
     IF    '${DUT_CONNECTION_METHOD}' == 'SSH'    RETURN
+
+    IF    '''${SEABIOS_BOOT_DEVICE}''' != ''
```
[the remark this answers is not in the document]
SergiiDmytruk: `'''` is just to make error slightly better if the variable will contain a line break or a single quote by mistake.

trenchboot/02-with-drtm.robot
```diff
@@ -86,19 +86,56 @@ WTD005.001 DRTM event log has DRTM entries (AMD)
...
 
 WTD006.001 DRTM log aligns with PCR values
     [Documentation]    Verify that all DRTM measurements are correctly reflected
     ...    in DRTM event log
-    Validate PCRs Against Event Log    /sys/kernel/security/slaunch/eventlog
+    Execute Command In Terminal
+    ...    tpm2_eventlog /sys/kernel/security/slaunch/eventlog > /tmp/event-log
+    # This extracts hashes for a combination of a PCR (first parameter) and an
+    # algorithm (second parameter) skipping over events with empty data (these
+    # are informational markers which don't extend PCRs)
```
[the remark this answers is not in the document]
SergiiDmytruk: Oh, right.  Changed it to skip only digests of 0 bytes (no input to hash algorithm) that correspond to an event with no event data.  Looks like empty Linux parameters can be measured.

https://github.com/Dasharo/open-source-firmware-validation/compare/48b501db8c5c5acb6d5a1b6fa34c1a250a4f91d7..53a6511b287eb5f5ed948681a90bb62ee0b4b286#diff-7e5ab2b2ca7ecfb07ecd58004916bea6a39f406b2ccb9c3cf7a9ffdd90218011R119-R139

trenchboot/02-with-drtm.robot
```diff
@@ -86,19 +86,56 @@ WTD005.001 DRTM event log has DRTM entries (AMD)
...
+    # ...    find_event 17 "Measured MB2 module"
+    # Should Be Equal As Strings    ${match}    MATCH\nMATCH    No "Measured MB2 module" entries
 
 WTD006.001 DRTM log aligns with PCR values
     [Documentation]    Verify that all DRTM measurements are correctly reflected
     ...    in DRTM event log
-    Validate PCRs Against Event Log    /sys/kernel/security/slaunch/eventlog
+    Execute Command In Terminal
+    ...    tpm2_eventlog /sys/kernel/security/slaunch/eventlog > /tmp/event-log
+    # This extracts hashes for a combination of a PCR (first parameter) and an
```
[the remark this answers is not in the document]
SergiiDmytruk: https://github.com/Dasharo/open-source-firmware-validation/compare/48b501db8c5c5acb6d5a1b6fa34c1a250a4f91d7..53a6511b287eb5f5ed948681a90bb62ee0b4b286#diff-7e5ab2b2ca7ecfb07ecd58004916bea6a39f406b2ccb9c3cf7a9ffdd90218011R111-R114

lib/bios/menus.robot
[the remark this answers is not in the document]
krystian-hebel: In that case, it must be added to `platform-configs/include/default.robot`, otherwise all other tests would break.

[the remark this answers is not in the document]
SergiiDmytruk: https://github.com/Dasharo/open-source-firmware-validation/compare/53a6511b287eb5f5ed948681a90bb62ee0b4b286..5ba442bef68e4001c78a9033c01b249625761e3c#diff-07c15f12435114d3af82c6b0f99da527081f7f6bb6455aee02cbdf868630f838

trenchboot/02-with-drtm.robot
```diff
@@ -86,19 +86,70 @@ WTD005.001 DRTM event log has DRTM entries (AMD)
...
+    # algorithm (second parameter).  The printed line have the form
+    #    {event data length}-{digest}
+    # to make it possible to ignore some events based on a combination of these
+    # fields.
+    #
+    # Backslash in front of some dollar signs is escaping AWK fields while
```
macpijan, line 111: This seems very convoluted.

Perhaps we shall we include such helper shell scripts in this repo, send them to DUT and just execute in the test suite?

krystian-hebel: Then we would have to have SSH and know the device IP. RF really isn't made for such tests, that's why I didn't fully understand the decision to remove such tests from cukinia.

macpijan: > Then we would have to have SSH and know the device IP.

I think we do have it. And we do not need to know DUT IP, this should be automatically retrieved by the keywords if we have shell over serial. I think we can also send a file over serial, if no SSH is available as well.

> RF really isn't made for such tests, that's why I didn't fully understand the decision to remove such tests from cukinia.

If you're referring to: https://github.com/3mdeb/meta-trenchboot/pull/33#discussion_r1666680491 then I proposed some `rule of thumb` that made sense to me at the time. If there were strong opinions on other proposal, I may have missed them. We need to make decisions to move forward. If I made a bad one, I will be happily proven wrong and change it if it will make our lives better.

Ultimately, we want to execute some commands in RF as test executor, as we have control over hardware there. Unless you have other suggestions here as well.

krystian-hebel: Would it make sense to have such scripts as part of meta-trenchboot? It could potentially make it harder to run the tests in final environment (i.e. kernel of some widely used distro), but we don't plan on running tests there in the foreseeable future. Unfortunately, it could also mean that modifications to tests would require changes in two separate repositories.

SergiiDmytruk: > I can feel the pain while wrapping bash functions inside this RF code.

Every time I deal with RF, it reminds me of that [worst programming language ever talk](https://www.youtube.com/watch?v=vcFBwt1nu2U) (although I watched an older version).  And because *everything* about RF is pain, this doesn't really standout.  Avoiding any reboots of the DUT on subsequent runs also helps making such things work.

Adding local variables for parameters might help in case of this particular function, because there will be no mixing of `$1` from both bash and awk.  Another possibility is to move this to Python and use its more adequate raw string literals, possibly with some format strings.  Although, just like with separate script files, this will mean that different parts of the test will be in different places.

macpijan: If you guys have suggestions on how to make our lives less miserable, I am all ears. We can create a ticket for discussion if needed.

macpijan: We have concluded with @krystian-hebel that we leave it as is since it works. 

Place to discuss it in the future: https://github.com/Dasharo/open-source-firmware-validation/issues/573

macpijan:
@krystian-hebel let me know if you have more comments here

krystian-hebel (approved):
I haven't been able to test it on HW (all apus were checked out, except for apu3 where RTE didn't start it's REST server for rte_ctrl, I've notified Dawid an he'll look into it) but running the steps manually worked. CI passed for QEMU so I'm assuming default `SEABIOS_BOOT_DEVICE` also works as it should, I was afraid it may need `${EMPTY}` or some other RF magic.

Let's make pre-commit happy before merging. There is one typo and apparently something else, because it reports 2 modified files. I hope it doesn't try to un-escape what needs to be escaped...

macpijan:
>  I'm assuming default SEABIOS_BOOT_DEVICE also works as it should, I was afraid it may need ${EMPTY} or some other RF magic.

This is what pre-commit suggests

macpijan (approved):
Feel free to merge once you confirm it works well paired with: https://github.com/3mdeb/meta-trenchboot/pull/45

PR #559, opened 2024-11-03
dasharo-compatibility/dmidecode.robot: don't reboot for each test

krystian-hebel:
All of the test cases get some line from output of dmidecode and compare it against the value from platform config. There's no need to do a clean boot between them. Move the common parts of all test cases to the suite setup. In case of MSI Z790-P with two tests skipped, this reduces total suite time by a factor of 8.

Before:
image

After:
image


macpijan:
I like it, thanks a lot @krystian-hebel 

Tested on MSI Z690.

macpijan (approved):

Issue #250, opened 2024-03-28
VBO009.001 fails on apu


mkopec:
### Device

PC Engines apu4

### RTE version

v1.1.0

### Affected component(s) or functionality

Vboot test suite

### Brief summary

VBO009.001 fails on apu. Boots into the wrong OS and times out waiting for the Ubuntu shell prompt.

### How reproducible

100%

### How to reproduce

- Run Vboot test suite
- Check the results

### Expected behavior

- All tests pass

### Actual behavior

VBO009.001 fails

### Link to screenshots or logs

[log.zip](https://github.com/Dasharo/open-source-firmware-validation/files/14787781/log.zip)

macpijan:
@mkopec It passed manually? Do we know what was the problem?

macpijan:
@BeataZdunczyk @JanPrusinowski Another vboot-related issue. Added label to group them. Would be nice to extend the check/solve them together while we are here?

JanPrusinowski:
There were some issues with DUT.
Ubuntu was not installed.
After installing OS there were some issues with USB which caused unwanted logs to apear in the console.
DUT's IP must have changed as it turned out that it was different than in our config files.
There were issues with keywords which cloned repos from git.
There was an issue where `Make Sure That Flash Locks Are Disabled` kwd was missing... I have copied it from dcu.robot ad DUT does not use DCU.
Also there was a problem with flashing on the machine... Chip information was missing from the config.

All of the above changes/fixes tests could finally be run. However the original issue did not reproduce on the branch forked from the actual develop.

macpijan:
@JanPrusinowski This is not the proper approach here.

I have tracked it down a bit further in git history.

@filipleple Can you help me understand / fix these two commits:

1. https://github.com/Dasharo/open-source-firmware-validation/commit/39d65c703c86988be30f706f99051bd9f49a901b#diff-8d01f6edc2ae98aada8d51a7adb6dc18a7b43ab692f74e7dd01350cda86ce6a2R130
- you split the `Make Sure That Flash Locks Are Disabled` keywords into two options lib
- you keep the same documentation string for both
- but the functionality is entirely different in each of them - one performs check in Linux, the other one makes changes in the firmware, and no checks in Linux; perhaps this is intentional as changes with DCU cannot be made when locks are enabled, but I think this needs better explanation

2. https://github.com/Dasharo/open-source-firmware-validation/commit/112a434c9d9e26443439ad93c4f15258762b3606
[...]

PR #566, opened 2024-11-05
renamed msi-pro-z790-p.robot to msi-pro-z790-p-ddr5.robot: to reflect

JanPrusinowski:


lib/options/uefi-setup-menu.robot
```diff
@@ -143,3 +143,20 @@ Measure Reboot Time
...
+    Log To Console    Make Sure That Flash Locks Are Disabled
+    Power On
+    Boot System Or From Connected Disk    ubuntu
+    Login To Linux
+    Switch To Root User
+    Get Flashrom From Cloud
+    ${out_flashrom}=    Execute Command In Terminal    flashrom -p internal
+    # Should Not Contain    ${out_flashrom}    read-only
+
+Login To Windows
```
macpijan, line 160: I am unsure how it helps. 

At best it triggers conflict:

```
Multiple keywords with name 'Login To Windows' found. Give the full name of the keyword you want to use:
    keywords.Login To Windows
    uefi-setup-menu.Login To Windows
```

lib/options/uefi-setup-menu.robot
```diff
@@ -143,3 +143,20 @@ Measure Reboot Time
     ${min}    ${max}    ${average}    ${stddev}=
     ...    Calculate Boot Time Statistics    ${durations}
     RETURN    ${min}    ${max}    ${average}    ${stddev}
+
+Make Sure That Flash Locks Are Disabled
```
macpijan, line 147: @JanPrusinowski https://github.com/Dasharo/open-source-firmware-validation/issues/250#issuecomment-2433043835

platform-configs/include/default.robot
```diff
@@ -14,7 +14,7 @@ ${RTE_S2_N_PORT}=                                   13541
 ${FLASH_SIZE}=                                      ${TBD}
 ${FLASH_LENGTH}=                                    ${TBD}
 ${TIANOCORE_STRING}=                                to boot directly
-${BOOT_MENU_KEY}=                                   F11
+${BOOT_MENU_KEY}=                                   ${F11}
```
krystian-hebel, line 17: Same kind of change would have to be done in `novacustom-common.robot`, might as well change it now as it is the last place that needs to be fixed.

scripts/capsules/capsule_update_tests.sh
```diff
@@ -151,3 +151,57 @@ BaseTools/BinWrappers/PosixLike/GenerateCapsule --encode \
...
                                                 --output $capsule_name$file_descriptor.cap
+
+
+
+
+
+
+
+
+echo "--- CREATING CAPSULE WITH SERIAL REDIRECTION ---"
```
krystian-hebel, line 162: Why is it required? System settings are supposed to be preserved, if they aren't it is an issue with the capsule update mechanism, which should result in test failure.

Capsule created by this fragment isn't used anywhere, please remove changes to this file.

krystian-hebel (changes requested):

lib/options/uefi-setup-menu.robot
JanPrusinowski: Removed in: f518bda2fe7846ef332062f00ef98b830ad5698a

platform-configs/include/default.robot
JanPrusinowski: added in: a04c99c94b4d73b9b8034667e5be29ce077ee958

scripts/capsules/capsule_update_tests.sh
JanPrusinowski: You are right. I have added it by mistake - its just my script to edit the capsule so that it contains a serial redirection 'fix' which I use just for myself. Removed in: 36dca57910b698fa7cb95dfe573d9259dea31e2f

macpijan:
Picked relevant changes to: https://github.com/Dasharo/open-source-firmware-validation/pull/588

I am not convinced by this commit right now why it was neede: https://github.com/Dasharo/open-source-firmware-validation/pull/566/commits/9f507d7d776e6009ab66a14b0b5fd9b6984887b0

Skipping it for now.

krystian-hebel:
> I am not convinced by this commit right now why it was neede: [9f507d7](https://github.com/Dasharo/open-source-firmware-validation/commit/9f507d7d776e6009ab66a14b0b5fd9b6984887b0)

It was an attempt (unsuccessful AFAIK) to fix the problem where `SSHLibrary.Read` was invoked immediately after `SSHLibrary.Write`, before the actual output of the command was produced. After taking a second look, it seems that in case of PiKVM, it is writing through KVM, but reading from telnet. There are also keywords that change connection method from KVM to SSH and back. Perhaps there was old output in one of the buffers between such switch.

Issue #364, opened 2024-07-09
ME Disabled (HAP) will no longer affect lspci output


filipleple:
### Device

Novacustom v5x

### RTE version

--

### OSFV version

9603436e11cf52ab699fe4e928cdb6acb106198a

### Affected component(s) or functionality

[ME-neuter test](https://github.com/Dasharo/open-source-firmware-validation/blob/develop/dasharo-security/me-neuter.robot#L66)

### Brief summary

ME Disabled (HAP) will no longer affect lspci output, so we need to verify it in some other way.

### How reproducible

100%

### How to reproduce

Run the [MNE004.001 Intel ME mode option Disabled (HAP) works correctly (Ubuntu)](https://github.com/Dasharo/open-source-firmware-validation/blob/develop/dasharo-security/me-neuter.robot#L66) case

### Expected behavior

It should pass

### Actual behavior

It will always fail, since the HAP disabling doesn't affect `lspci` output anymore - ME will still appear as 00:16.0

### Link to screenshots or logs

--

macpijan:
We have a change in progress here: https://github.com/Dasharo/open-source-firmware-validation/pull/346
But it seems it's been applied to static test, can we apply this to dynamic checks as well @wiktormowinski ?
Perhaps we can have a keyword to check ME state, and reuse it across these tests.

wiktormowinski:
I just did the same in 3 test cases from me-neuter. By ''the same'' i mean the no-keyword version ... for now of course.
Same place:
https://github.com/Dasharo/open-source-firmware-validation/pull/346

BeataZdunczyk:
still working on fix here: https://github.com/Dasharo/open-source-firmware-validation/pull/375

mkopec:
If ME is enabled, we can check Operation Mode field in HFSTS1. When the HAP bit is set, the ME should go into Debug Mode, indicated by the Current Operation Mode field set to 2.

Or check CBMEM logs:

```
[DEBUG]  ME: HFSTS1                      : 0x80022054
[DEBUG]  ME: Current Operation Mode      : 2
[INFO ]  HECI: coreboot in recovery mode; found CSE in expected Debug Mode state, skipping EOP
```

wiktormowinski:
Newest version working on the new Develop branch https://github.com/Dasharo/open-source-firmware-validation/pull/403

JanPrusinowski:
[...]

PR #579, opened 2024-11-12
Disabled me fix

wiktormowinski:
finding solution for https://github.com/Dasharo/open-source-firmware-validation/issues/364


JanPrusinowski (approved):

krystian-hebel:
I think using `Check ME Out` instead of `Check ME State` in all test cases would be better. Right now only `MNE004.001` uses the latter directly.

wiktormowinski:
It must have slipped through. Here's the fix: https://github.com/Dasharo/open-source-firmware-validation/pull/581/commits/1e4edbedb2c59fa5978c70fab85ca470d511dba2

Issue #536, opened 2024-10-16
MNE - Me-Neuter tests fail on MSI


JanPrusinowski:
### Device

msi-pro-z690-a-wifi-ddr4

### RTE version

1.1.4

### OSFV version

develop

### Affected component(s) or functionality

me-neuter tests

### Brief summary

Set UEFI Option is unable to set Intel ME mode

### How reproducible

100%

### How to reproduce

`robot -v snipeit:no -L TRACE -v rte_ip:192.168.10.199 -v config:msi-pro-z690-a-wifi-ddr4 -v sonoff_ip:192.168.10.169 -v pikvm_ip:192.168.10.16 -v device_ip:192.168.10.155 -v fw_file:./msi_ms7d25_v1.1.4_ddr4.rom dasharo-security/me-neuter.robot`

### Expected behavior

Intel ME mode is set to correct value

### Actual behavior

Test fails on trying to set the option.

### Link to screenshots or logs

![image](https://github.com/user-attachments/assets/bf4e11e2-4bb5-4d1c-9f20-743a428a37bb)
![image](https://github.com/user-attachments/assets/f3c43a60-9565-48e2-b8eb-90f5768ba793)
[log2.zip](https://github.com/user-attachments/files/17392485/log2.zip)

macpijan:
It appears that help text is being parsed, instead of the menu text.

I'd start with confirming whether it is reproducible, or if it was some one-time problem.

macpijan:
The reason of this failure is addition of these two help text lines in the selectable options area:

![image](https://github.com/user-attachments/assets/05f3ddcc-3005-439a-b786-4f3319bf3830)

@BeataZdunczyk Can we try to ensure that we run some test working on certain menu when adding changes to such menus?

macpijan:
@BeataZdunczyk @JanPrusinowski So the solution would be to ignore the lines marked in yellow when parsing.

We've had something like this that checked if line was on option we could select, and dropped otherwise: https://github.com/Dasharo/open-source-firmware-validation/blob/develop/lib/bios/menus.py#L7

My suggestion would be:
- analyze the call trace - is this keyword used here?
- if no - why not, can we use it here?
- if yes - why it does not work? can we come up with better regexp?

JanPrusinowski:
[...]

PR #580, opened 2024-11-12
menus.robot: Added check if the option is not part of a description

JanPrusinowski:


wiktormowinski:
tested on `msi-pro-z690-a-ddr5` and looks good 
[me-neuter.robot_log.zip](https://github.com/user-attachments/files/17717327/me-neuter.robot_log.zip)

krystian-hebel:
Which part of the test failed before? Are there any logs with the failure I could look at?

lib/bios/menus.robot
```diff
@@ -226,9 +226,13 @@ Get Index Of Matching Option In Menu
     [Documentation]    This keyword returns the index of element that matches
     ...    one in given menu
     [Arguments]    ${menu_construction}    ${option}    ${ignore_not_found_error}=${FALSE}
+    Set Suite Variable    ${MATCHES}    ${FALSE}
```
krystian-hebel, line 229: No need for suite variable, please make it local (unless removed, see next comment).

lib/bios/menus.robot
```diff
@@ -226,9 +226,13 @@ Get Index Of Matching Option In Menu
...
     FOR    ${element}    IN    @{menu_construction}
-        ${matches}=    Run Keyword And Return Status
-        ...    Should Match    ${element}    *${option}*
+        # Ignore if ${option} is in quotes
+        ${is_in_quotes}=    Run Keyword And Return Status    Should Match    ${element}    *\'${option}\'*
```
krystian-hebel, line 232: This solution is valid for a very specific case only. It fixes just one of the symptoms, not the underlying problem.

Have you tried to use `opt_only` argument set to `TRUE` for call to `Enter Submenu From Snapshot And Return Construction` in `Set UEFI Option` (and, if it works, `Get UEFI Option`)? If I understand that code correctly, it should skip the informative lines from menu construction.

If that works, we may even consider changing the default in`Enter Submenu From Snapshot And Return Construction`, assuming it won't break anything else.

krystian-hebel (changes requested):

lib/bios/menus.robot
philipandag: https://github.com/Dasharo/open-source-firmware-validation/pull/580/commits/b72630e8dadf673f020969f8803813b114d33c37#diff-6ad7c5fb97cb1099c7ecd42f15dde3caf06d38b9fe5244afb96394454b6d6bfaL229-R229

lib/bios/menus.robot
philipandag: Because
> I have tried solutions you suggested - but they did not change anything.
I have added a more precise regex for detecting lines containing options in `Get Index Of Matching Option In Menu` instead of changing `Enter Submenu From Snapshot And Return Construction`
https://github.com/Dasharo/open-source-firmware-validation/pull/580/commits/b72630e8dadf673f020969f8803813b114d33c37#diff-6ad7c5fb97cb1099c7ecd42f15dde3caf06d38b9fe5244afb96394454b6d6bfaR232-R235

philipandag: I will check if using the `Check If Menu Line Is An Option` instead wouldn't be better

philipandag: > Have you tried to use opt_only argument set to TRUE

I have tried this suggestion and it seems it works perfectly on a z790-ddr5. 
[me-neuter.robot_log.zip](https://github.com/user-attachments/files/17991477/me-neuter.robot_log.zip)

There is an issue with the me-neuter test suite, where it assumes that the ME lspci entry exists after disabling it using the HAP bit. It *sometimes* does, for example on protectli VP2420 if I recall correctly, but it certainly is not always the case. Created an issue for that
https://github.com/Dasharo/open-source-firmware-validation/issues/621

miczyg1:
There are 4 commits thad add just 2 lines with TRUE value. It looks like it needs a cleanup. @philipandag 

Is this PR still valid and fixes something?

philipanda:
> There are 4 commits thad add just 2 lines with TRUE value. It looks like it needs a cleanup. @philipandag
> 
> Is this PR still valid and fixes something?

@miczyg1 
One of the commits was a revert of all the previous ones. I have removed them instead.
This PR fixes https://github.com/Dasharo/open-source-firmware-validation/issues/536

miczyg1 (approved):

PR #596, opened 2024-11-18
591 duplicate keywords

EduKav1813:
I solved the problem of the duplicate keywords in the auto-generated documentation by allowing users to choose which version to generate: DCU or UEFI. After the fix the amout of duplicate keywords is 0


m-iwanicki:
@EduKav1813 you should probably rebase instead of merge especially if PR has only couple of commits as currently you introduced another commit. If you merge this PR then you will have yet another merge commit

README.md
```diff
@@ -14,17 +14,24 @@ appear here.
...
+   7. [Getting started](#getting-started)
+      1. [Initializing environment](#initializing-environment)
+      2. [Running tests](#running-tests)
+      3. [Running tests via wrapper](#running-tests-via-wrapper)
+      4. [Running tests with additional arguments](#running-tests-with-additional-arguments)
+      5. [Running regression tests](#running-regression-tests)
+      6. [Running regression tests with additional arguments](#running-regression-tests-with-additional-arguments)
+   8. [Useful refactoring tools](#useful-refactoring-tools)
+   9. [Generating documentation](#generating-documentation)
+   10. [Additional documents](#additional-documents)
```
m-iwanicki, line 34: Why those changes?

README.md
```diff
@@ -318,12 +325,28 @@ $ firefox keywords.html
...
+It also allows to choose the version of the
+documentation, which basically decides if you want to perform the UEFI
+configuration via "UEFI setup app" (`UEFI`) or via "Dasharo Config
+Utility tool" (`DCU`):
+
+```bash
+$ ./scripts/create-docs.sh UEFI
+Generating UEFI version of the documentation
+Documentation generated and saved as all-keywords.html
+$ firefox all-keywords.html
```
m-iwanicki, line 340:
```suggestion
Generating UEFI version of the documentation
[DEBUG] Skip dcu.robot
Documentation generated and saved as ./docs/index.html
$ firefox docs/index.html
```

At least that's output I'm getting.

README.md
```diff
@@ -318,12 +325,28 @@ $ firefox keywords.html
...
+
+Or
 
 ```bash
-$ ./scripts/create-docs.sh
-Documentation generated and saved as docs/index.html
+$ ./scripts/create-docs.sh DCU
+Generating DCU version of the documentation
+Documentation generated and saved as all-keywords.html
+$ firefox all-keywords.html
```
m-iwanicki, line 349:
```suggestion
Generating DCU version of the documentation
[DEBUG] Skip uefi-setup-menu.robot
Documentation generated and saved as ./docs/index.html
$ firefox docs/index.html
```

m-iwanicki (changes requested):

scripts/create-docs.sh
```diff
@@ -1,9 +1,30 @@
...
+    echo "Generating DCU version of the documentation"
+else
+    echo "Please provide version of the documentation: (UEFI/DCU)"
+    echo "Exit"
+    exit 0
+fi
 
 cat keywords.robot > "$TEMP_DIR/all-keywords.robot"
-find lib -name "*.robot" -exec cat {} + >> "$TEMP_DIR/all-keywords.robot"
+find lib -name "*.robot" -type f | while read -r file; do
```
philipandag, line 17: Maybe instead of seperating the keywords into two seperate documentations they could be shown together by prefixing the library name to the keyword name, just like the robot framework does when there are ambiguous keyword definitions. 

That would require recognizing the keyword names in their definitions in the contents of the file. With that appending the basename of the library file to every one of them in the library file would be easy. I think some code that could help with parsing the files should exist in the robot framework python package.

If not then maybe the keywords could be extracted and replaced using a regex like this? `^[A-Za-z0-9_\ ]+$` The keywords should always start from the first column, should and no other text should be after it. The keywords also only appear after the `*** Keywords ***` line. To make this method work only the contents after the `*** Keywords ***` line should be appended to the `all-keywords` file, assuming not including the `*** Settings ***` and other sections won't break libdoc.

I believe that if doing it in a more generic way, like adding the prefixes, would be more future-proof and won't require maintaining multiple documentations.

README.md
EduKav1813: Accidental formatter change. Reverted in cabe22eb4e74b7652d80eee50286961dc1c79a92

scripts/create-docs.py
```diff
@@ -0,0 +1,42 @@
...
+# Collect all .robot files
+robot_files = ["keywords.robot"]
+for root, dirs, files in os.walk(lib_dir):
+    for file in files:
+        if file.endswith(".robot"):
+            robot_files.append(os.path.join(root, file))
+
+# Regex pattern to detect keyword definitions
+keyword_pattern = re.compile(
+    r"\n(?:(?!Library|Resource|Variables|Documentation)[A-Z][a-z0-9_]+[\ ]+.*)"
```
m-iwanicki, line 17:
```suggestion
    r"^(?:(?!Library|Resource|Variables|Documentation)[A-Z][a-z0-9_]+[\ ]+.*)",
    r.MULTILINE
```

scripts/create-docs.py
```diff
@@ -0,0 +1,42 @@
...
+)
+
+delimiter = "."  # Between filename and the keyword
+
+# Create the combined resource file with prefixed keywords
+with open(output_resource, "w", encoding="utf8") as res_file:
+
+    for robot_file in robot_files:
+        file_prefix = os.path.splitext(os.path.basename(robot_file))[0] + delimiter
+        index_offset = 1  # Initial value = length of \n
```
m-iwanicki, line 27:
```suggestion
        index_offset = 0
```

if you use previous suggestion (use `^` instead of `\n`)

scripts/create-docs.py
```diff
@@ -0,0 +1,42 @@
...
+            text = file.read()
+
+            for m in re.finditer(keyword_pattern, text):
+                start = m.start(0)
+                text = (
+                    text[: start + index_offset]
+                    + file_prefix
+                    + text[start + index_offset :]
+                )
+                index_offset += len(file_prefix)
```
m-iwanicki, line 39: Not sure if modifying text you are looping through is such a good idea if there are easier ways.
Maybe you could read each file line by line and check if line matches pattern? Working example

```py
import os
import re
import sys

lib_dir = "lib"
output_resource = sys.argv[1] if len(sys.argv) > 1 else "all-keywords.robot"

# Collect all .robot files
robot_files = ["keywords.robot"]
for root, dirs, files in os.walk(lib_dir):
    for file in files:
        if file.endswith(".robot"):
            robot_files.append(os.path.join(root, file))

# Regex pattern to detect keyword definitions
keyword_pattern = re.compile(
    r"^(?:(?!Library|Resource|Variables|Documentation)[A-Z][a-z0-9_]+[\ ]+.*)",
    re.MULTILINE
)

delimiter = "."  # Between filename and the keyword

with open(output_resource, "w", encoding="utf8") as write_file:
    for robot_file in robot_files:
        file_prefix = os.path.splitext(os.path.basename(robot_file))[0] + delimiter
        index_offset = 0

        with open(robot_file, "r", encoding="utf8") as read_file:
            for line in read_file:
                if re.match(keyword_pattern, line):
                    write_file.write(file_prefix + line)
                else:
                    write_file.write(line)
```

m-iwanicki: I think it's a little easier to modify that way (unless we need multi-line processing)

m-iwanicki (changes requested):

scripts/create-docs.py
m-iwanicki: If you implement reading line by line then you don't likely don't need `re.MULTILINE` flag in keyword_patter

scripts/create-docs.py
```diff
@@ -0,0 +1,42 @@
+import os
```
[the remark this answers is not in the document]
EduKav1813: Fixed in 6ccadecdb5207ec2e3208ce21f47551a14474889

scripts/create-docs.py
EduKav1813: Fixed in 2774df89044aa7b0999e4f32309836c74e45a346

scripts/create-docs.py
EduKav1813: Fixed in 2774df89044aa7b0999e4f32309836c74e45a346

scripts/create-docs.py
EduKav1813: Fixed in 5830fdb4f34817ec39ebdf1d340b5d3f2f1de161

EduKav1813: Now redundant multiline argument also was removed in 515779bb9ff69645ff471f72796e0dca3b3062da

m-iwanicki (approved):
Looks good, maybe squash some commits before merge

m-iwanicki:
@EduKav1813 your last batch of commits is unsigned

EduKav1813:
> @EduKav1813 your last batch of commits is unsigned

Thanks, I fixed that and now my commits are properly signed

EduKav1813:
> Looks good, maybe squash some commits before merge

I prefer to squash the entire branch as one merge commit, would this be ok?

m-iwanicki:
@EduKav1813 yeah, it's ok.

README.md
m-iwanicki: Changed here: https://github.com/Dasharo/open-source-firmware-validation/pull/596/commits/718a2f4bf1f766e056609cf1db8d977c364f8b5a
@EduKav1813 you should've mentioned it here

scripts/create-docs.sh
m-iwanicki: @philipandag your request should be implemented in current version, can you check again and confirm if it's ok?

PR #597, opened 2024-11-18
requirements.txt: update dependencies so they work with Python 3.13

m-iwanicki:


m-iwanicki:
Doesn't fix 100% as telnetlib was removed in Python 3.13:

```
[ ERROR ] Error in file '/home/miwanicki/projects/open-source-firmware-validation/dasharo-compatibility/network-boot.robot' on line 6: Importing library 'robot.libraries.Telnet' failed: ModuleNotFoundError: No module named 'telnetlib'
Traceback (most recent call last):
```

macpijan:
> Doesn't fix 100% as telnetlib was removed in Python 3.13:

Which makes it imperfect proposal, as we need this lib for nearly all tests.

I am afraid we need to stick with this workaround for now instead: https://github.com/Dasharo/open-source-firmware-validation/issues/598#issuecomment-2493624029

PR #601, opened 2024-11-19
540 reset to defaults restore serial

philipandag:


lib/bios/menus.robot
```diff
@@ -410,11 +410,21 @@ Reset To Defaults Tianocore
...
     ...    setting menu.
+    Power On
+    ${main_menu}=    Enter Setup Menu Tianocore And Return Construction
     Read From Terminal
     Press Key N Times    1    ${F9}
     Read From Terminal Until    ignore.
     Write Bare Into Terminal    y
 
+    ${dasharo_menu}=    Enter Dasharo System Features    ${main_menu}
+    ${serial_menu}=    Enter Dasharo Submenu    ${dasharo_menu}    Serial Port Configuration
```
krystian-hebel, line 421: I expect this to fail on platforms that don't have serial port menu. It would be best to run these steps only if `DASHARO_SERIAL_PORT_MENU_SUPPORT` is set.

philipandag: https://github.com/Dasharo/open-source-firmware-validation/pull/601/commits/df81f204ccb46eec1845c46fef3e36e99722ebb7

At this moment every platform supports it. It's defined as ${TRUE} in the `default.robot` and never overwritten. But that's a good point. Thanks.

lib/bios/menus.robot
```diff
@@ -410,11 +410,23 @@ Reset To Defaults Tianocore
...
     Read From Terminal
     Press Key N Times    1    ${F9}
     Read From Terminal Until    ignore.
     Write Bare Into Terminal    y
 
+    IF    ${DASHARO_SERIAL_PORT_MENU_SUPPORT}
+        ${dasharo_menu}=    Enter Dasharo System Features    ${main_menu}
+        ${serial_menu}=    Enter Dasharo Submenu    ${dasharo_menu}    Serial Port Configuration
+        TRY
+            Set Option State    ${serial_menu}    Enable COM0 Serial    ${TRUE}
```
macpijan, line 424: Why is it needed? Can we have different naming here on different platforms? This deserves a comment at least.

platform-configs/msi-pro-z690-a-ddr5.robot
```diff
@@ -13,3 +13,18 @@ ${DEVICE_AUDIO2}=                   Raptorlake HDMI
 
 ${CPU_MAX_FREQUENCY}=               5200
 ${CPU_MIN_FREQUENCY}=               300
+
+
+${CPU_TESTS_SUPPORT}=                           ${TRUE}
```
macpijan, line 18: Probably pre-commit complains on lack of alignment.

macpijan:
@philipandag please fix pre-commit prior review request

lib/bios/menus.robot
```diff
@@ -410,11 +410,23 @@ Reset To Defaults Tianocore
     [Documentation]    Resets all Tianocore options to defaults. It is invoked
     ...    by pressing F9 and confirming with 'y' when in option
     ...    setting menu.
+    Power On
```
macpijan, line 413: This keyword is always used when we are already in the UEFI setup, right?

Do we have to use hard power cycle, can we instead issue reset?

platform-configs/msi-pro-z690-a-ddr5.robot
macpijan: Also, does it belong to this MR?

If so, have we checked the other MSIs have these flags set as well?

platform-configs/msi-pro-z690-a-ddr5.robot
```diff
@@ -13,3 +13,18 @@ ${DEVICE_AUDIO2}=                   Raptorlake HDMI
...
+${INTEL_HYBRID_ARCH_SUPPORT}=                   ${TRUE}
+${HYPER_THREADING_SUPPORT}=                     ${TRUE}
```
macpijan, line 20: These 3 flags might go to msi-common? Only the CPU count number are per-unit specific?

philipandag:
> @philipandag please fix pre-commit prior review request
> 
> Do you have it installed locally in repo?

This time I had it installed. There's a different behaviour of the pre-commit in this repo right now:

`/code/lib/bios/menus.robot:582:1 [W] 0501 Keyword 'Boot System Or From Connected Disk' is too long (52/50) (too-long-keyword)`

macpijan:
Very weird... let's try to rebase then.

platform-configs/msi-pro-z690-a-ddr5.robot
philipandag: They do not. I figured it should be added as I needed to add these flags to check if the tests work now, and it was only performed on this single platform. Although you are probably right that he flags  should be either added everywhere, or nowhere...

lib/bios/menus.robot
philipandag: https://github.com/Dasharo/open-source-firmware-validation/pull/601/commits/4364e09f38fab8ea9bca2f303d4b42a33af0280b

platform-configs/msi-pro-z690-a-ddr5.robot
philipandag: >Probably pre-commit complains on lack of alignment.

That also was a problem, although it was not listed explicitly like the issue with the "Boot System..." keyword, so I haven't noticed that, and so turns out I haven't had pre-commit installed.

@macpijan 
I think I have solved this issue for good by making pre-commit install itself automatically in every repo that uses it using git templates:
```
$ pre-commit init-templatedir ~/.git-template
$ git config --global init.templateDir ~/.git-template
```
The `~/.git-template` can be any directory.

Now when I clone any repository, the pre-commit hook is automatically installed thanks to the template.
Do you think we could add the instructions on how to setup this somewhere?

platform-configs/msi-pro-z690-a-ddr5.robot
philipandag: https://github.com/Dasharo/open-source-firmware-validation/compare/628944c..0f55d7ec3cb33c76e147a229949b4eb966f24b71

lib/bios/menus.robot
philipandag: I suppose yes. Changed here: https://github.com/Dasharo/open-source-firmware-validation/compare/4364e09f38fab8ea9bca2f303d4b42a33af0280b..a0485cccb7f9f1c6c2c70ff4763da0b2ad0ae52e
Will update with results of a quick test if it does indeed work as well.

philipandag:
Rebased onto develop

platform-configs/msi-pro-z690-a-ddr5.robot
macpijan: > Do you think we could add the instructions on how to setup this somewhere?

https://github.com/3mdeb/workstation-config ?

lib/bios/menus.robot
macpijan: > # The Serial Redirection option is called differently in some versions of Dasharo

This could be more specific - which platform version you have spotted this on.

This is q bit unfortunate workaround, as we do not fix the problem at it's roots, we only workaround one of the instances. If we use this menu elsewhere, we will hit this problem again.

> maybe the menus.py library could be modified a bit?

Definitely, we should fix (or apply a workaround) there. Do you have some ideas?

As a last restort, we can create issue for this, but ideally we at least apply workaround (with a proper comments) in the library, not in one of it's calls.

miczyg1: `Enable COM0 Serial Console Redirection`  <- we changed it per Protectli request at some point.

platform-configs/msi-pro-z690-a-ddr5.robot
philipandag: I also thought about this repository. I haven't worked with ansible before, so I will just issue an issue https://github.com/3mdeb/workstation-config/issues/53

lib/bios/menus.robot
philipandag: This library is actually only used in the `Set/Get UEFI Option` keywords and is not used in this test case. The way the Serial Redirection is set here is a standard among other tests, for example the `reset-to-defaults.robot`. The paths are hard-coded into the tests and if the path changes in some version of Dasharo, all the tests that use it would need to be updated to check or alternatives. I think this is a reasonable workaround for now until we want to do a rework of how the options are being found in all of the tests.

And about the possible rework:
The main question here is how much of variance in the paths is accepted in the future? Will it only be alternative names of single options? What about the names of submenus? Maybe whole paths will be changed, or the options will be moved between existing menus?

As for my ideas for the rework:
If the `menus.py` library was used everywhere menus are traversed then exceptions in names could be easy to fix. I'd imagine the library could just keep alternative paths to the options and the alternatives could be tried one by one.

So instead of a list of steps:
```
    "SerialRedirection": [
        "Dasharo System Features",
        "Serial Port Configuration",
        "Enable Serial Port Console Redirection",  # Enable COM0 Serial Console Redirection??
    ],
```
A tree could be used. 
```
    "SerialRedirection": [
        "Dasharo System Features": [
            "Serial Port Configuration": [
                "Enable Serial Port Console Redirection",
                "Enable COM0 Serial Console Redirection"
            ]
        ]
    ],
```
This would easily solve simple renames, but if the option would be moved between two existing menus then finding it could require traversing the menus backwards ...
```
     "A": [
        "B": [ # B exists, enter
            "Option" # No "Option" in this submenu
        ],
        "C": [ # Need to go up to try "C"
            "Option" # "Option" found
        ]
    ],
```

Also there's the option to define paths and menu option names in the menus as config variables, which would need creating and maintaining a whole lot of them. There are 15 paths, 7 submenus and 16 option names defined in menus.py. And there already are too many variables in the configs to remember. 
```
${SETUP_SUBMENU_DASHARO_SYSTEM_FEATURES}=    "Dasharo System Features"
${SETUP_SUBMENU_SERIAL_PORT_CONFIG}=    "Serial Port Configuration"
${SETUP_OPTION_SERIAL_PORT_REDIRECTION}=    "Enable COM0 Serial Console Redirection"
@{OPTION_PATH_SERIAL_PORT_REDIRECTION}=
...    ${SETUP_SUBMENU_DASHARO_SYSTEM_FEATURES}
...    ${SETUP_SUBMENU_SERIAL_PORT_CONFIG}
...    ${SETUP_OPTION_SERIAL_PORT_REDIRECTION}
```
This way though any future change in the menus structure would be easily fixable on per-platform basis and there would be no need to dynamically look around for where the options are located, which could take a lot of time depending on the amount of possible alternative paths.

macpijan: We definitely would like to keep the menu names in a single place. If Dasharo versioning would be standardized, we could define which menu layout corresponds to the given Dasharo version. Let's move your ideas to an issue so we can continue discussing there.

macpijan: @philipandag merging already, but please remembr  of this issue

macpijan (approved):

lib/bios/menus.robot
macpijan: Ok, there it is: https://github.com/Dasharo/open-source-firmware-validation/issues/614

PR #610, opened 2024-11-25
vp3230_rc4_fixes

SebastianCzapla:


lib/options/uefi-setup-menu.robot
```diff
@@ -86,7 +86,7 @@ Measure Coldboot Time
     Log To Console    \n
     # Do one more iteration than requested, as we may hit first boot which is always longer.
     FOR    ${index}    IN RANGE    0    ${iterations}+1
-        Power Cycle On    power_button=${TRUE}
+        Power Cycle On
```
macpijan, line 89: @SebastianCzapla Please see and rebase on top of it instead using similar approach: https://github.com/Dasharo/open-source-firmware-validation/pull/615

Adding this argument for one platform, and removing for another (without testing the results on different platforms at the same time) will not lead to anything good.

The updated `Power Cycle On` version from my PR should help here.

.gitignore
```diff
@@ -18,3 +18,4 @@ dl-cache/*
 !dl-cache/.gitkeep
 qemu-data/
 dcu/
+/logs
```
macpijan, line 21: it's looks weird to add absolute path to git ignore, doesn't it?

SebastianCzapla: Absolute path notation in gitignore is based on repository's root.
/logs ignores logs directory only within the root, while "logs" would ignore all logs directories, without concert at which depth of a repo it is.

"The pattern hello.* matches any file or directory whose name begins with hello.. If one wants to restrict this only to the directory and not in its subdirectories, one can prepend the pattern with a slash, i.e. /hello.*; the pattern now matches hello.txt, hello.c but not a/hello.java."

Taken from https://git-scm.com/docs/gitignore

macpijan: Thanks for the link. I would argue then `logs` may be better as we do not want to include any logs dirs. But this is fine too I guess.

PR #615, opened 2024-12-01
unify approach to Power On

macpijan:
* Use Power Supply keywords from osfv_cli libs
* Unify Power On keywords in include/default.robot in a for of "Power On Default" keyword, as most of them looked more or less identical
* Platfom configs can choose to use this keyword as Powe On, or define platform-specific one as before

Tested following platforms via the
BPS001.001 from util/basic-platform-setup.robot:
- protectli-vp2410
- protectli-vp2420
- protectli-vp4630
- protectli-vp4670
- protectli-vp6670
- pcengines-apu2
- msi-pro-z690-a-ddr5
- odroid-h4-plus
- minnowboard-turbot
- optiplex-7010

### DIFF
--- a/keywords.robot
+++ b/keywords.robot
@@ -702,45 +702,37 @@ Get DUT To Start State
     END
 
 Turn On Power Supply
-    ${pc}=    Get Variable Value    ${POWER_CTRL}
-    IF    'sonoff' == '${pc}'
-        ${state}=    Sonoff On
-    ELSE
-        ${state}=    Rte Relay Set    on
-    END
+    Rte Psu On
 
 Power Cycle On
-    [Documentation]    Clears telnet buffer and perform full power cycle with
-    ...    RTE relay set to ON.
-    [Arguments]    ${power_button}=${FALSE}
+    [Documentation]    Clears telnet buffer and performs full power cycle
+    ...    by setting power supply to OFF, and then to ON. If platform needs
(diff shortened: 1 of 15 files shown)


keywords.robot
```diff
@@ -702,45 +702,37 @@ Get DUT To Start State
...
-        Telnet.Read
-    EXCEPT
-        Log    Could not clear Telnet buffer
+        Rte Psu Off
+        Sleep    10
+        TRY
+            Telnet.Read
+        EXCEPT
+            Log    Could not clear Telnet buffer
+        END
```
miczyg1, line 732: if it is a generic keyword, is it actually good to use Telnet for every possible platform?

I know it was here before, but... Shouldn't it use `Read From Terminal` or whatever the generic keyword is called?

miczyg1:
Generally looks good. Also looked at changes on the osfv-scripts side.

Finally proper use of `DEFAULT_POWER_STATE_AFTER_FAIL` :) Just one thing that bothers me a little is this Telnet buffer clearing.

keywords.robot
macpijan: I think that was the assumption back then, yes.

This might be still valid, as telnet would be the only one we can use that early (we won't use ssh here).

Nevertheless, updated that, still works after this change on mentioned platforms.

PR #626, opened 2024-12-03
Add functionality for running tests on a remote machine

pietrushnic:
This commit introduces a new feature that allows users to run tests on a remote machine. This is particularly useful when running QEMU on the local machine is not desirable due to security concerns or memory limitations.

The changes include updates to the documentation to explain the new feature and a new script, `remote-runner.sh`, which handles the synchronization of a local directory to a remote machine and the execution of QEMU based on parameters on the remote machine.

### DIFF
--- a/lib/bios/common.robot
+++ b/lib/bios/common.robot
@@ -36,7 +36,7 @@ Enter Setup Menu
         Write Bare Into Terminal    ${SETUP_MENU_KEY}
         IF    '${BIOS_LIB}' == 'seabios'
             ${menu}=    Get Boot Menu Construction
-            Enter Menu From Snapshot    ${menu}    \[setup\]
+            Enter Submenu From Snapshot    ${menu}    \[setup\]
         END
     END
 
--- a/lib/bios/common.robot
+++ b/lib/bios/common.robot
@@ -167,3 +167,60 @@ Boot System Or From Connected Disk    # robocop: disable=too-long-keyword
         ${system_index}=    Get Index Of Matching Option In Menu    ${menu_construction}    ${system_name}
     END
     Select Boot Menu Option    ${system_index}    ${ARROW_DOWN}
+
+Make Sure That Network Boot Is Enabled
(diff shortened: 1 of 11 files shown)


lib/bios/common.robot
```diff
@@ -167,3 +167,60 @@ Boot System Or From Connected Disk    # robocop: disable=too-long-keyword
...
+    IF    '${BIOS_LIB}' == 'seabios'
+        Power On
+        Enable Network/PXE Boot
+    ELSE
+        Set UEFI Option    NetworkBoot    ${TRUE}
+    END
+
+Get IPXE Boot Menu Construction
+    [Documentation]    Keyword allows to get and return iPXE menu construction.
+    [Arguments]    ${lines_top}=1    ${lines_bot}=0    ${checkpoint}=${EDK2_IPXE_CHECKPOINT}
```
miczyg1:
```suggestion
    [Arguments]    ${lines_top}=1    ${lines_bot}=0    ${checkpoint}=${IPXE_CHECKPOINT}
```

Seems cosmetic, but may cause confusion when it is defined for SeaBIOS-based firmware. But it doesn't;t seem to be EDK2-specific, especially if we include the same iPXE built-in menu.

miczyg1: @pietrushnic ?

pietrushnic:
@miczyg1, why did you get back to this PR after so long? Is that helpful code somewhere?

miczyg1:
> @miczyg1, why did you get back to this PR after so long? Is that helpful code somewhere?

I had a JIRAfication day yesterday. Too many red tasks, tasks in verification to unblock someone, etc. piled up. The task related to this PR is in progress since the beginning and there is no perspective for it to be closed soon. If this has close to no value, then maybe the task should be closed? @pietrushnic

pietrushnic:
@miczyg1 no, it has value, at least for me. Please ping me in an internal task for that. So we can clarify WTH.

PR #635, opened 2024-12-17
CFC test suite remade

philipanda:


dasharo-compatibility/coreboot-fan-control.robot
```diff
@@ -32,7 +33,9 @@ CFN001.001 CPU temperature and fan speed can be read (Debian)
     Power On
     Boot From USB
     Serial Root Login Linux    debian
-    ${rpm}    ${temperature}=    Get CPU Temperature And CPU Fan Speed
+    ${rpm}=    Get Fan RPM
+    ${temperature}=    Get CPU Temperature CURRENT
+    # ${rpm}    ${temperature}=    Get CPU Temperature And CPU Fan Speed
```
miczyg1: Why leave commented code? Is it a WIP PR?

dasharo-compatibility/coreboot-fan-control.robot
```diff
@@ -49,20 +52,24 @@ CFN002.001 CPU fan speed increases if the temperature rises (Debian)
     # start of the test case while waiting for the temperature and fan speed
     # drop.
     FOR    ${iteration}    IN RANGE    0    ${COOLING_PROCEDURE_ITERATIONS}
-        ${rpm}    ${temperature}=    Get CPU Temperature And CPU Fan Speed
+        ${rpm}=    Get Fan RPM
+        ${temperature}=    Get CPU Temperature CURRENT
```
miczyg1: Why the keyword has CURRENT in capital letters? Why into just name the keyword `Get CPU Temperature`?

dasharo-compatibility/coreboot-fan-control.robot
```diff
@@ -49,20 +52,24 @@ CFN002.001 CPU fan speed increases if the temperature rises (Debian)
...
         IF    ${rpm}>=3000 or ${temperature}>=40
             Sleep    60s
         ELSE
             BREAK
         END
     END
-    ${rpm_1}    ${temperature_1}=    Get CPU Temperature And CPU Fan Speed
+    ${rpm_1}=    Get Fan RPM
+    ${temperature_1}=    Get CPU Temperature CURRENT
     Telnet.Execute Command    stress-ng --cpu 16 --io 8 --vm 4 --vm-bytes 4G --timeout 60s --metrics
```
miczyg1: The temperature should rise almost immediately. I would just start the stress-ng in the background for 30s, wait 25 seconds and then read the temperature. Whe the stress test is over, the temperature and fan speed may already be lower from what we want to test while we read the values from sensors.

```
stress-ng --cpu 16 --io 8 --vm 4 --vm-bytes 4G --timeout 30s --metrics &
Sleep 25s
Get Fan RPM
Get CPU Temperature CURRENT
```

After 5 seconds the stress-ng will finish in the background during the next 240s of sleep.

Also we must not use any specific library to execute commands, right? We have a wrappers for executing commands in OS that will use the needed library.

dasharo-compatibility/coreboot-fan-control.robot
```diff
@@ -49,20 +52,24 @@ CFN002.001 CPU fan speed increases if the temperature rises (Debian)
...
-    ${rpm_2}    ${temperature_2}=    Get CPU Temperature And CPU Fan Speed
+    ${rpm_2}=    Get Fan RPM
+    ${temperature_2}=    Get CPU Temperature CURRENT
     Sleep    240s
     # Due to the temperature increasing fan speed should rise.
-    ${rpm_3}    ${temperature_3}=    Get CPU Temperature And CPU Fan Speed
+    ${rpm_3}=    Get Fan RPM
+    ${temperature_3}=    Get CPU Temperature CURRENT
     IF    ${temperature_1}>=${temperature_2}
         FAIL    Temperature not increased
```
miczyg1:
```suggestion
        FAIL    Temperature did not increase
```

lib/sensors.robot
```diff
@@ -0,0 +1,121 @@
...
+
+    Import Variables    ${CURDIR}/../platform-configs/${SENSORS_CONFIG_FILE}
+    ${cpu_temperature_measurement_method}=    Get From Dictionary    ${CPU_TEMPERATURE_MEASUREMENT}    method
+    ${fan_pwm_measurement_method}=    Get From Dictionary    ${FAN_PWM_MEASUREMENT}    method
+    ${fan_rpm_measurement_method}=    Get From Dictionary    ${FAN_RPM_MEASUREMENT}    method
+
+    ${lm_sensors_used}=    Catenate
+    ...    '''${cpu_temperature_measurement_method}''' == '''system76-acpi''' or
+    ...    '''${cpu_temperature_measurement_method}''' == '''lm-sensors''' or
+    ...    '''${fan_rpm_measurement_method}''' != '''lm-sensors''' or
```
miczyg1: This looks weird. If we don't use lm-sensors for fan RPM measurement, then we use lm-sensors anyway?

lib/sensors.robot
```diff
@@ -0,0 +1,121 @@
...
+        ...    none. Either it should be changed or this test should not be
+        ...    performed on this platform.
+    ELSE IF    '''${fan_pwm_measurement_method}''' == '''hwmon'''
+        ${fan_pwm_measurement_hwmon_path}=    Get From Dictionary    ${FAN_PWM_MEASUREMENT}    hwmon_path
+        ${pwm}=    Execute Command In Terminal
+        ...    cat ${fan_pwm_measurement_hwmon_path}
+        ${pwm}=    Convert To Number    ${pwm}
+        RETURN    ${pwm}
+    ELSE
+        Fail    Wrong platform configuration. FAN_RPM_MEASUREMENT["method"] is
```
miczyg1: We check `FAN_PWM_MEASUREMENT` in this keyword, so why we complain on `FAN_RPM_MEASUREMENT["method"]` here?

lib/sensors.robot
```diff
@@ -0,0 +1,121 @@
...
+    [Documentation]    Get current CPU fan RPM
+    ${fan_rpm_measurement_method}=    Get From Dictionary    ${FAN_RPM_MEASUREMENT}    method
+    IF    '''${fan_rpm_measurement_method}''' == '''lm-sensors'''
+        ${fan_rpm_measurement_sensor}=    Get From Dictionary    ${FAN_RPM_MEASUREMENT}    lm_sensors_sensor_name
+        IF    '''${fan_rpm_measurement_sensor}''' != '''none'''
+            ${rpm}=    Execute Linux Command
+            ...    sensors ${fan_rpm_measurement_sensor} 2> /dev/null | grep -E 'fan1' | tr -s ' ' | cut -d ' ' -f2
+            ${rpm}=    Convert To Integer    ${rpm}
+            RETURN    ${rpm}
+        END
```
miczyg1: ELSE a Fail is missing here? If the `fan_rpm_measurement_sensor` is none then this function will return nothing. I don't think it is correct.

platform-configs/include/default.robot
```diff
@@ -252,6 +253,8 @@ ${TEMPERATURE_TEST_MEASURE_INTERVAL}=               1
 ${CUSTOM_FAN_CURVE_TEST_DURATION}=                  30
 # Interval between the following readings in custom fan curve tests
 ${CUSTOM_FAN_CURVE_MEASURE_INTERVAL}=               1
+# Module for a sensor used in CFC tests
+${KERNEL_MODULE_IT87_SUPPORT}=                      ${FALSE}
```
miczyg1: I don't see this variable being used anywhere

platform-configs/include/sensors/default-fan-curve-config.yaml
```diff
@@ -0,0 +1,71 @@
...
+#   evaluation_rpm: [0, 3000]
+#   rpm_tolerance: 100
+
+
+temperature_curve_silent:
+  - range: [0, 65]
+    evaluation_pwm: [20, 25]
+    tolerance_pwm: 6
+    evaluation_rpm: [0, 0]
+    rpm_tolerance: 10000
```
miczyg1: This RPM tolerance is huge. It is beyond what typical fan can spin up to. Is this really correct?

platform-configs/include/sensors/default-fan-curve-config.yaml
```diff
@@ -0,0 +1,71 @@
...
+#   evaluation_pwm: [20, 25]
+#     # range of values in this temperature range, can be undefined if
+#     # measuring pwm is not supported on given platform
+#     # accepted tolerance of pwm values - by how much can it differ from the
+#     # expected value
+#   tolerance_pwm: 6
+#   evaluation_rpm: [0, 3000]
+#     # accepted tolerance of rpm values - by how much can it differ from the
+#     # expected value
+#   tolerance_rpm: 100
```
miczyg1: Is it tolerance_rpm or rpm_tolerance? It is inconsistent across the file.

platform-configs/include/sensors/protectli-vpxxxx-fan-curve-config.yaml
```diff
@@ -0,0 +1,62 @@
...
+    tolerance_rpm: 300
+
+  - range: [85, 100]
+    evaluation_pwm: [100, 100]
+    tolerance_pwm: 6
+    evaluation_rpm: [3100, 3100]
+    tolerance_rpm: 400
+
+
+
```
miczyg1: Too many blank lines

platform-configs/include/sensors/protectli-vpxxxx-fan-curve-config.yaml
```diff
@@ -0,0 +1,62 @@
...
+  - range: [0, 30]
+    evaluation_pwm: [0, 1]
+    tolerance_pwm: 6
+    evaluation_rpm: [250, 250]
+    tolerance_rpm: 250
+
+  - range: [30, 40]
+    evaluation_pwm: [0, 40]
+    tolerance_pwm: 6
+    evaluation_rpm: [800, 800] # Anywhere between 0 and 1300 + 200 tolerance
```
miczyg1: It actually is 1400 + 200 tolerance.

But why is it so big? Can't it be just `[200, 1300]` with 200 tolerance like the comment really says?

platform-configs/raptor-cs_talos2.robot
```diff
@@ -190,6 +190,28 @@ ${STABILITY_DETECTION_WARMBOOT_ITERATIONS}=         2
...
+# The name of the sensor in `sensors` command if FAN_RPM_MEASUREMENT_METHOD
+# is set to lm-sensors. For example `w83795g-i2c-1-2f`. `none` if lm-sensors
+# is not used
+${FAN_RPM_MEASUREMENT_SENSOR}=                      ${TBD}
+# Kernel module that might need to be enabled using modprobe in order to use
+# the sensor. Dictionary keys:
+# - module - name of the kernel module
+# - force_id - optional force_id arg for modprobe
+&{FAN_RPM_MEASUREMENT_SENSOR_MODULE}=               module=w83795
+
```
miczyg1: If we use PWM measurement, then we don't need RPM measurement right? I think the RPM measurement variables should be None

miczyg1 (changes requested):

platform-configs/include/sensors/default-fan-curve-config.yaml
philipanda: I chose to make the default curves "no curves", so that everything should pass. Should it be more realistic and more of an approximate of an average fan?

dasharo-compatibility/coreboot-fan-control.robot
philipanda: https://github.com/Dasharo/open-source-firmware-validation/compare/cddd9af%5E..cddd9af#diff-18cd6de46364ae2c67ffd2e3dbde89ae3299f4e2708eaf83001ef97c18d69b2cL38

dasharo-compatibility/coreboot-fan-control.robot
philipanda: It was like that before, I've only changed the implementation.
https://github.com/Dasharo/open-source-firmware-validation/compare/cddd9af%5E..cddd9af#

dasharo-compatibility/coreboot-fan-control.robot
philipanda: > Also we must not use any specific library to execute commands, right? We have a wrappers for executing commands in OS that will use the needed library.

That's an ancient test case. I've only replaced the keywords that read the RPM and cpu temperature, because I've cleaned them up into a library. It's a good point that it could use some attention, but should it be done in this PR?

dasharo-compatibility/coreboot-fan-control.robot
philipanda: Again, not my changes. It won't hurt fixing a typo though?

lib/sensors.robot
philipanda: I've made it so that it will just not filter by sensor if `fan_rpm_measurement_sensor` is none. I think filtering by sensor name should be optional.

https://github.com/Dasharo/open-source-firmware-validation/compare/20dcc78%5E..20dcc78#diff-1650d666a365ac4e12449afd31a0e1ff7bc6215193f74ba46540a97a9eb59f7eL89-R95

platform-configs/include/default.robot
philipanda: https://github.com/Dasharo/open-source-firmware-validation/compare/20dcc783386252686c397010955a2eccc0eb1745..2388f0fe119a63b3846a8fb588a67ab0a5fbf918#diff-07c15f12435114d3af82c6b0f99da527081f7f6bb6455aee02cbdf868630f838L256-L257

lib/sensors.robot
philipanda: https://github.com/Dasharo/open-source-firmware-validation/compare/20dcc78%5E..20dcc78#diff-1650d666a365ac4e12449afd31a0e1ff7bc6215193f74ba46540a97a9eb59f7eL80

platform-configs/include/sensors/default-fan-curve-config.yaml
miczyg1: I don't understand a single thing from above explanation.

dasharo-compatibility/coreboot-fan-control.robot
miczyg1: Maybe it is not yours, but in the test case you modify now. Worth fixing while at it.

dasharo-compatibility/coreboot-fan-control.robot
miczyg1: >  It's a good point that it could use some attention, but should it be done in this PR?

Not fixing it will mean this test case will only work with serial console. How can I approve PR with a test like that? It should be a simple replacement of the keyword

dasharo-compatibility/coreboot-fan-control.robot
miczyg1: I saw it later during review. Looks weird at least. EIther remove CURRENT or make it `Get Current CPU Temperature`.

Also not entirely my fault, that low quality code lands in the tree, but let's fix these small things. They don't cost a lot if it is just a rename.

platform-configs/include/sensors/protectli-vpxxxx-fan-curve-config.yaml
philipanda: The tolerance in the comment does not mean the same thing as the `tolerance_rpm`. The expected RPM is calculated using a linear function joining the two extremes of `evaluation_rpm`. It's the same method used before in the custom-fan-curve test suite. The `tolerance_rpm` parameter defines by how much the actual RPM can differ from the one calculated from a linear function, to still be treated as a pass.

But in this deadband zone of 30-40 C the RPM is not linear. It can be 0 when the CPU temperature is at 0C, but as well it can be 0 when at 40C, because the temperature has just started increasing.

With the `evaluation_rpm` equal to `[800, 800]` the RPM is expected to be constant at 800 RPM. With the `tolerance_rpm` set to `800` an acceptable value will be anything from 0 to 800+800=1600.

Now I think it might be better to add an additional parameter that defines what type of function is used for a given range.

It could be a constant (one value + tolerance), a range (two values, without tolerance), a linear function (two values + tolerence like now), maybe even some more complex functions, like a parabola (three "range" values and three "evaluation" values + tolerance).

This way we could describe the behaviour at the deadband more clearly and maybe define some non-linear curves in the future. What do you think?

platform-configs/raptor-cs_talos2.robot
philipanda: This was the only platform that supported Coreboot Fan Control test suite and had a keyword in keywords.robot for extracting the temperature and RPM that would work only on raptor cs talos.
Reading PWM from hwmon was only used on a couple novacustom laptops. That's my bad it was here.

https://github.com/Dasharo/open-source-firmware-validation/compare/2388f0fe119a63b3846a8fb588a67ab0a5fbf918..ebc95c56b6792117f5df43aa2b0a331ce5530112#diff-9c1adfa0b7e2a22cea71c9d2a22e4ded564acfd4bcc4d1435b81a4d92d84d30aL199-R208

platform-configs/include/sensors/protectli-vpxxxx-fan-curve-config.yaml
philipanda: https://github.com/Dasharo/open-source-firmware-validation/compare/2388f0fe119a63b3846a8fb588a67ab0a5fbf918..ebc95c56b6792117f5df43aa2b0a331ce5530112#diff-08a44cf2654f5076f8d9fcf3f6b4783e0cc60a2a286d8424adb1aa09ce74d57fL30-L31

platform-configs/include/sensors/default-fan-curve-config.yaml
philipanda: Should be consistent now.

https://github.com/Dasharo/open-source-firmware-validation/compare/2388f0fe119a63b3846a8fb588a67ab0a5fbf918..ebc95c56b6792117f5df43aa2b0a331ce5530112#diff-1e02bf3203ee8997b15391c3f52fd1683a99b3ac96208e2b78a431d4b46e8490L25-R26

platform-configs/include/sensors/default-fan-curve-config.yaml
philipanda: The exact value of tolerance here is random. It's big so that any actual RPM value would be considered a pass by default and not cause an undefined variable error.

Specifying the actual curves used in a platform requires creating a new config file and setting `${CUSTOM_FAN_CURVE_FILE}` to it's filename.

Do you think the default config should instead cause FAILS, because someone must have forgot to define the curves?

In the previous comment I suggested that maybe the default config should define some pretty reasonable RPM curves, that could, in theory, be achieved on some actual fan. Now the tolerance is huge so that any RPM is considered valid.

dasharo-compatibility/coreboot-fan-control.robot
philipanda: Oh, I thought I've commited your suggestion right away. Must have missclicked something.

dasharo-compatibility/coreboot-fan-control.robot
philipanda: The keywords used to log-in to linux should also be updated then
https://github.com/Dasharo/open-source-firmware-validation/compare/767f8875a162af6464e95bb1f752b392ce045da7..805bbff5e989d68c70d9fc3345a892bb6ca58265

But in that case I would consider removing this test suite and `cpu-fan-speed-measure.robot` altogether, as they are only testing:
1) CFN001.001 - if the temperature and fan RPM can be read at all
2) CFN002.001- if the RPM and CPU temperature increase when a stresstest is run
3) FAN001.001 - Again, if the fan RPM can be read at all

It seems pretty redundant. It might be joined into one test suite.

dasharo-compatibility/coreboot-fan-control.robot
philipanda: It was fixed immediately, the compare in my previous comment shows that. Is there anything more left to resolve this conversation?

dasharo-compatibility/coreboot-fan-control.robot
miczyg1: > The keywords used to log-in to linux should also be updated then

Then let's do it. Otherwise something may just not work if we keep old keywords.

> It seems pretty redundant. It might be joined into one test suite.

You are probably right. Also ` if the temperature and fan RPM can be read at all` then CFC suite must work.
`if the RPM and CPU temperature increase when a stresstest is run` is already done here in CFC right?
` Again, if the fan RPM can be read at all` - really redundant. Seems like the could be simply dropped. We would have to track down which platforms have these tests in the scope. I would leave it for a separate cleanup PR though.

lib/sensors.robot
miczyg1: If you set it to empty, then the test will probe all sensors, and you may get multiple results (i.e risking t ocheck the values which are not relevant for the test) . The test may simply fail because of it... All we have to do is to fail the test if the sensor is not defined. Otherwise we may end up in unexpected results or silent PASSes where there should be a fail.

platform-configs/include/sensors/default-fan-curve-config.yaml
miczyg1: By the looks of the values:

```
    evaluation_pwm: [20, 25]
    tolerance_pwm: 6
    evaluation_rpm: [0, 0]
    rpm_tolerance: 10000
```

it seems that this range define a test for PWM rather than RPM. Right? So rpm_tolerane should be 0, because we have pwm_tolerance. Either pwm_tolerance or rpm_tolerance should be non-zero. Never both should be zero or non-zero at the same time. Or I misunderstood something?

platform-configs/include/sensors/protectli-vpxxxx-fan-curve-config.yaml
miczyg1: And my proposal is actually better: `[200, 1300]` with 200, which translates to:

200 - 200 = 0
1300 + 200 = 1500

Gives teh same as `Anywhere between 0 and 1300 + 200 tolerance`. The way it is described right now misuses the term tolerance. Tolerance is a value of deviation from the standard operating range. So

`evaluation_rpm: [800, 800] # Anywhere between 0 and 1300 + 200 tolerance` is incorrect. It should either be:

1. `evaluation_rpm: [200, 1300] # Anywhere between 0 and 1300 + 200 tolerance` with `tolerance_rpm: 200` OR
2.  `evaluation_rpm: [800, 800] # Anywhere between 0 and 800 + 800 tolerance` with `tolerance_rpm: 800`. 

See the difference? Using a big deviation means that measurements are very inaccurate, which doesn't look good in a test that checks the quality of the firmware. Tolerance is a margin of error acceptable by the test and should be as less as possible. If it is too high it means that measurements are unreliable and meaningless.

Of course RPM and temperature inertia may be different depending on platform and fan.

dasharo-compatibility/coreboot-fan-control.robot
philipanda: https://github.com/Dasharo/open-source-firmware-validation/issues/692

platform-configs/include/sensors/protectli-vpxxxx-fan-curve-config.yaml
philipanda: Is it all about the comment? Because it was wrong. I've been fiddling with the values to make sense and forgot to change/remove it.


That's roughly the shape of the theoretical performance curve at this moment, taking the histeresis into account.
Start spinning at 40, stop spinning at 30, max speed at 75.
image

Please help me understand how linear function can roughly describe what's happening at temps <30, 40>.
That's this range:
image

You say we should define the linear function like this:
(Orange solid line is the function, dotted lines mark the tolerance, so the real value has to fit in between the dotted lines)
image
But it won't cover the red line - when the fans are not spinning yet and only will start so when the CPU reaches 40 degrees.

The [800,800]+800 looks like that:
image
And covers Both lines.

I agree that the exact values could be more fine-tuned as I've added a rather large margin of error.

lib/sensors.robot
philipanda: https://github.com/Dasharo/open-source-firmware-validation/compare/ca0a718d29e3850a1f19a4cf8d18fb2295a75000..8ea3cc9c2dbeac2dc53229a401c56f26e353ddd4

platform-configs/include/sensors/default-fan-curve-config.yaml
philipanda: > it seems that this range define a test for PWM rather than RPM. Right? 

The values for PWM are what was used in Custom Fan Curve before. All I wanted to achieve is not to break it with my changes. Now that we know that the old test was dead code, these values are not needed.

I'd remove the file for now and instead add checks in the tests, so that they fail if no custom fan curves are defined for the device. Or maybe you know of some values that would make sense as default ones?

philipanda: https://github.com/Dasharo/open-source-firmware-validation/compare/8b1fa1c%5E..8b1fa1c

platform-configs/include/sensors/protectli-vpxxxx-fan-curve-config.yaml
miczyg1: > The [800,800]+800 looks like that:

Essentially [200 , 1400] + 200 should cover the same range right?

philipanda: evaluation_rpm of [200, 1400] for temperature range [30, 40] means that the expected value is on a line connecting the points (30, 200) and (40, 1400) (Marked on the graphs with orange lines)

The tolerance is how much the actual value can be off this linear function to still be considered a pass (Marked on the graphs with dotted orange lines)

In the temperatures range of <30, 40> the rpm can be anywhere from ~0 to over a thousand. Because of histeresis there is no proportionality between the temperature and RPM in this range.
That's why the curve I used in this range is a flat line at 800 RPM +/0 800 to allow anything between 0 and 1600 RPM no matter the temperature.

miczyg1: Aaah ok, because I thought it works differently:

Assumed that in temperature range [30,40] the RPM may be anywhere in [200, 1400] range with 200 tolerance. Meaning it can be anywhere between [0, 1600].

Ok, so the final conclusion is `Is it all about the comment? Because it was wrong. `. Please fix the comment, and it should be good.

miczyg1:
@philipandag we have all discussion resolved now. I would say it is ready to go in. But I would like to see the logs fro the tests run from the latest commit from this PR.

philipanda:
[custom-fan-curve.robot_log.zip](https://github.com/user-attachments/files/18763317/custom-fan-curve.robot_log.zip)
The logs from the tests after the rebases and review changes - performed on [43a0eac](https://github.com/Dasharo/open-source-firmware-validation/pull/635/commits/43a0eac2910e734464997f08d5335986f8499f24).

lib/sensors.robot
```diff
@@ -0,0 +1,123 @@
...
+    ${lm_sensors_used}=    Catenate
+    ...    '''${cpu_temperature_measurement_method}''' == '''system76-acpi''' or
+    ...    '''${cpu_temperature_measurement_method}''' == '''lm-sensors''' or
+    ...    '''${fan_rpm_measurement_method}''' != '''lm-sensors''' or
+    ...    '''${fan_pwm_measurement_method}''' == '''lm-sensors'''
+    ${lm_sensors_used}=    Evaluate    ${lm_sensors_used}
+
+    IF    ${lm_sensors_used} == ${TRUE}
+        Detect Or Install Package    lm-sensors
+        Execute Command In Terminal    sudo sensors-detect --auto
```
miczyg1: Detecting sensors must happen after modules are loaded.

miczyg1 (changes requested):
I can see the test failed because the fan speed was decreasing slower than the temperature. Temperature can get very low quickly on the sensor, but the fan often needs couple of seconds to slow down.

The test may be failing all the time because of that. I wonder if it is really necessary to measure the fall time of fan speed based o temperature. If anything we may only check if the fan slows down to the expected temperature after 30 seconds for example. Right now I see the stress tests lasts 3m, and measurements are made with 0.05 minute intervals (3 seconds, 60 iterations). That is more than the stress test length, because there are delays caused by execution of test instructions.

lib/sensors.robot
philipanda: It seems to have been working fine though? At least in this one case it didn't matter much.
https://github.com/Dasharo/open-source-firmware-validation/compare/43a0eac2910e734464997f08d5335986f8499f24..0ab2b98ce42a75edd4ab18f839b33c6f3d9ae213

philipanda:
@miczyg1 

> Temperature can get very low quickly on the sensor, but the fan often needs couple of seconds to slow down.

At first I thought it is caused by the fan just being a physical object with a mass, but then the RPM would lag by a couple seconds at most. The tests show that the fan stays at ~1600 RPM for 15 seconds. It is definitely something caused by the software controlling it, not the fan itself.

> The test may be failing all the time because of that. 

If the code controlling the fans introduces some additional lag when slowing down, then wouldn't that just mean that the fan curves are wrong? That the fan behavior is more complex than we assume it is in the test suite?
Are you sure that after some time the fans will eventually slow down? Wouldn't we want to test how much time it takes for them to stop if it is controlled by firmware?

dasharo-compatibility/coreboot-fan-control.robot
```diff
@@ -33,7 +34,8 @@ CFN001.001 CPU temperature and fan speed can be read (Debian)
     Boot System Or From Connected Disk    ubuntu
     Login To Linux
     Switch To Root User
-    ${rpm}    ${temperature}=    Get CPU Temperature And CPU Fan Speed
```
philipanda, line 36: @miczyg1 
We have decided that the slowing down of the fans should not be included in the test.

In that case I would make the stress-ng timeout very long and kill the process once the measurements are done.
It will be more reliable than guessing how by much executing the test instructions will affect the time span across which the measurements are performed.

It seemed to work for a short 1m test. I'll rerun the whole suite to make sure everything works.
https://github.com/Dasharo/open-source-firmware-valida
[custom-fan-curve.robot_log.zip](https://github.com/user-attachments/files/18799670/custom-fan-curve.robot_log.zip)
https://github.com/Dasharo/open-source-firmware-validation/compare/91df194%5E..91df194

philipanda: Not measuring the fan when it slows down would make the tests useless. It will only measure how the fans behave when the CPU is at it's highest temperature, while the point is to cover the whole temperature range.

Too bad I believed that the tests were making sense before, because they were not.
The duration and measurement interval are given in minutes, while even one minute is very long for such test.
The tests are waiting for 5 seconds after starting the stress, which effectively skips the temperature rising.
Originally the test duration was around 30 minutes, both the fan speed and CPU temperature must have been at their peaks for the whole time.

I don't know what was the original intent behind those tests, but trying to add a new test case was a bad idea, as they definitely had a completely different purpose than what we would like to achieve here. 

To make the PASSes actually usefull we should try to either increase the temperature of the CPU slowly, for example by increasing the number of workers as time passes, or make the measurements as quickly and often and possible.

philipanda: And the fans are in fact not slowing down below ~1600 RPM when on performance curve.
I've left it be for a couple minutes and the RPMs don't fall
```
root@3mdeb:/home/ubuntu# sensors | grep -E "fan|C"
Adapter: PCI adapter
Adapter: PCI adapter
Composite:    +27.9°C  (low  = -273.1°C, high = +81.8°C)
                       (crit = +84.8°C)
Sensor 1:     +27.9°C  (low  = -273.1°C, high = +65261.8°C)
Sensor 2:     +28.9°C  (low  = -273.1°C, high = +65261.8°C)
fan1:        1679 RPM  (min =   28 RPM)
fan2:        1704 RPM  (min =   10 RPM)
temp1:        +29.0°C  (low  =  +0.0°C, high = +127.0°C)  sensor = Intel PECI
temp2:       -128.0°C  (low  = +89.0°C, high = -126.0°C)
temp3:       -128.0°C  (low  = -47.0°C, high = +30.0°C)
Package id 0:  +26.0°C  (high = +100.0°C, crit = +100.0°C)
Core 0:        +21.0°C  (high = +100.0°C, crit = +100.0°C)
Core 4:        +24.0°C  (high = +100.0°C, crit = +100.0°C)
Core 8:        +22.0°C  (high = +100.0°C, crit = +100.0°C)
Core 9:        +22.0°C  (high = +100.0°C, crit = +100.0°C)
Core 10:       +22.0°C  (high = +100.0°C, crit = +100.0°C)
Core 11:       +22.0°C  (high = +100.0°C, crit = +100.0°C)
Core 12:       +23.0°C  (high = +100.0°C, crit = +100.0°C)
Core 13:       +23.0°C  (high = +100.0°C, crit = +100.0°C)
Core 14:       +23.0°C  (high = +100.0°C, crit = +100.0°C)
Core 15:       +23.0°C  (high = +100.0°C, crit = +100.0°C)
```

philipanda: The time it takes for the CPU to jump from ~25C to ~80C is less than 0.1s. We can not test this feature without some kind of gradation of the load.

Unless we want to test the curve when the cpu cools down. It should be fine provided the fans will actually slow down on the performance curve. There is either a bug in the code controlling the fans or we cannot physically achieve the CPU temperatures required to test the performance curve.

If no sensor is able to read the actual CPU temperature used to control the fans, then where the code controlling the fans gets the temperature from?

lib/sensors.robot
miczyg1: @philipandag probably because the sensor detection was run before?

dasharo-compatibility/coreboot-fan-control.robot
miczyg1: The thing is the fans are controlled by HW, there is no code that do it the software way as you think.
The whole logic is implemented by HW, so we have to deal with it the way it is.

There is no simple way to force a certain temperature on CPU for convenience of testing. When the CPU is loaded, it will usually try to utilize its full performance given the restrictions placed on it (throttling temperature, power limits, etc).

If you would like to try reaching certain temperatures, you would have to prepare some sophisticated load scenarios that would stress the CPU only partially. But then you enter the hell of CPU-specific list of tests and parameters to stress-ng, because every CPU has different performance and would react differently.

> And the fans are in fact not slowing down below ~1600 RPM when on performance curve.
I've left it be for a couple minutes and the RPMs don't fall

The `Package id 0: ` and `temp1` should be showing the same temperature, but or some reason the HW sensor the ITE SIO has an offset, which may impact the fan control too (the chip may think the CPU is hotter or cooler than in reality). It looks like some HW bug. Question is if the fans are spinning at 1600RPM before you start the stress test?

tl;dr

I tried to advice as best as possible by proposing only rising temperature test, because it is extremely hard to get the results from intermediate temperature points. The way I tested it was simply max temperature and min temperature, because I have no idea how else it can be tested + I'm not a pro benchmarker/tester

philipanda: Using stress-ng it was possible to cover the temperature range a bit better. There are still some areas not covered by the measurements around 40-60 C despite increasing the cpu load by 1 percent at a time.


image

lib/sensors.robot
philipanda: It must have been

dasharo-compatibility/coreboot-fan-control.robot
philipanda: Changes:
- dropping configurable percent of most out-of-place measurements to eliminate noise
- passing the test if the percent of invalid measurements is below configurable amount

The graphs include the tolerance ranges to allow for easier manual validation in case of a fail.

The tests below Passed with:
- dropping of 10% of the most invalid results
- passing if the total amount of invalid measurements after the drop is less than 10%.
The percentage values are defined in `protectli-vp66xx-fan-curve-config.yaml` for this one platform.

[2025_02_18_15_51_08.zip](https://github.com/user-attachments/files/18847572/2025_02_18_15_51_08.zip)

philipanda: @miczyg1
I's me who started this conversation, but I'd like to ask you to resolve it if the results above are looking good to you

miczyg1:
Can I see the results from latest tests code revision?

philipanda:
> Can I see the results from latest tests code revision?

@miczyg1 
They are here https://github.com/Dasharo/open-source-firmware-validation/pull/635#discussion_r1959940280

miczyg1:
LGTM. @philipanda let's just fix the `[ ERROR ] Error in file '/home/runner/work/open-source-firmware-validation/open-source-firmware-validation/keywords.robot' on line 22: Variable file 'platform-configs/fan-curve-config.yaml' does not exist` error for the QEMU. I know it may not necessarily make the workflows pass.

miczyg1:
Okay right now we got `No match found for 'to boot directly`. Even on QEMU :) Well, nothing to do about it since it appears on HW as well and was not introduced in this PR...

miczyg1 (approved):

PR #638, opened 2024-12-17
platform-configs/include/protectli-vp32xx.robot: add eth ports order

wiktormowinski:


macpijan:
@wiktormowinski is this finalized? Any test logs? Any reviewer to assign? I guess @miczyg1 could take a look once it's ready as he was involved in this test.

wiktormowinski:
@macpijan nowhere near, currently i am working on other issues, however i plan on completing this PR today

macpijan:
@wiktormowinski I recommend looking at https://github.blog/news-insights/product-news/introducing-draft-pull-requests/

miczyg1:
> @wiktormowinski is this finalized? Any test logs? Any reviewer to assign? I guess @miczyg1 could take a look once it's ready as he was involved in this test.

Generally it looks ok. THere should not be anything more needed for this test. Simply the ID must match the controllers PCI device ID and MACs have to exactly match those found on the platform in lab.

wiktormowinski:
@miczyg1 1. adding these is easy as log as there is a sticker on port or platform is aviable on snipeit and even if they are free there is some weird stuff going on them (for example on 2410 ubuntu is just going to grub and that's it)
as of now we are still missing (because of reasons mentioned above): 
- 24xx series (all 3 platforms)

2. and because most of the platforms are being used right now and half of them doesnt even have dasharo on it, making this test case useless i can't run the test case on every single platform in relatively quick time.
3. also we have 2x VP4670 so i only included the one with 2.0 next to it (cuz 1<2 so it's better)

miczyg1:
> @miczyg1 1. adding these is easy as log as there is a sticker on port or platform is aviable on snipeit and even if they are free there is some weird stuff going on them (for example on 2410 ubuntu is just going to grub and that's it) as of now we are still missing (because of reasons mentioned above):
> 
> * 24xx series (all 3 platforms)
> 
> 2. and because most of the platforms are being used right now and half of them doesnt even have dasharo on it, making this test case useless i can't run the test case on every single platform in relatively quick time.
> 3. also we have 2x VP4670 so i only included the one with 2.0 next to it (cuz 1<2 so it's better)

Ad. 1. Each Protectli platform can boot DTS over iPXE. From DTS shell you may extract all the info you need.
Ad. 2. No other solution than to wait till it is free or ask the person who checked out the platform to give it to you for 5 minutes...
Ad. 3. Right... This is problematic to have different revisions of the same HW platform in lab. Mayeb the test could be extended to have a 2-dimensional array with two sets of MACs for that particular platform, then one or the other set should match. Test could be modified accordingly.

wiktormowinski:
alright that's all of them, thanks for the tip @miczyg1, i forgor we can use dts for that

https://github.com/Dasharo/open-source-firmware-validation/pull/638/commits/563953cbafc8451c2ea27b90bf64e5f52032bed9 i've also added skip because this test raised many questions when it was running on every platform ;p

platform-configs/include/protectli-vp24xx.robot
```diff
@@ -12,3 +12,5 @@ ${MAX_CPU_TEMP}=                95
 ${DMIDECODE_MANUFACTURER}=      Protectli
 ${DMIDECODE_VENDOR}=            3mdeb
 ${DMIDECODE_FAMILY}=            Vault Pro
+
+${ETHERNET_ID}=                 8086:1539
```
miczyg1: I see vp2420 has different port than vp2410. VP2430 will also have different. I would move it to the board-specific config rather than include into common vp24xx config.

miczyg1 (changes requested):
vp24xx is not a series of the same platform variant, it has 3 entirely different platofrm

platform-configs/include/protectli-vp24xx.robot
wiktormowinski: k makes sense, https://github.com/Dasharo/open-source-firmware-validation/pull/638/commits/30d89c0f00cc0495d0fe4d91dbdbcb01bc62d2a3

SebastianCzapla:
image
Works good on vp2430, and order is not messed up when we plug additional Ethernet connections (for example using loopback to test eth speeds)

(tested via cherrypick)

platform-configs/include/protectli-vp24xx.robot
miczyg1: You should have changed only the `platform-configs/include/protectli-vp24xx.robot` file. The rest was ok...

wiktormowinski: myy bad, wanted to act smart. Here's the actual fix: https://github.com/Dasharo/open-source-firmware-validation/pull/638/commits/fba7fd6aa85cb14dbaf3d4745bfcd3584dab650e

miczyg1 (approved):

PR #640, opened 2024-12-18
Change dts extensions checkpoint

DaniilKl:
Related PR:
https://github.com/Dasharo/meta-dts/pull/208

### DIFF
--- a/dts/dts-e2e.robot
+++ b/dts/dts-e2e.robot
@@ -31,8 +31,6 @@ ${DTS_SPECIFICATION_WARN}=          Does it match your actual specification? (Y|
 ${DTS_DEPLOY_WARN}=                 Do you want to deploy this Dasharo Firmware on your platform (Y|n)
 ${DTS_HW_PROBE_WARN}=               Do you want to participate in this project?
 ${DTS_HEADS_SWITCH_QUESTION}=       Would you like to switch to Dasharo heads firmware? (Y|n)
-${DPP_PACKAGES_CHECKPOINT}=
-...                                 If you did not buy any DPP\npackages - feel free to continue.\n${DTS_CONFIRM_CHECKPOINT}
 # DTS initial deployment menupoints:
 ${DTS_DCR_UEFI_MENUPOINT}=          Community version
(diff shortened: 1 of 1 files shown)


m-iwanicki:
@DaniilKl why do you target `main` instead of `develop`?

DaniilKl:
> why do you target main instead of develop?

Well, you have catched me.

m-iwanicki (approved):

m-iwanicki (changes requested):
Fix documentation as it doesn't check any warning. Maybe even change keyword name because keyword just writes credentials from global variabe and waits for `Press Enter`.
Currently `Press Enter ...` will show only if you can't login or you can and packages were installed successfully: https://github.com/Dasharo/dts-scripts/blob/1cdf6313557f03895b446d2dd8bbd477a02e7a84/include/dts-functions.sh#L1498.

DaniilKl:
@m-iwanicki, I have deleted KW `Provide DPP Credentials Without Packages`, it seems that the checkpoint is now being printed always: 35c0352f9dd95ce1f1041657e21d7162ac234d00.

m-iwanicki (approved):
It won't be printed if `check_avail_dpp_packages` or `install_all_dpp_packages` fails, in that case `Provide DPP Credentials` will get stuck and fail after timeout. If that's how it should work then you can merge.
PS: is this how it should work? https://github.com/Dasharo/open-source-firmware-validation/blob/35c0352f9dd95ce1f1041657e21d7162ac234d00/dts/dts-e2e.robot#L604? Won't `Go Through Update` get stuck?

DaniilKl:
> It won't be printed if check_avail_dpp_packages or install_all_dpp_packages fails, in that case Provide DPP Credentials will get stuck and fail after timeout. If that's how it should work then you can merge.

If smth in DTS unexpectedly fails - the tests should fail too, this is the way.

> PS: is this how it should work? 

This one is interesting, did not notice it previously. No, it should not work this way, will check it.

DaniilKl:
> Won't Go Through Update get stuck?

Fixed here: 796e0c192c8a5da205115f764b8088c75ac4c047. But surprisingly, the test passed in both cases, with the changes and without.

m-iwanicki (approved):

PR #507, opened 2024-09-18
dasharo-security/tpm-support.robot: Refactor TPM version and support …

SebastianCzapla:
…tests

Currently, TPM Support and TPM Version test are split into different tests groups. Merging of version and support checks into one allows us to skip few extra reboot cycles during testing. Additionally, it provides opportunity to add coverage for TPM 1.2

PR #641, opened 2024-12-18
VP2430 v0.9.0-rc2 Fixes

SebastianCzapla:
Based on: https://github.com/Dasharo/open-source-firmware-validation/pull/507

### DIFF
--- a/dasharo-compatibility/esp-scanning.robot
+++ b/dasharo-compatibility/esp-scanning.robot
@@ -102,8 +102,6 @@ ESP002.001 ESP Scan after deleting additional .efi files
     [Documentation]    This test aims to verify that none of the systems linger
     ...    on in the boot menu after we've deleted their files from /EFI/.
     Skip If    not ${TESTS_IN_FIRMWARE_SUPPORT}    ESP002.001 not supported
-    Skip If    not ${ESP_SCANNING_SUPPORT}    ESP002.001 not supported
-
     Power On
     Clear Out EFI Partition
     Power On
--- a/dasharo-compatibility/esp-scanning.robot
+++ b/dasharo-compatibility/esp-scanning.robot
@@ -13,29 +13,14 @@ Resource            ../keywords.robot
 Resource            ../keys.robot
 
 # Library    ../lib/QemuMonitor.py    /tmp/qmp-socket
-# Required setup keywords:
-#    Prepare Test Suite - generic setup keyword for all tests
-# Required teardown keywords:
-#    Log Out And Close Connection - generic setup keyword for all tests,
-#    closes all connections to DUT and PiKVM
-Suite Setup         Run Keywords
-...                     Prepare Test Suite
-...                     AND
-...                     Skip If    not ${ESP_SCANNING_SUPPORT}    ESP scanning tests not supported
-...                     AND
-...                     Prepare Required Files For Qemu
-...                     AND
-...                     Prepare EFI Partition With System Files
-Suite Teardown      Run Keywords
-...                     Clear Out EFI Partition    AND
-...                     Log Out And Close Connection
+Suite Setup         Setup Esp Scanning Suite
+Suite Teardown      Teardown Esp Scanning Suite
 
 
 *** Test Cases ***
 ESP001.001 ESP Scan with OS-specific .efi files added
     [Documentation]    This test aims to verify that any properly added .efi
     ...    files will have boot menu entries created for them.
-    Skip If    not ${TESTS_IN_FIRMWARE_SUPPORT}    ESP001.001 not supported
     Power On
     ${boot_menu}=    Enter Boot Menu Tianocore And Return Construction
     FOR    ${system}    IN    @{SYSTEMS_FOR_ESP_TESTING}
@@ -49,7 +34,6 @@ ESP003.001 ESP Scan ignores OSes on removable media
     [Documentation]    This test aims to verify that the bootable /EFI
     ...    partitions of removable media are ignored by the scan and aren't
     ...    listed in boot menu, except for DTS.
-    Skip If    not ${TESTS_IN_FIRMWARE_SUPPORT}    ESP003.001 not supported
     Power On
     Download ISO And Mount As USB    ${DL_CACHE_DIR}/CorePlus-current.iso
     ...    ${TINYCORE_URL}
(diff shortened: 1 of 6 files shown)


dasharo-compatibility/usb-detect.robot
```diff
@@ -81,7 +81,8 @@ UDT001.002 USB detection after warmboot
             ${usb}=    Evaluate    0
             Power On
             IF    '${PAYLOAD}' == 'tianocore'
-                ${menu}=    Enter Tianocore And Return Menu
+                Enter Boot Menu Tianocore
+                ${menu}=    Read From Terminal Until    ESC to exit
```
miczyg1: What is different than in `Enter Tianocore And Return Menu` ?

dasharo-compatibility/esp-scanning.robot
```diff
@@ -102,8 +102,6 @@ ESP002.001 ESP Scan after deleting additional .efi files
     [Documentation]    This test aims to verify that none of the systems linger
     ...    on in the boot menu after we've deleted their files from /EFI/.
     Skip If    not ${TESTS_IN_FIRMWARE_SUPPORT}    ESP002.001 not supported
-    Skip If    not ${ESP_SCANNING_SUPPORT}    ESP002.001 not supported
```
miczyg1, line 105: Why? It may be disabled in certain builds.

dasharo-compatibility/dcu.robot
```diff
@@ -42,7 +42,7 @@ DCU001.001 Change the UUID
     Flash Firmware    ${FW_COPY}
     Make Sure New Firmware Is Booted After Flashing
 
-    Power On
```
miczyg1: Do we get rid of `Power On` now? THere are still hundreds of occurrences of Power On that should be changed to `Power Cycle On` then... Why do we change `Power On` to `Power Cycle On` only in selected places?

dasharo-security/tpm-support.robot
```diff
@@ -26,16 +26,18 @@ Suite Teardown      Run Keyword
...
+    ...    detected and logged correctly by FW via cbmem, directly in Ubuntu
     Skip If    not ${TESTS_IN_UBUNTU_SUPPORT}    TPM001.001 not supported
     Power On
     Boot System Or From Connected Disk    ubuntu
     Login To Linux
     Switch To Root User
-    Get Cbmem From Cloud
-    ${out}=    Execute Command In Terminal    cbmem -L
-    Should Contain Any    ${out}    TPM2 log    TCPA log
+    ${result}=    Run Keyword And Ignore Error    Validate Expected TPM Via Cbmem Chip
```
miczyg1:
```suggestion
    ${result}=    Run Keyword And Ignore Error    Validate Expected TPM Chip Via Cbmem
```

It is not gramatically correct. Please change the keyword name

keywords.robot
```diff
@@ -1349,7 +1349,7 @@ Reboot Via Ubuntu By Tianocore
     [Documentation]    Reboot system with Ubuntu installed on the DUT while
     ...    already logged into Tianocore.
     ${boot_menu}=    Enter Boot Menu Tianocore And Return Construction
-    Enter Submenu From Snapshot    ${boot_menu}    ubuntu
+    Enter Submenu From Snapshot    ${boot_menu}    Ubuntu
```
miczyg1: If there is no consistency in boot options, maybe check for the option ignoring the case?

miczyg1 (changes requested):

dasharo-compatibility/usb-detect.robot
SebastianCzapla: Keyword no longer exists.
image

dasharo-compatibility/esp-scanning.robot
SebastianCzapla: We check it already in the suite setup, this test was the only one that still had it, and since it's checked on suite setup we would not even reach this code.

image

dasharo-compatibility/dcu.robot
SebastianCzapla: In the early testing phase, I attributed an error I got to this keyword, but it was not right, it should be left as Power On. Will change it.

keywords.robot
SebastianCzapla: Good suggestion, will do.

dasharo-compatibility/esp-scanning.robot
miczyg1: Then we could move `${TESTS_IN_FIRMWARE_SUPPORT}` check to the Suite Setup too. it occurs in every test case...

miczyg1 (changes requested):

miczyg1:
Many changes I reviewd earlier simply vaporized. I must say I am a bit confused. Next time set the target branch properly so that only your changes are included. Reviewing files that are not supposed to be part of the PR is a waste of time and resources. Unless you had so many dirty fixes in so many places, then it would explain why they disappeared. But it still arises a question: what happened to them?

keywords.robot
miczyg1: Seems there is no change here anymore in the newest diff. What has been done to that part?

SebastianCzapla: My attempts at case ignore were unsuccessful, in meantime, dependency on this keyword was removed in this commit: https://github.com/Dasharo/open-source-firmware-validation/pull/659/commits/e720969dc1d22c0df34b57b20842628d2c29038d

As of now, the keyword Reboot Via Ubuntu By Tianocore is not used anywhere in OSFV.

SebastianCzapla:
I should have clarified immidiately, my bad. The now gone changes were due to this branch being rebased onto https://github.com/Dasharo/open-source-firmware-validation/pull/507/

Some of the other stale changes, that I removed were things like hacky power state variables, that are no longer valid, but were during the testing.

keywords.robot
miczyg1: Thanks for reference

dasharo-compatibility/esp-scanning.robot
SebastianCzapla: Added a new commit, which isolates Suite Setup and Teardown into local keywords, which should contain proper skip conditions

miczyg1 (approved):

miczyg1:
@SebastianCzapla can't fast-forward merge it. Needs a rebase

SebastianCzapla:
@miczyg1 rebased

PR #647, opened 2024-12-23
scripts/ci/ipxe-run.sh: fix script

EduKav1813:


macpijan:
@DaniilKl @m-iwanicki I assume this was related to DTS testing. Any idea if this is needed? I assume no, since we executed tests since then without this change merged.

m-iwanicki:
Probably not

DaniilKl:
I remember I reviewed this issue some time ago, but idk why this was not merged. Related to very old thread, when we removed `nfs` from our iPXE scripts from `boot.dasharo.com/dts/`, because we are not using `nfs` in DTS and it cauesd problems during DTS boot. I would say it is a logical change, but I am ok with that if it does not cause problems.

PR #637, opened 2024-12-17
contributing.md: Add guideline for creating documentation

philipanda:

PR #664, opened 2025-01-14
Documenting most important Keywords

philipanda:


keywords.robot
```diff
@@ -25,10 +25,29 @@ Variables       platform-configs/fan-curve-config.yaml
...
+    ...    is running
+    ...
+    ...    - ``${s2n_port}``: ``string`` - The port on which ser2net is
+    ...    listening
+    ...
+    ...    === Return Value ===
+    ...    - None
+    ...
+    ...    === Side Effects ===
+    ...    - Sets up serial connection via telnet. Allows using Telnet Library
```
mkopec: I'm not sure if `Side effects` is the right term here, if that's the entire point of the keyword :)

does it make sense to split into `Effect` and `Side effects`?

philipanda: I agree that calling it "Side" effects might be wrong.
But that's just what it's called when a procedure call produces some permanent changes, AFAIK.
Functional programming languages are said to have no "side effects" when calling functions.

And who we are to decide what is an "Effect" and what is a "Side Effect"? That's up to the user of the Keyword to decide. 

So maybe just "Effects"? Or "Impact"? Or just leave it as it is as I think it is obvious what the term means.

What do you think?

macpijan: It's been probably been discussed in where you proposed the template for this documentation, and I was not there (yet?).

My thought when seeing your discussion was that we have kind of two streams there.

One is the "regular" programming flow, where you have the input arguments and return some return values.

The other one is the "platform/DUT state" so to speak. So we have some (expected/assumed) state this keyword needs to operate properly, and we have also some resulting DUT state when this keyword finishes.

philipanda: It is possible for a keyword to have a requirement/side effect as in the "regular" programming flow, as global variables are used heavily.

Looking at it like that, a better naming scheme could be used, I suppose? 

I think in the case of PiKVM platforms the DUT_CONNECTION_METHOD global variable is sometimes changed to switch between telnet/ssh connection. That would be a purely "regular" programming side effect. No changes to the state of the platform are done. After switching to SSH and doing a reboot, trying to enter the boot menu on a PiKVM platform without changing the variable back would fail.

There might be more exceptions, but I don't recall any at this moment.

lib/options/options.robot
```diff
@@ -0,0 +1,192 @@
...
+
+
+*** Keywords ***
+Set UEFI Option
+    [Documentation]
+    ...    Sets an UEFI option.
+    ...    Implementations in ``/lib/options/``
+    ...
+    ...    === Requirements ===
+    ...    None
```
m-iwanicki, line 27: Not exactly, at least currently DCU version needs to be logged in linux with `dcu` repo cloned in home dir.
In cases like this we should probably force some kind of required beginning state and deal with that in keyword implementation so it could be used in any test. If there is no requirement then dcu version should be able to work with any (within a reason) beginning state.

philipanda: I agree, that bothers me too. Depending on the implementation of the keyword it works differently. The same is with the `Power On` keyword. In some places it is used as a way to reboot the platform, but on devices with `$POWER_CTRL==none` `Power On` just does nothing.

Documenting all the possible implemantations seems wrong. It's the implementations that should follow the "declaration". Otherwise creating new tests requires taking into account all the possible different implementations for all the platforms.

philipanda: >  dcu repo cloned in home dir.

That is not true. DCU is not cloned onto the DUT. DCU is called on the device running RF. Only the modified binaries are sent and flashed internally.

> needs to be logged in linux

That is true. And it is different from the uefi-setup-menu version.

m-iwanicki: > DCU is called on the device running RF

You are right I missed it (probably because I never used this command).

philipanda: At this moment I have no better idea than just say that the implementations might have additional requirements/side effects, and their documentation should be checked.

Ideally the implementations should not break the defined "interface" of the keyword, so that it can be used in new tests in the same way, no matter the DUT. I don't think the keywords themselves should be changed in this PR.

philipanda: https://github.com/Dasharo/open-source-firmware-validation/compare/44fd97e..0c00a6c5881625789c291c7070d52153ac52b879

m-iwanicki: > say that the implementations might have additional requirements/side effects, and their documentation should be checked

I'd say that makes using those keywords very unappealing, at least in tests.
Is dcu imported automatically based on `OPTIONS_LIB`?
If yes then I'm not sure how many of our tests work as I see a lot of `Set UEFI Option` without being logged in Linux.
I think only NovaCustom uses dcu and as we don't have any in our lab for remote testing then we just ignore `dcu` library.

m-iwanicki: If only we had some kind of `state` during tests so we could check e.g. where we are (e.g. logged in linux, somewhere in BIOS menu) then we could implement something like e.g.:

```robot
dcu.Set UEFI Option
    ${is_shell}=    In Linux Shell
    IF    not ${is_shell}
        Power On
        Login to Linux
    END 
```

keywords.robot
philipanda: The same suggestion was resolved in the base PR: https://github.com/Dasharo/open-source-firmware-validation/pull/637/files#diff-419d79ae7d440d49b3fa5fa7739f5f3a359e3ca42c3fa942e5604ae3474eb35b

lib/options/options.robot
philipanda: > I'd say that makes using those keywords very unappealing, at least in tests.

It very much does. That's just how they are at this moment. Documenting them just brought the much needed attention to this fact.

> Is dcu imported automatically based on OPTIONS_LIB? (...) I see a lot of Set UEFI Option without being logged in Linux.

It is imported automatically. Only on Novacustom laptops the default is changed to `dcu` though.

> I think only NovaCustom uses dcu and as we don't have any in our lab for remote testing then we just ignore dcu library.

The tests need to also support laptops. I don't think the documentation should hide important information, like that the keyword might not work at all in some scenarios. The keywords just need to be improved. Now that we are documenting them comprehensively the differences are visible and it will be more apparent what needs to be fixed.

> If only we had some kind of state during tests so we could check

It's possible to save state. Keywords like `Set Test Variable`, `Set Suite Variable` and `Set Global Variable` exist for that purpose. Keeping the state valid would be a bit harder as there are multiple ways in which the shell can be exited - mainly powering off the device. It's a good idea for a feature request

m-iwanicki: > Keywords like Set Test Variable, Set Suite Variable and Set Global Variable exist for that purpose

I'm not sure how well that would work. We already have a lot of variables. 
I had a thought that it would be easier if we had some kind of main abstract base python class e.g.:

```python
class Menu(ABC):
    def go_to_menu(menu: str):
        ...
    ...


class EDK2(Menu):
    ...


class Dasharo(EDK2):
    ...


class AMI(Menu):
    ...
```

with some generic keywords to move around that remember where we currently are, current screen state and could automatically move to completely different menu (e.g. by passing absolute or relative path).
Probably would need heavy rewrite and possibly use of python. I'm not sure if it could be implemented in robot and if it would be maintainable.

macpijan: Very interesting discussion out there.

Do we have some conclusions after this one, and documenting some of the keywords, to be put in the issue as potential improvements?

keywords.robot
macpijan: I keep thinking of this, whether tracking the state (or the expected state) of the platform could help with usability of keywords and reliability of environment.

Each kwd doing "something" with the platform, could set the expected state in the end of execution.
At the beginning, it could also check if it's being called with a proper initial platform state.

Where possible, we could also verify the expected state vs reality.

Maybe something worth discussing further.

lib/options/options.robot
philipanda: > We already have a lot of variables.

I agree. Some hermetization of the variables and keywords would be perfect, but that is not possible in RF. Keywords might be marked with `[Tags]    robot:private`, but only thing it does is produce a warning if you use the keyword outside of it's private scope. Even LSPs continue to suggest it outside of the scope.

> I had a thought that it would be easier if we had some kind of main abstract base python class e.g.:

I think that would be great

> I'm not sure if it could be implemented in robot and if it would be maintainable.

Yes, it would probably be a mess. RF is constantly being improved and allows more and more useful features of programming languages, but why not just use a proper programming language instead?

Rewriting some (if not most) of the keywords in Python is a good idea. Complex behaviors shouldn't be defined in RF. Maybe at first the ration of libraries/RF resources was balanced, but it seems that because of continuously expanding our test environment over time, the amount of logic defined in RF grew a little to large.

I am a fan of rewriting in python, especially that it doesn't actually need to be a total rewrite as we can call existing RF keywords from python:
```python
BuiltIn().import_resource('keywords.robot')
BuiltIn().run_keyword("My Keyword", arg1, arg2)
```
So the rewrite could be done iteratively, with lower risk

> Do we have some conclusions after this one

Yes, documenting the keywords exposed some potential for improvement. I'll sum it up in an issue

keywords.robot
philipanda: I think that would be great.

I wonder if something like this could be implemented to protect from invalid sequences during the creation of tests. Otherwise setting and checking the state would only find errors during the execution, and executing the tests takes a lot of time due to the interaction with the DUTs.


### Should the keywords which detect an invalid state should bring it to an expected one, or just fail?

I think in the long run the keywords should not implicitly change the state. A keyword should require a specific state, not adapt to anything it is thrown upon.
 a) It would make them a lot more complicated
 b) the actual state of the platform at any given moment could be hard to follow and could make it harder to debug/optimize the test cases
 
I think after a perfect rewrite, the keywords which change the state of the platform should do it explicitly, and do nothing else. If, let's say, we want to change an UEFI option, which requires a reboot to be applied - changing the option should be one keyword, and rebooting should be another one.

Would it be possible to write *short* test scenarios?
The amount of checks and worrying about the state could be overwhelming.

Would it be possible to write *device agnostic* scenarios?
Continuing the UEFI options example:
- Changing a UEFI option on a device with serial access requires a reboot, then setting it, and then rebooting again.
- On a SSH device, there is no need for a reboot initially, but the device has to be flashed, and consequently rebooted during flashing.
So that might not be possible

#### Defined layers of abstraction?

So maybe we could create two layers of abstraction on how the device is interacted with and a way to easily distinguish between them? Maybe a naming convention?

- low-level keywords, which work on specific platforms, in specific scenarios, require specific starting conditions and always leave the device in one, documented state.
- high-level keywords, don't care at all on which device are called. Not everything can be done using them. It won't necessarily by done optimally, but it should work.

Calling a high-level keyword after a low-level one would work every time, but not the other way around. A set of keywords for checking the state of the platform would need to be created for every way a state could change. The low-level keywords would define what state they require and  it would be up to the caller to make sure it's valid.

Calling only low-level keywords would not require checking the state, as it should be well known just by the sequence of calls of the keywords.

Calling only high-level keywords would not require checking the state, as it doesn't matter, the keywords will adapt.

It seems it doesn't vary that much from what we already have in OSFV, it's just that it is not made clear which keywords ale "high-level" and which are not. And most of the keywords are somewhere in between, so they will only adapt to some deviations from the expected state. Like install a required package if it is not present, but not login to Ubuntu, if the device is not logged in.

lib/terminal.robot
```diff
@@ -62,9 +106,27 @@ Read From Terminal
...
+    ...    ``${expected}``: ``string`` - The text up to which the terminal
+    ...    will be read
+    ...
+    ...    === Return Value ===
+    ...    ``string`` - All the text from the terminal buffer, from the last
+    ...    time it was cleared, up to the moment ``${expected}`` is found, or
+    ...    the timeout passes
+    ...
+    ...    === Effects ===
+    ...    The terminal buffer is read and consequently cleared
```
m-iwanicki: Is buffer cleared in this case? I think it works like FIFO buffer it just discards what you read so in this case it only clears up to the first occurrence of `${expected}`

lib/terminal.robot
```diff
@@ -166,7 +301,25 @@ Execute Command In Terminal
     RETURN    ${output}
 
 Execute UEFI Shell Command
-    [Documentation]    Universal keyword to execute command in Shell.
+    [Documentation]
+    ...    Universal keyword to execute command in Shell.
+    ...
+    ...    === Requirements ===
```
m-iwanicki, line 309: Should be run inside EFI shell to make sense. Keyword itself would probably succeed anywhere, not that it would do anything sensible.

lib/terminal.robot
```diff
@@ -166,7 +301,25 @@ Execute Command In Terminal
...
+    ...
+    ...    === Arguments ===
+    ...    - ``${command}``: ``string`` - The command to execute
+    ...    - ``${timeout}``: ``string`` = ``30s`` - The DUT Response Timeout for
+    ...    \ executing the command
+    ...    - ``${uefi_shell_input_latency}``: ``string`` - Time in Robot
+    ...    Framework format
+    ...
+    ...    === Return Value ===
+    ...    None
```
m-iwanicki: As of now it returns output of used command (after one second of sleep time)

lib/terminal.robot
```diff
@@ -166,7 +301,25 @@ Execute Command In Terminal
...
+    ...
+    ...    === Effects ===
+    ...    The ``${command}`` is written to the terminal and the keyword waits
+    ...    until the execution ends or ``${timeout}`` passes.
```
m-iwanicki, line 328: Currently keyword waits 1 second to read commands output. Timeout is used I think to make sure whole command was written before using it (pressing enter) as some shells could be quite slow.

lib/terminal.robot
philipanda: Yes
https://github.com/Dasharo/open-source-firmware-validation/compare/85b571ac5ad84867cb35000a8a0268a9671c8cce..95010bdf4a944e2bfad1a1540aa5f78f61a18507#diff-01b85c8135c6f83b72248ce44ded097ca7bf1698f59edef65b09dd6d385848deL129-R127

lib/terminal.robot
philipanda: I think it would work fine, just with some additional delays
https://github.com/Dasharo/open-source-firmware-validation/compare/85b571ac5ad84867cb35000a8a0268a9671c8cce..95010bdf4a944e2bfad1a1540aa5f78f61a18507#diff-01b85c8135c6f83b72248ce44ded097ca7bf1698f59edef65b09dd6d385848deL308-R311

lib/terminal.robot
philipanda: `Execute Command In Terminal` was also missing the return value.

https://github.com/Dasharo/open-source-firmware-validation/compare/85b571ac5ad84867cb35000a8a0268a9671c8cce..95010bdf4a944e2bfad1a1540aa5f78f61a18507#diff-01b85c8135c6f83b72248ce44ded097ca7bf1698f59edef65b09dd6d385848deL318-R324

lib/terminal.robot
philipanda: It also waits up to ${timeout} for the prompt to appear, as the output is read using `Read From Terminal Until Prompt`

philipanda: But only after rebasing :)
https://github.com/Dasharo/open-source-firmware-validation/compare/85b571ac5ad84867cb35000a8a0268a9671c8cce..95010bdf4a944e2bfad1a1540aa5f78f61a18507#diff-01b85c8135c6f83b72248ce44ded097ca7bf1698f59edef65b09dd6d385848deL330-R336

m-iwanicki: I would also change this `timeout` to `response_timeout` to not use the same variable later which can confuse someone (I missed that it's being assigned different value to prior to sleep)

philipanda: https://github.com/Dasharo/open-source-firmware-validation/compare/95010bd..58c43

EduKav1813 (approved):

lib/bios/menus.robot
```diff
@@ -371,7 +735,7 @@ Set Option State
         RETURN    ${FALSE}
     END
 
-Try To Insert Non-numeric Values Into Numeric Option
+Try To Insert Non-numeric Values Into Numeric Option    # TODO should be moved? It's a test case helper, not a library kwd
```
EduKav1813: IMO we should move this, but in another PR. Might document it here, as it is within the scope of this PR

lib/bios/menus.robot
```diff
@@ -454,8 +454,24 @@ Enter Dasharo APU Configuration
     ...    Dasharo APU Configuration
     RETURN    ${apu_menu}
 
-Enter Dasharo Submenu    # TODO remove: redundant keyword, only used to change the default parameter
+Enter Dasharo Submenu    # TODO redundant keyword, only used to change the default parameter
+    [Documentation]
+    ...    Enters given Dasharo submenu and returns construction
+    ...
+    ...    === Requirements ===
+    ...    - To be in the Dasharo setup menu
```
EduKav1813, line 446: nit: to be consistent, can rewrite as `Must be called from a Dasharo setup menu`

lib/bios/menus.robot
```diff
@@ -386,26 +386,6 @@ Enter Submenu From Snapshot And Return Construction
...
-    ...    === Return Value ===
-    ...    None
-    ...
-    ...    === Side Effects ===
-    ...    - The performed changes will be saved and applied.
-    ...    \ Most options require a reboot to take effect. Some of them don't.
-
-    Press Key N Times    1    ${F10}
-    Write Bare Into Terminal    y
-
```
EduKav1813: Do we use this keyword in tests? Would be good to check and replace all occurrences with a successor keyword

EduKav1813 (approved):

EduKav1813:
Note: due to massive number of changes, I did the review commit-by-commit, so some of my comments or requested changes may be implemented by older commits.

lib/bios/menus.robot
EduKav1813: Implemented by older commit 👍

lib/bios/menus.robot
philipanda: It was an unused duplicate of `menus.Save Changes`

lib/bios/menus.robot
philipanda: It was moved in [8e6651a](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-sandbox/workbench/workbench.html)
If it is only used in one test case then there is no need for longer documentation

PR #485, opened 2024-09-03
Add submodule

wiktormowinski:

PR #665, opened 2025-01-14
pikvm: use package instead of submodule

macpijan:
To be used with: https://github.com/3mdeb/pikvm-rest-api/pull/19

This is a prerequisite to continue with: https://github.com/Dasharo/open-source-firmware-validation/pull/485

The most basic test works: `dasharo-compatibility/custom-boot-menu-key.robot`. 

Still need to check and confirm other uses of PiKVM keywords, and update them if necessary (although care has been taken in pikmv-rest-api changes so that no changes in tests should be necessary).

Submodule has been removed and Python package is used instead in requiremts.txt, as this allows to import Library from class and initialize it just once with static values (such as credentials PIKVM_IP).

### DIFF
--- a/dasharo-compatibility/custom-boot-menu-key.robot
+++ b/dasharo-compatibility/custom-boot-menu-key.robot
@@ -11,7 +11,6 @@ Library             RequestsLibrary
 Resource            ../variables.robot
 Resource            ../keywords.robot
 Resource            ../keys.robot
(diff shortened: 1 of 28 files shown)


macpijan:
Tried with ESP and Secure Boot suites on MSI, where mount operations are also used (in addition to just keyboard, which is used in all tests).

I have looked through robot files and cannot see more keywords needing update with regard to this change.

Issue #606, opened 2024-11-21
Upgrade to RF7


macpijan:
- Possibly resolving limitation mentioned in: https://github.com/Dasharo/open-source-firmware-validation/pull/501#discussion_r1852073048

pekkaklarck:
Upgrading at least to RF 6.1 gives you Python 3.12 compatibility. Without it the `$var` syntax doesn't work:
https://github.com/robotframework/robotframework/issues/4771

Upgrading to RF 7.1 gives official Python 3.13 support, but we didn't actually need to make any code changes to support it so also earlier versions are compatible:
https://github.com/robotframework/robotframework/issues/5091

macpijan:
There is a new thread regarding switch to this lib, that it must be updated to telnetlib3, as telnetlib got removed with Python 3.13: https://docs.python.org/3/library/telnetlib.html

SergiiDmytruk:
Not exactly new, it was mentioned in some relevant issues.  `exscript` module is another option and looks more like `telnetlib`, because `telnetlib3` seems to be asynchronous which wouldn't be used in RF anyway (I assume there is a way to force async functions to be synchronous if needed).

macpijan:
Thanks for this useful background

DaniilKl:
@macpijan, what is the status here? Zarhus Team needs OSFV to use RF version 7.x. here:

* https://github.com/Dasharo/open-source-firmware-validation/pull/863
* https://github.com/Dasharo/open-source-firmware-validation/issues/655

For now we are considering a separate branch for DTS E2E tests, that will use RF 7.x. This can cause problems with rebasing later.

macpijan:
No activity here means no status update.

There was no activity around this as this was not strictly necessary so far.

DaniilKl:
Please, consider scheduling the update.

PR #673, opened 2025-01-16
Fix Inter Suite Cross Contamination

wiktormowinski:
Some tests behave differently when launching them inside whole suite vs individually. This fact suggest that after one suite is complete (by returning either Pass Fail or Skip) the environment doesn't return to it's original state. The most common (but not the only one) manifestation of this issue is telnet related error:  "EOFError: telnet connection closed"

This MR aims to find these problematic spots and in effect, bring the regression and individually run tests as close in result as possible.

### DIFF
--- a/dasharo-compatibility/apu-configuration-menu.robot
+++ b/dasharo-compatibility/apu-configuration-menu.robot
@@ -18,6 +18,8 @@ Suite Setup         Run Keywords
 ...                     Prepare Test Suite    AND
 ...                     Skip If    not ${APU_CONFIGURATION_MENU_SUPPORT}    APU configuration tests not supported.
 Suite Teardown      Run Keywords
+...                     Log Out And Close Connection
+...                     AND
 ...                     Skip If    '${SUITE_STATUS}' == 'SKIP'    Skipping Teardown since Suite was skipped as well
 ...                     AND
 ...                     Flash Firmware    ${FW_FILE}
--- a/dasharo-compatibility/apu-configuration-menu.robot
+++ b/dasharo-compatibility/apu-configuration-menu.robot
@@ -17,14 +17,8 @@ Resource            ../lib/linux.robot
 Suite Setup         Run Keywords
 ...                     Prepare Test Suite    AND
 ...                     Skip If    not ${APU_CONFIGURATION_MENU_SUPPORT}    APU configuration tests not supported.
-Suite Teardown      Run Keywords
-...                     Log Out And Close Connection
-...                     AND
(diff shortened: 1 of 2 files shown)


wiktormowinski:
The change in APU suite teardown we did a while ago was a good addition and served its purpose well. However, it seems like it is causing a new problem to tests following.
image
_Figure. 1 The affected area: suites between APU and CBK (being the first suite where it starts to PASS)_

After trying a few fixes i've finally found the culprit. The Skip not only skips the obsolete reflash it also gets rid of `Log Out And Close Connection`. The absence of this keyword is connected to the weird behavior in the next few suites. So after forcibly adding the missing kwd before `Skip If` the results are looking good:
image
_Figure. 2 The affected area: suites between APU and CBK_
NOTE- the few audio and cpu tests failed because of different issues (for example missing config contents)

The experiment was a success. The only problem now is how do we handle this issue @macpijan ? As far as I know, we can't put condition in teardown directly, that's why these commits: https://github.com/Dasharo/open-source-firmware-validation/pull/673/commits/8474e5d9f947ff76bf14539109da34d6488bbae1 https://github.com/Dasharo/open-source-firmware-validation/pull/673/commits/dcb979cd114289194265d4b00de5ca87643e3c67 created a kwd just for this occasion (the name is wip)

Here are the logs: 
[logs-fixes.zip](https://github.com/user-attachments/files/18443306/logs-fixes.zip)

philipanda:
> As far as I know, we can't put condition in teardown directly

There is the keyword "Run Keyword If" which allows to run another keyword if a condition is true. It should be possible to run it from "Run Keyword" in Suite Setup / Suite Teardown

SebastianCzapla:
Probably the best solution here is to change contributing guidelines, and enforce suite init/teardown as local keywords:
image
image

It has multiple benefits, clear skip conditions, we can install required packages as part of the initialization, we can set dut into any state we need, and we can clearly unmake the changes, and return to the default state DUT

philipanda:
I think that creating local keywords for setup and teardown won't allow to achieve more than using `Run Keywords` for the procedures, like it is now. What will it help with is making complex setup & teardown procedures easier to create and read.

I am not sure about enforcing defining those procedures as keywords.
- Complex setup/teardown is not needed in most test suites.
- All keywords in robot framework are of global scope. You can add the tag `robot:private`, but all that achieves is throwing a warning if the keyword is used.  Even the `libdoc` does not care about the tag (as of version 5.0 we are using, the latest version won't document them https://github.com/Dasharo/open-source-firmware-validation/issues/606)

wiktormowinski:
related: https://github.com/Dasharo/open-source-firmware-validation/issues/930

PR #684, opened 2025-01-27
self-tests/terminal.robot: Remove failing keyword

SebastianCzapla:


philipanda:
Can you explain why does this keyword fail?
`Execute UEFI Shell Command` shouldn't ever finish if the prompt is not set. The keyword waits until the prompt appears on the terminal, which means the command was executed.

PR #707, opened 2025-03-05
Protectli vp2420 1.2.1 rc6

matmacieje:
Platform-specific changes from v1.2.1-rc6 testing on vp2420.


keywords.robot
```diff
@@ -1545,3 +1545,7 @@ Should Contain All
     FOR    ${substring}    IN    @{substrings}
         Should Contain    ${string}    ${substring}
     END
+
```
filipleple: @matmacieje  AFAIK, we can't have `Power On` = `Power On Default` in our main keywords file, it's gonna break things for some other platforms. 

The approach is that the Power On keyword is defined in the generic `platform-configs/include/*-common.robot` files for each set of platforms, which are included in the model-specific configs that inherit from them.

If you were getting an error about Power On being ambiguous/missing, something went wrong in the chain of includes from `protectli-common` through `protectli-pro` down to the `vp2430` config -- either something's missing, or the keyword occurs twice.

So the right fix should probably only happen in the `platform-configs` folder, please rework this one

filipleple (changes requested):

filipleple:
Also, we generally like to satisfy pre-commit before marking things as ready for review. If your workstation setup went according to plan, it should work like:

```
pre-commit install
pre-commit run --all-files
```
called from the OSFV repo.

lib/terminal.robot
```diff
@@ -256,9 +256,9 @@ Write Bare Into Terminal
     ...
     ...    === Effects ===
     ...    The ``${text}`` is written to the terminal
-    [Arguments]    ${text}
+    [Arguments]    ${text}    ${interval}=None
```
miczyg1, line 259: Is the interval still needed somewhere? I recall only PC Engines apu platforms with SeaBIOS needed that in iPXE.

matmacieje: My reason: dasharo-compatibility/network-boot-utilities.robot, NBT005.001 and NBT007.001; without interval, tests are failing on partial command input. Tested with Write Bare from official robot framework Telnet library.

keywords.robot
matmacieje: Done in https://github.com/Dasharo/open-source-firmware-validation/pull/707/commits/58f57b7fdfe4fdb656fd5f58793a1515594f4b0a

filipleple:
@matmacieje by the ~30 extra changes it looks like something didn't work out when rebasing onto develop.

PR #724, opened 2025-03-13
options-lib_dcu: Boot System...: compare same case entry names

philipanda:
The menus/dcu makes the bootentry lowercase, but the default implementation (via serial) does not. This causes the bootentry name in configs only compatible with one of them (dcu wants it to be lowercase, the other one requires it to take letter case into account). This change should make the dcu implementation find the bootentries even if they have upperace letters in them.

### DIFF
--- a/lib/options/options-lib_dcu.robot
+++ b/lib/options/options-lib_dcu.robot
@@ -142,6 +142,7 @@ Set Nextboot
 
     ${os_boot_id}=    Set Variable    ${EMPTY}
     ${os_bootentry_name}=    Get From Dictionary    ${ENV_ID_OS_BOOTMENU_NAMES}    ${env_id}
+    ${os_bootentry_name}=    Convert To Lower Case    ${os_bootentry_name}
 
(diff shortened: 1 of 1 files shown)


os-config/environment-test-ids.py
```diff
@@ -20,6 +20,6 @@
 ENV_ID_OS_BOOTMENU_NAMES = {
     ENV_ID_UBUNTU: "ubuntu",
     ENV_ID_FEDORA: "fedora",
-    ENV_ID_WINDOWS_11: "Windows Boot",
+    ENV_ID_WINDOWS_11: "Windows",
```
SebastianCzapla: Stale ID, should be ENV_ID_WINDOWS

SebastianCzapla (changes requested):

os-config/environment-test-ids.py
philipanda: The PR is behind the current develop. Needs a rebase

SebastianCzapla (approved):

PR #727, opened 2025-03-13
Vp3210 config

miczyg1:
The VP3230 uses 90W (12V 7.5A) PSU, which requires sonoff, but VP3210 uses 60W (12V 5A) PSU which is still suitable for RTE relay power control.

Also add missing information about VP3210 based on lspci and lscpu output from DTS.


miczyg1:
@macpijan I forgot to push a small fix for CMOS clear test to the last PR (robot tries to match the quotes)

miczyg1:
@macpijan PR is ready for review

macpijan (approved):

PR #728, opened 2025-03-14
Naming convention comment old names

philipanda:


krystian-hebel (changes requested):
Still needs to be added:
- ETHPERF001.001 and ETHPERF002.001 - they have new IDs now
- DSP002.*:
  - missing old IDs in the comments
  - not sorted alphabetically, might as well fix it now

philipanda:
The old IDs as comments were added initally: https://github.com/Dasharo/open-source-firmware-validation/compare/8fd975c7914e73ad90475a1f653bd1ae165ce2e9%5E..8fd975c7914e73ad90475a1f653bd1ae165ce2e9#diff-a547482065bb927cc5e537616259f90585e4426aff5f0c18b480419e813728c1L46
Do you think they are not visible enough? When adding them in a new line robocop complains that there must be an empty line between the comment and the keyword. Maybe it would be best to add them to the keyword documentation, not as a comment then?

dasharo-compatibility/display-ports-and-lcd-support.robot
```diff
@@ -52,7 +52,16 @@ DSP002.201 - External HDMI display in OS (Ubuntu)
     Skip If    "${ENV_ID_UBUNTU}" not in "${TESTED_LINUX_DISTROS}"
     DSP002 - External HDMI Display    ${ENV_ID_UBUNTU}
 
-DSP002.301 - External HDMI display in OS (Windows)
+DSP002.202 - External HDMI display in OS (Fedora)
+    [Documentation]    Check whether an external HDMI display is visible in
+    ...    Fedora OS. An external HDMI display must be provided in
+    ...    the platform config.
+    Skip If    not ${EXTERNAL_HDMI_DISPLAY_SUPPORT}    DSP002.002 not supported
+    Skip If    not ${TESTS_IN_UBUNTU_SUPPORT}    DSP002.002 not supported
```
krystian-hebel: These report old IDs, should be updated. Same for two other test cases, which GH won't allow me to comment on.

krystian-hebel (changes requested):
> Do you think they are not visible enough?

Sorry for not being clear, I meant that those `.90x` IDs were not added, but after a second thought, those were added and removed in the same PR, so it is unlikely that they would be used in any testing. It is good as it is for DSP, but old ETHPERF IDs are still missing.

> Maybe it would be best to add them to the keyword documentation, not as a comment then?

This may be a good idea, that way it would be propagated to the documentation site as well. Let's also define and add the format of such info to the documentation. I'm thinking about something like `Previous IDs: OLDER001.001, OLD001.001`, with each newer added at the end (so the order of changes is easy to follow without having to search in git history), but if you have other ideas, I'm open to suggestions.

dasharo-compatibility/display-ports-and-lcd-support.robot
philipanda: Updating the keywords: https://github.com/Dasharo/open-source-firmware-validation/compare/f43737703314c0863d12c8b0473dcfcf6206ec9d..c04cc022d6b24d724fb91c0cbfc284a151d8bea4

Updating the docs: https://github.com/Dasharo/open-source-firmware-validation/compare/c04cc022d6b24d724fb91c0cbfc284a151d8bea4..38ace08e7525c63875ef2f729c27f9f7043e7a4e

macpijan: @krystian-hebel any more comments?

krystian-hebel: Sorry, I missed it.

krystian-hebel (dismissed):

philipanda:
@krystian-hebel rebased

dasharo-compatibility/display-ports-and-lcd-support.robot
```diff
@@ -52,25 +52,25 @@ DSP002.201 - External HDMI display in OS (Ubuntu)
     [Documentation]    Check whether an external HDMI display is visible in
     ...    Linux OS. An external HDMI display must be provided in
     ...    the platform config.
-    ...    Previous IDs: DSP002.002
+    ...    Previous IDs: DSP002.001
```
krystian-hebel, line 55: Where does this change come from? From what I can see in the database, it used to be .002, .001 is `Internal LCD in firmware`.

dasharo-compatibility/display-ports-and-lcd-support.robot
```diff
@@ -52,25 +52,25 @@ DSP002.201 - External HDMI display in OS (Ubuntu)
...
+    Skip If    not ${EXTERNAL_HDMI_DISPLAY_SUPPORT}    DSP002.202 not supported
+    Skip If    "${ENV_ID_FEDORA}" not in "${TESTED_LINUX_DISTROS}"    DSP002.202 not supported
     External HDMI Display    ${ENV_ID_FEDORA}
 
 DSP002.301 - External HDMI display in OS (Windows)
     [Documentation]    Check whether an external HDMI display is visible in
     ...    Windows OS. An external HDMI display must be provided in
     ...    the platform config.
-    ...    Previous IDs: DSP002.003
+    ...    Previous IDs: DSP002.002
```
krystian-hebel, line 73: Where does this change come from? From what I can see in the database, it used to be .003.

dasharo-compatibility/display-ports-and-lcd-support.robot
philipanda: Sorry for changing that during a rebase instead of a separate commit.

I thought that was an error. Looking at the old Test & Feature matrix, as well as the [osfv-results repo](https://github.com/Dasharo/osfv-results/blob/main/boards/NovaCustom/MTL_14th_Gen/V560TNX/v0.9.1-results.csv#L143) and[ the docs ](https://github.com/dasharo/docs/commit/471e58e) it seems the previous one was wrong

dasharo-compatibility/display-ports-and-lcd-support.robot
philipanda: Same as DSP002.201

dasharo-compatibility/display-ports-and-lcd-support.robot
krystian-hebel: Then it would mean that we have an error in the database. It was imported that way from old spreadsheet, I wonder at which point it changed :thinking: @macpijan CC, this needs verification.

krystian-hebel: @macpijan ping, this blocks merging, which blocks few other tasks.

macpijan: Conclusion after off-line discussion: this looks fine, it has been confirmed in DB and against old spreadsheet.

krystian-hebel (approved):

PR #737, opened 2025-03-21
dts/dts-e2e.robot: Use new DPP credentials

PLangowski:


dts/dts-e2e.robot
```diff
@@ -514,7 +514,7 @@ E2E003.009 MSI PRO Z690-A DDR-4 update (Coreboot + UEFI -> Coreboot + UEFI) - DP
     Provide DPP Credentials
 
     # 4) Start update:
-    Go Through Update
+    Go Through Update Decline Heads
```
DaniilKl, line 517: Could you explain the purpose of this change in commit message? Because I am missing context here.

dts/dts-e2e.robot
```diff
@@ -1162,6 +1159,19 @@ Go Through Update
     Wait For Checkpoint And Write    ${DTS_SPECIFICATION_WARN}    Y
     Wait For Checkpoint And Write    ${DTS_DEPLOY_WARN}    Y
 
+Go Through Update Decline Heads
+    [Documentation]    This KW goes through standard Dasharo update workflow
+    ...    choosing all needed menu options and answering all questions. It
```
DaniilKl, line 1164: Maybe extend `Go Through Update` KW with `Heads`, `DPP UEFI`, `Seabios` and `UEFI` options, since after migration to MinIO DTS do not choose update version automatically?

DaniilKl (changes requested):

dts/dts-e2e.robot
PLangowski: The problem with this is that these cases have different workflows. When deploying seabios, it doesn't matter if we have another type available - we always pick the seabios option. When deploying UEFI, the workflow is different depending on whether we have Heads access or not. If we do, we have to press `N` so that DTS doesnt perform transition. It's going to require some work to unify this, so maybe we should push this for another release (create an issue).

dts/dts-e2e.robot
PLangowski: https://github.com/Dasharo/open-source-firmware-validation/pull/737/commits/8a79950fdefca265aeab86157335eaf659654b60

dts/dts-e2e.robot
DaniilKl: So, during update, DTS still is choosing to which version to update automatically. But the Heads transition should be chosen manually?

dts/dts-e2e.robot
DaniilKl: > When trying to deploy coreboot + UEFI, if Heads fw is available, DTS
> will ask the user if they would like to switch to Heads. The user has to
> decline in order to install UEFI.

Why we could not extend already existing `Go Through Update` KW with argument Heads? That will make sure we are making a transition to Heads, but without the argument it will not make a transition.

dts/dts-e2e.robot
PLangowski: yes, fw switching is a manual choice

DaniilKl (approved):

PR #744, opened 2025-03-26
Add acpi driver testing

JakubRed:


dasharo-compatibility/acpi-driver.robot
```diff
@@ -0,0 +1,59 @@
...
+...                     AND
+...                     Import Variables    ${CURDIR}/../platform-configs/${SENSORS_CONFIG_FILE}
+Suite Teardown      Run Keyword
+...                     Log Out And Close Connection
+
+*** Variables ***
+@{EXPECTED_OUTPUT}=    dasharo_acpi-acpi-0
+...    Adapter: ACPI interface
+...    CPU 0:
+...    GPU 0:
```
mkopec: Not all systems have a discrete GPU, so this pattern won't match on iGPU only models. Let's remove GPU from the list, we can only safely expect that the fist 2 lines will appear
```
dasharo_acpi-acpi-0
Adapter: ACPI interface
```

That should be enough to prove that the driver has loaded and enumerated the ACPI device.

JakubRed: https://github.com/Dasharo/open-source-firmware-validation/pull/744/commits/521edd951e7456649e797ce7223bbf7e58859350

JakubRed:
There is currently no implemented package installation in Robot for Fedora, the current solution is temporary.

mkopec (approved):

PR #751, opened 2025-03-31
lib/linux.robot: Remove TPM interrupt error from allowlist

miczyg1:


miczyg1:
Tested on VP2420. Th dmesg contains only these errors:

```
ata1.00: Read log 0x00 page 0x00 failed, Emask 0x4
ath10k_pci 0000:05:00.0: wmi unified ready event not received
ath10k_pci 0000:05:00.0: could not init core (-110)
ath10k_pci 0000:05:00.0: could not probe fw (-110)
```

The TPM interrupt should generally be checked against and fixed in firmware.

macpijan:
@mkopec Hasn't it originated from laptops?
Is it still there?

miczyg1:
> @mkopec Hasn't it originated from laptops? Is it still there?

It was present not only on laptops.

krystian-hebel (approved):

PR #752, opened 2025-04-02
Add new test case: TPM003.101 Change active PCR banks with TPM PPI (EDK2 UEFI)

JakubRed:


lib/tpm.robot
```diff
@@ -104,3 +104,25 @@ Verify Presence Of Any PCRs Via Sysfs
     [Documentation]    Check sysfs interface for presence of any PCR
     ${pcr_state}=    Execute Command In Terminal    ls /sys/class/tpm/tpm0/pcr-sha* &>/dev/null && echo "Found PCRs"
     Should Contain    ${pcr_state}    Found PCRs
+
+Check TPM2 Banks State After FW Changes
```
krystian-hebel: This should be in `lib/tpm2.robot` instead.

lib/tpm.robot
```diff
@@ -104,3 +104,25 @@ Verify Presence Of Any PCRs Via Sysfs
...
+    [Arguments]    ${sha1_desired}    ${sha256_desired}
+    Save Changes And Reset
+    Read From Terminal Until    Press F12 to change the boot measurements to use PCR bank(s) of the TPM
+    Press Key N Times    1    ${F12}
+    Prepare TPM Test On Ubuntu
+    ${sha1}    ${sha256}=    Check Which TPM2 Banks Are Enabled
+    Should Be Equal    ${sha1}    ${sha1_desired}
+    Should Be Equal    ${sha256}    ${sha256_desired}
+
+Enter The TCG2 Configuration Menu
```
krystian-hebel: For TPM1.2, the menu is called `TCG Configuration` It has some similar options, and there will never be both menus at the same time. Is it possible to use regex as argument to `Enter Submenu From Snapshot` below? That way the keyword could be used for both, otherwise it should be moved to `lib/tpm2.robot`.

krystian-hebel (changes requested):

lib/tpm.robot
JakubRed: 5b099576e82a143714c20830dd87e7d0a913ae8a

lib/tpm.robot
JakubRed: 5b099576e82a143714c20830dd87e7d0a913ae8a

krystian-hebel: Please also drop `2` from keyword name and update the documentation to mention that it works for both of those. There is also a commented out old version, remove it.

dasharo-security/tpm-support.robot
```diff
@@ -122,12 +122,43 @@ TPM003.003 Check TPM Physical Presence Interface (Windows)
...
-#    Skip If    not ${TPM_SUPPORTED_VERSION} == None    TPM003.004 not supported
-#    Skip If    not ${TESTS_IN_UBUNTU_SUPPORT}    TPM003.004 not supported
-# TODO: https://docs.dasharo.com/unified-test-documentation/dasharo-security/200-tpm-support/#tpm003004-change-active-pcr-banks-with-tpm-ppi-firmware
+TPM003.101 Change active PCR banks with TPM PPI (EDK2 UEFI)
+    [Documentation]    This test aims to verify that the TPM Physical Presence
+    ...    Interface is working properly in the firmware by changing active TPM PCR banks.
+    Skip If    not ${TPM_SUPPORTED_VERSION} == 2    TPM003.101 not supported
+    Skip If    not ${TESTS_IN_UBUNTU_SUPPORT}    TPM003.101 not supported
+    Prepare TPM Test On Ubuntu
+    Execute Reboot Command
```
krystian-hebel: Change both to `Power On`.

krystian-hebel (changes requested):

lib/tpm.robot
JakubRed: https://github.com/Dasharo/open-source-firmware-validation/pull/752/commits/de91ae8e75fc1175185e1fd8614ebc945cf70a1c

dasharo-security/tpm-support.robot
JakubRed: https://github.com/Dasharo/open-source-firmware-validation/pull/752/commits/3068c10178673745d1cbcb45f092950185c4b4d7

lib/tpm.robot
krystian-hebel: Keyword documentation still mentions TCG2 only.

JakubRed: c830d9e9ec82c71d7e8a78125a9779248b423351 - I already squashed some commits

krystian-hebel (approved):

PR #760, opened 2025-04-04
osfv: Change invalid ENV_IDs

SebastianCzapla:


keywords.robot
```diff
@@ -71,7 +71,7 @@ Login To Linux Via OBMC
 
 Login To Windows
     [Documentation]    Universal login to Windows.
-    Boot System Or From Connected Disk    ${ENV_ID_WINDOWS}
```
philipanda: I must've made a mistake when changing the variable name and hadn't changed it in os-configs or something. We prefer to not distinguish between Windows versions. May I ask you to fix this the other way around? So that `${ENV_ID_WINDOWS}` is used consequently instead of `${ENV_ID_WINDOWS_11}`.

philipanda (changes requested):

keywords.robot
SebastianCzapla: changed in https://github.com/Dasharo/open-source-firmware-validation/pull/760/commits/1db407a4081cc8b5cb9ed61ed7816013e519169d

philipanda (approved):

PR #776, opened 2025-04-09
Create more detailed naming convetion of the result test dirs

JakubRed:
- In this PR I enchanced naming so now for example the command:
DIR_PREFIX=**release-testing** RTE_IP=0.0.0.0 DEVICE_IP=192.168.4.184 SNIPEIT_NO=True CONFIG=novacustom-v540tu ./scripts/run.sh dasharo-compatibility/**dcu**.robot
will result of test result path: logs/novacustom-v540tu/**release-testing**_**dcu**_2025_04_09_12_32_30/dcu_debug.log.
- The variable `DIR_PREFIX` is optional.
- In case that just one test suite is being run then the directory and logs will have the name of the test suite file. 
- If the whole test module is  being run then the directory and logs will have the name of that test module.
- Also this PR removes creating subfolder with module name - I have reasons to suspect that was never the intended outcome.
- I removed `.robot` as a part of logs names.
- I aslo protected this solution from some edge cases (that I could think of).

### DIFF
--- a/scripts/lib/robot.sh
+++ b/scripts/lib/robot.sh
@@ -86,6 +86,13 @@ execute_robot() {
   # Check if the required environment variables are set
   check_env_variable "CONFIG"
 
+  # DIR_PREFIX (optional) additional description of test result dir
+  if [ -n "${DIR_PREFIX}" ]; then
+    dir_prefix="${DIR_PREFIX}_"
+  else
+    dir_prefix=""
+  fi
+
   # RTE_IP environment variable is not required for some platforms
   if [ -n "${RTE_IP}" ]; then
     rte_ip_option="-v rte_ip:${RTE_IP}"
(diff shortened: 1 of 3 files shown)


scripts/lib/robot.sh
```diff
@@ -138,12 +145,24 @@ execute_robot() {
   # Thanks to detecting spacebars in arguments before _robot_args can now
   # safely be concatenated into a string and these arguments will still be
   # passed correctly.
+  #
+  # Firstly, the provided argument will be parsed to get the propre name
```
matmacieje: proper (typo)

matmacieje (changes requested):

scripts/lib/robot.sh
JakubRed: thx, fixed

miczyg1:
> Also this PR removes creating subfolder with module name - I have reasons to suspect that was never the intended outcome.

I think it was intended when you run regression.sh,  because it runs all tests, so having the logs divided into subfolders per suite was actually good.

JakubRed:
> I think it was intended when you run regression.sh, because it runs all tests, so having the logs divided into subfolders per suite was actually good.

@miczyg1 To my knowledge, right now regresion wraper creates one directory and puts all logs there without subfolders.
I pushed a change that keeps it that way.

scripts/lib/robot.sh
miczyg1: Now we have `proprer` insated of `proper`

JakubRed: fixed d0c5d7a754afbf3aa1f742597e89e6ed69681f1f

macpijan: wasn't this typo detected by pre-commit ?

miczyg1: Does it check the code too?

JakubRed: @macpijan It wasn't actually. The pre-commit had found another typo but not this one. IDK why.

JakubRed: @miczyg1 I think id does. It had found an unused variable which I had forgotten to remove.

miczyg1 (approved):

macpijan (approved):

PR #781, opened 2025-04-10
util/basic-platform-setup.robot: connection check in Run Ansible Play…

matmacieje:
…book

Added Check Internet Connection On Linux in Run Ansible Playbook keyword, to avoid SSH "host unreachable" error when network interface initialization takes a little longer.

### DIFF
--- a/util/basic-platform-setup.robot
+++ b/util/basic-platform-setup.robot
@@ -237,6 +237,7 @@ Run Ansible Playbooks
         # ansible will fail no matter the timeouts if host is unreachable
         # (not booted yet)
         Login To Linux
+        Check Internet Connection On Linux
 
         # Create temporary inventory file for given platform and OS
         ${inventory_file}=    Catenate    [host] \n


philipanda (approved):
I suppose that could happen on devices which use RTE as the main connection method.
Looks good.

I believe we could also use the `Get Hostname Ip` keyword used as a part of `Check Internet Connection On Linux` to get the IP device's IP address automatically when connected via RTE or PiKVM. That wouldn't work for laptops as the IP has to be known from the start, but shouldn't break anything.

miczyg1:
> I believe we could also use the `Get Hostname Ip` keyword used as a part of `Check Internet Connection On Linux` to get the IP device's IP address automatically when connected via RTE or PiKVM. 

That won't work with PiKVM, because OS login will be on graphical display instead of serial with PiKVM... Only fully functional RTE serial connection can use `Get Hostname Ip` to get an IP address if it is not known upfront (if you know the IP you can use SSH and not query for the IP).

miczyg1 (approved):

PR #785, opened 2025-04-11
Add ubuntu battery percentage check (kwd)

JakubRed:


SebastianCzapla:
Two major points here, first, keywords.robot should gain no new keywords, we use lib for all new keywords now, as per docs/contributing.md

Second point is about the KWD itself, it's too specific for a problem it solves. Some tests are shorter, and might need lower %, while longer tests might need like 50% to (RF might rerun a test if deviation is too high, increasing runtime)

Adding a param might be good solution here

mkopec (approved):

PR #788, opened 2025-04-11
Qemu lib refactor

### LINKED ISSUE #680: Refactor QEMU lib
Problems:

Errors from QemuMonitor are ignore, so many problems are ignored under the hood. We do now know when things fail in QemuMonitor. That was the result of false positive in some tests in the past, it seems.

Catching errors has been added here: b956e6fb9d60

But it results in stockpiling errors when handling disk / USB connection / disconnection.

A good suite to test this is `esp-scanning`.

The end result of this task should be that the esp-scanning and secure-boot suite can reliably pass on QEMU.

philipanda:


philipanda:
The changes work if VirtIO is supported in the firmware as shown [the CI runs](https://github.com/aronowski/coreboot/actions/runs/16078076279/job/45377893079) in https://github.com/Dasharo/coreboot/pull/704.

Once a QEMU release is published with the VirtIO support, ESP scanning will work again in the OSFV CI.

Issue #734, opened 2025-03-19
Out of bound access to config list


miczyg1:
### Device

VP3210

### RTE version

Any

### OSFV version

develop

### Affected component(s) or functionality

`Get Current CONFIG Stop Index` keyword

### Brief summary

robot is unable to find the config if given config is the last in the list

### How reproducible

Always

### How to reproduce

`robot -L TRACE -v rte_ip:192.168.10.211  -v config:protectli-vp3210 -v device_ip:192.168.10.21 -v fw_file:protectli_vp32xx_v0.9.0-rc10.rom dasharo-compatibility/usb-hid-and-msc-support.robot`

### Expected behavior

Suite setup doesn't fail.

### Actual behavior

Suite setup fails.

### Link to screenshots or logs

Logs attached

[log.zip](https://github.com/user-attachments/files/19342696/log.zip)

### Additional context

Problem is in `Get Current CONFIG Stop Index` keyword. The line `${index}=    Set Variable    ${start+1}`  causes the robot to not enter the FOR loop, because `${index}`is incremented to 122 (the exact size of `${config_list}`). So it will always fail to find config for the last platform in the `${config_list}`, whatever it would be.

miczyg1:
@macpijan the faulting line `${index}=    Set Variable    ${start+1}` is in the keywords.robot since the initial commit. SO I am not sure who was the author and what was the intent here. Can you help here?

wiktormowinski:
@miczyg1 looks like we got rid of this keyword entirely:

![Image](https://github.com/user-attachments/assets/36e7584e-5a7d-4584-8125-1fc72dd5b2a9)
_Fig 1. Fragment of log from current `develop`, issue not prevalent. Reason: problematic line removed._ 

Full log attached:

[no_reproduction.zip](https://github.com/user-attachments/files/19761195/no_reproduction.zip)

wiktormowinski:
i don't know if i understand it correctly but this discovery implies the `variables.robot` file is not used in the current osfv workflow???

miczyg1:
The keywords are still used. Just grepping `Get Current CONFIG` is enough. Although it is not used in Suite Setup anymore, it is still used by `Boot System Or From Connected Disk` (lib/bios/menus.robot). So IMO it may still fail. I mean the `Get Current CONFIG {List Param, Item, Item Param}`.

BTW we have the same keyword name `Boot System Or From Connected Disk` in:

- lib/options/options-lib_dcu.robot
[...]

PR #794, opened 2025-04-16
keywords.robot: fix

wiktormowinski:


miczyg1 (changes requested):
While it fixes out of bounds error, I am not sure if this is still working properly.

In the `Get Current CONFIG` we have:

`${config}=    Get Slice From List    ${config_list}    ${start}    ${stop}`

This should get a slice of the config that contains all components of given CONFIG for platform. However, if the slice contains only RTE, nothing is returned here.

image

So to fix this we would need:

`${config}=    Get Slice From List    ${config_list}    ${start}    ${stop+1}`

Apparently `Get Slice From List ` cannot extract a single slice if start = end. End has to be at least 1 index higher than start. Apparently, passing stop+1 works even if that results in the index higher than the list size, maybe this is how it was designed:

image

wiktormowinski:
@miczyg1 ok so something like [that](https://github.com/Dasharo/open-source-firmware-validation/pull/794/commits/0497e3eb5b940f0e049f2c3fe7020b2cf80bcdbc) should scoop the config now and not leave empty return

miczyg1 (approved):

miczyg1:
Fixes: https://github.com/Dasharo/open-source-firmware-validation/issues/734

PR #802, opened 2025-04-18
Fixes made while testing v540tu v1.0.0-rc2

JakubRed:


dasharo-performance/disk-io.robot
```diff
@@ -164,6 +171,7 @@ DIO008.201 Random Write Performance (Ubuntu) (Battery)
     Skip If    not ${TESTS_IN_UBUNTU_SUPPORT}
     Skip If    not ${LAPTOP_PLATFORM}    The Platform is not a Laptop
     Skip If    not ${BATTERY_PRESENT}    Battery not present
+    Skip If    ${AC_CONNECTED}    The platform is not connected to AC
```
SebastianCzapla, line 174: We could probably skip laptop and battery check at all. If no AC is connected, and there is no battery, platform won't boot, and we won't even reach this far.

os-config/environment-test-ids.py
```diff
@@ -18,7 +18,7 @@
 ENV_ID_WINDOWS = "301"
 
 ENV_ID_OS_BOOTMENU_NAMES = {
-    ENV_ID_UBUNTU: "ubuntu",
+    ENV_ID_UBUNTU: "Ubuntu",
```
SebastianCzapla, line 21: Why are we changing this? AFAIK this was working well, if there is a problem somewhere in tests, could it be fixed by changing letter case there?

JakubRed: @SebastianCzapla The optput that is compared to this ENV_ID is for example:
` Boot0003* Ubuntu (on SSDPR-PX700-02T-80) PciRoot(0x0)/Pci(0x6,0x0)/Pci(0x0,0x0)/NVMe`. The comparison keyword is case sensitive. I tried to use conversions but they had genereated different kinds of errors. 
So it helps in my case and if you use this veriable już as a OS indicator the content of it shouldnt be a problem.

dasharo-performance/disk-io.robot
JakubRed: Probably, but this checks are in similar tests and I didn't want to create differences between them.

mkopec (approved):

Issue #817, opened 2025-04-30
New TPM test cases IDs collide with dasharo.docs


philipanda:
### Device

-

### RTE version

-

### OSFV version

develop

### Affected component(s) or functionality

TPM tests
Automated results impossible to upload to osfv-dashboard

### Brief summary

Newly added TPM003 and TPM004 IDs are colliding with existing cases.
TPM011 TPM012 missing from test_cases.json
XCP-NG and ESXi test have invalid IDs

### How to reproduce

check the test suite file and dasharo.docs test documentation

### Expected behavior

Tests must have unique IDs compliant with the naming convention.
Every ID must be reflected in `test_cases.json`

### Actual behavior

One ID collides with OSFV and docs, the other one only with docs

### Link to screenshots or logs

no need

PR #818, opened 2025-04-30
tpm-support: Change colliding test case IDs

philipanda:
Might fix https://github.com/Dasharo/open-source-firmware-validation/issues/817

The IDs are the next available numbers after whats already defined in [docs.dasharo](https://docs.dasharo.com/unified-test-documentation/dasharo-security/200-tpm-support/#tpm007001-encrypt-and-decrypt-rootfs-partition-ubuntu).

The test matrices were not updated and should be if the PR was to be merged

### DIFF
--- a/dasharo-security/tpm-support.robot
+++ b/dasharo-security/tpm-support.robot
@@ -152,7 +152,7 @@ TPM003.301 Check TPM Physical Presence Interface (Windows)
     ${out}=    Execute Command In Terminal    tpmtool getdeviceinformation
     Should Contain    ${out}    PPI Version: 1.3
 
-TPM003.101 Change active PCR banks with TPM PPI (EDK2 UEFI)
+TPM011.101 Change active PCR banks with TPM PPI (EDK2 UEFI)
     [Documentation]    This test aims to verify that the TPM Physical Presence
     ...    Interface is working properly in the firmware by changing active TPM PCR banks.
     ...    Previous IDs: TPM003.004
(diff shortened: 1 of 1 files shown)


philipanda:
The suggested IDs were changed in a force-push, because docs.dasharo used the second ID segment of TPM005 test incorrectly, so its ID should be split into four: https://github.com/Dasharo/docs/pull/1054/commits/9777f22d0a1e85f6102b63d0fc4295ce91772402#diff-b922c3df1236bfba72e32adca5a1734900e0a227cfdc063e6b80a3842cf89cc0L618-L699

krystian-hebel (dismissed):
I don't understand why this is changed. Where exactly is the conflict? As far as I can tell, some of the IDs that were modified in OSFV weren't propagated to docs, but the tests themselves exist there. Why not just update the documentation?

philipanda:
> I don't understand why this is changed. Where exactly is the conflict? As far as I can tell, some of the IDs that were modified in OSFV weren't propagated to docs, but the tests themselves exist there. Why not just update the documentation?

@krystian-hebel
There is no reason to update the tests instead of the documentation other than the fact that the documentation took those IDs first. They weren't free to take in the first place. The overtaken tests are not present on osfv-dashboard nor the deprecated Dasharo Test & Feature Matrix, so probably changing the documentation instead is not a big deal.

philipanda:
The `TPM003.101` test is actually also colliding with some other tests in this file, the `TPM003.001`, `TPM003.201`, `TPM003.301`. The two tests do two entirely different things yet only differ in the second segment.

mkopec (approved):

PR #819, opened 2025-04-30
Add manual cpu performance win tests

JakubRed:


dasharo-performance/manual_cpu_performance.robot
```diff
@@ -0,0 +1,170 @@
...
+...                 AND
+...                 Pause Execution    It is advised to run this test via ssh
+...                 AND
+...                 Detect Or Install Phoronix Test Suite On Windows
+# TODO: human-readable representation of setup menu key for all platforms
+
+
+*** Variables ***
+@{TESTS}=               smallpt    crafty    cachebench    compress-7zip    blake2
+@{TESTS_SCORE}=         25.784    8913626    107819.3    22954    4.05    # avg or medium
```
mkopec: shouldn't the reference values be specified locally in board configs instead of globally in the test? You're gonna get wildly different results between an MSI Alder Lake desktop platform and something like a pc engines apu2

JakubRed: I refactored the code so it does not need platform specific input. Now result validation is based on the deviation parameter.

mkopec: so now the user has to get reference results and calculate the deviation manually? why not do it the same way as is done in other performance tests, with reference test results specified in board config?

mkopec (approved):

mkopec:
@JakubRed we have some conflicts here

dasharo-performance/manual_cpu_performance.robot
JakubRed: I removed the deviation completely.

PR #821, opened 2025-05-06
dasharo-compatibility/sata-detect.robot fix: wrong expected value

JakubRed:


dasharo-compatibility/sata-detect.robot
```diff
@@ -40,7 +40,7 @@ SAT001.002 SATA support in OS (Ubuntu)
         ${out}=    Execute Command In Terminal    sudo smartctl -i /dev/${disk}
         Log    ${out}
         ${sata_present}=    Run Keyword And Return Status    Should Contain    ${out}    SATA Version is:
-        Pass Execution If    '${sata_present}' == 'TRUE'    'SATA disk found, passing test'
+        Pass Execution If    '${sata_present}' == 'True'    'SATA disk found, passing test'
```
wiktormowinski, line 43: that's so weird that it checks string letter by letter, lets break this lineage by adding a logical condition: 
```  
Pass Execution If    ${sata_present}    SATA disk found, passing test 
```
just like in all those `Skip Ifs` all over the `osfv` tests , this argument will just look if `${sata_present}` a logical true

Worth noting: both my suggestion and yours work though so this is purely a cosmetic change i guess, what do you think @JakubRed ?

JakubRed: It is more readable as is so I would leave it without changes.

wiktormowinski (approved):

wiktormowinski:
[before.zip](https://github.com/user-attachments/files/20061062/before.zip)
[after.zip](https://github.com/user-attachments/files/20061066/after.zip)
some logs testing this fix

dasharo-compatibility/sata-detect.robot
macpijan: > that's so weird that it checks string letter by letter

I do not understand. How else to compare strings?

What is more weird here, that we cast bool to string and compare it with string literal. Rather than simply:

`Pass Execution If    ${sata_present}`

wiktormowinski: ye, like i said 

however, we reached conclusion that string comparison is more readable for test developers so lets leave it like that ig

macpijan: I disagree with this conclusion. It's less readable for me. It's very subjective, so it mustn't be an argument here.

This MR is a proof that this is simply unnecessary and creates possibility to make more mistakes. That is why, in the future, we should avoid it.

PR #835, opened 2025-05-13
platform-suspend-and-resume.robot: ensure ME is enabled for S0ix

mkopec:


macpijan:
@mkopec is it supposed to be in draft still?

@matmacieje has it been tested since then?

matmacieje:
@macpijan I don't think so.

PR #841, opened 2025-05-21
Osfv ci tests scope determining

philipanda:
## Short description
The script is supposed to help determine which tests should be run in order to verify if the changes performed to OSFV are breaking anything.
The script's functionality is tested using `scripts/ci/regression-scope/regression_scope_selftests.py`, which also serves a purpose of demonstrating how the rules files can be created.
## Related Issue(s)

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/other_changes.md
+++ b/.github/PULL_REQUEST_TEMPLATE/other_changes.md
@@ -1,0 +1,10 @@
+<!--
+SPDX-FileCopyrightText: 2024 3mdeb <contact@3mdeb.com>
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+<!-- markdownlint-disable MD041 -->
+## Short description
+
+## Related Issue(s)
--- a/.github/PULL_REQUEST_TEMPLATE/tests.md
+++ b/.github/PULL_REQUEST_TEMPLATE/tests.md
@@ -1,0 +1,23 @@
+<!--
+SPDX-FileCopyrightText: 2024 3mdeb <contact@3mdeb.com>
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+<!-- markdownlint-disable MD041 -->
+## Short description
+
+## Related Issue(s)
+
+## Checklist
+
+- [ ] When modifying `platform-configs/`, the minimal regression passes on all
+      of the affected devices.
+      (`scripts/regression.sh` -- -i minimal-regression`)
+    + [ ] <logs_device_1.zip>
+    + [ ] <logs_device_2.zip>
+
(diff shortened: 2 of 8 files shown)


filipleple:
This looks great, I'm just not sure if there isn't a chance that a single test file gets picked multiple times for the same PR? As in, if a large PR touches multiple files that are explicitly matched by one rule and implicitly used (via FILE_CONTAINS_MATCH) by another.

Then we’d get repeated execution of the same test suite, increasing CI time without benefit.

philipanda:
>This looks great, I'm just not sure if there isn't a chance that a single test file gets picked multiple times for the same PR? As in, if a large PR touches multiple files that are explicitly matched by one rule and implicitly used (via FILE_CONTAINS_MATCH) by another.
>
>Then we’d get repeated execution of the same test suite, increasing CI time without benefit.

Creating two rules that both suggest running one test suite **WILL** result in the same suite appearing two times in the output of either `filenames`, `commands`, or `robot_wrapper_args`.

Eg. after modifying `dasharo-performance/boot-time-measure.robot` and `lib/cbmem.robot`:
```
$ python scripts/ci/regression-scope/osfv_regression_scope.py commands
export RTE_IP=127.0.0.1 export FW_FILE=scripts/ci/qemu_q35.rom export CONFIG=qemu
scripts/run.sh dasharo-performance/boot-time-measure.robot -- -v snipeit:no
export RTE_IP=127.0.0.1 export FW_FILE=scripts/ci/qemu_q35.rom export CONFIG=qemu
scripts/run.sh dasharo-performance/boot-time-measure.robot dasharo-performance/fast-boot.robot -- -v snipeit:no
$
```

Two distinct rules can define that the suites matched by them should be run with arbitrary optional robot arguments and with arbitrary environment variables set.

For this reason I am not sure if preventing this could be done in a way smarter than just looking for duplicated sets of {ENV_VARS+ROBOT_ARGS+SUITE} in the sum of the outputs of all the parsed rules.

Nevertheless that is something worth optimizing.

philipanda:
@filipleple In the last few commits I've added what you've asked for.
If multiple rules would result in running a test case with the exact same environment variables and robot arguments, then it will not be executed more than once. If the environment variables or robot arguments differ, every unique requested combination will be run. There are two test cases to cover this https://github.com/Dasharo/open-source-firmware-validation/pull/841/files#diff-a92cab7ff5d4000dfd89849ecd23ee450590d996f0026ae4b38fc9fbb2d75e7fR342-R473

filipleple (approved):
Neat, LGTM now :fire:

macpijan (commented):
Changes look good, but maybe we should squash some of these commits prior merging?

philipanda:
@macpijan I've just squashed the commits, is that enough?

PR #850, opened 2025-05-29
test_cases.json: add new file for synchronization with database

krystian-hebel:
The file was obtained with the following command:

```
curl http://user:password@<DB-IP>:<DB-PORT>/test_cases/_all_docs?include_docs=true \
    | jq '[.rows[] | select(.doc._id | startswith("_design") | not) | del(.doc._rev) | pick(.doc)] | .[].doc' \
    > test_cases.json
```

Since the synchronization is expected to work in the other direction (from this repository towards the database) the command shouldn't be required anymore, but is left for completeness.

### DIFF
--- a/docs/tests-naming-convention.md
+++ b/docs/tests-naming-convention.md
@@ -59,3 +59,16 @@ DSP002.201 - External HDMI display in OS (Ubuntu)
     ...    Previous IDs: DSP002.001
     (...)
 ```
+
+# Synchronization with database
+
+Each time a test name, ID or module is changed, the change must be reflected in
+`test_cases.json` file. This includes previous IDs, which are saved in
+`changed_to` key. Note that there is only one ID in that key. Each ID change
+results in _creation_ of a new object in JSON file, and _modification_ of the
+existing one by pointing to the new ID.
+
+The file is ordered alphabetically by `_id`, keep it that way.
+
+Test IDs are **never removed**. This is required to keep references in the old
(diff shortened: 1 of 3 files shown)


artur-rs (approved):

philipanda (approved):
That's a good idea.

Although I am worried by the redundancy regarding the ID renamings.
The new IDs need to be included in test case docstrings in .robot files, in this json, they need then to be updated in the database, and docs.dasharo, despite being slowly deprecated, still has to be as up to date as we are able to keep it.
It would be perfect if this process could've been made a little more straightforward.

I believe that renaming the test cases could be done automatically pretty reliably provided that will be the only change in a commit. That way the diff could be parsed and converted into an update to this json file, which in return could automatically be used to make a transaction to the database.
Modifying the docstrings, and docs.dasharo could also be automated in the same way, although I am not sure if creating such automation tools would pay off, because the process has not matured yet and might very well change soon

*edit: Oh! Nevermind! :smile: https://github.com/Dasharo/open-source-firmware-validation/pull/852/files
The docstrings and docs.dasharo redundancy still applies though

krystian-hebel:
> The docstrings and docs.dasharo redundancy still applies though

Docstrings could be validated (and updated, if needed) with pre-commit. As for docs.dasharo, I've heard that we're migrating away from having the tests there anyway and all tests, even the manual ones, will be described here. In fact, some of them are already present in the copy, as it is a dump of the database as of today, but it will have to be validated at some point.

PR #852, opened 2025-05-30
scripts/synchronize-db.py: script for updating the database

krystian-hebel:
Script assumes to be started from top repo directory. It currently has IP hardcoded to localhost for testing purposes. User credentials are obtained interactively to avoid storing them in bash history, this may have to be changed to work with GH workflow, but it wouldn't be able to access our DB instance anyway.

### DIFF
--- a/scripts/synchronize-db.py
+++ b/scripts/synchronize-db.py
@@ -1,0 +1,161 @@
+#!/usr/bin/env python
+
+# SPDX-FileCopyrightText: 2025 3mdeb <contact@3mdeb.com>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import json
+import sys
+from getpass import getpass
+
+import requests
+from requests.auth import HTTPBasicAuth
+
+# FIXME: no hardcoded IPs!
+IP = "127.0.0.1"
+COUCHDB_URL = f"http://{IP}:5984/test_cases"
+
+FAIL = "\033[91m\033[1m"
+ENDC = "\033[0m"
+
+
+def main():
+    with open("test_cases.json", "r") as file:
+        local_json = json.load(file)
+
+    try:
+        response = requests.get(
+            f"{COUCHDB_URL}/_all_docs?include_docs=true",
+            auth=HTTPBasicAuth("web", "site"),  # Add authentication
+        )
+    except Exception as e:
+        print(f"{FAIL}Error connecting to the database: {repr(e)}{ENDC}")
+        sys.exit(1)
+
+    if response.status_code in (200, 201):
+        print(f"Test cases successfully downloaded.")
+    else:
+        print(
+            f"{FAIL}Failed to download test cases. HTTP {response.status_code}: {response.text}{ENDC}"
+        )
+        sys.exit(1)
+
+    remote_json_full = response.json()["rows"]
+    remote_json = []
+
+    # Make remote JSON look like the copy in file
+    for doc in remote_json_full:
+        # skip design docs
+        if doc["id"].startswith("_design"):
+            continue
+
+        # make a copy of just the body of the document (no id, key or value)
+        new_doc = doc["doc"].copy()
+
+        # remove _rev
+        del new_doc["_rev"]
+
+        remote_json.append({"doc": new_doc})
+
+    if local_json == remote_json:
+        print(f"Remote is up to date, nothing to do.")
+        sys.exit(0)
+
+    # Filter out cases that aren't changed from both lists
+    for doc in remote_json[:]:
+        try:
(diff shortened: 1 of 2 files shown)


scripts/synchronize-db.py
```diff
@@ -0,0 +1,161 @@
...
+
+import json
+import sys
+from getpass import getpass
+
+import requests
+from requests.auth import HTTPBasicAuth
+
+# FIXME: no hardcoded IPs!
+IP = "127.0.0.1"
```
artur-rs: Ideas:
* local config, ex. `~/.config/osfv-dashboard/db.yml` with correct value and README setup update?
* or if it's meant for GH Actions, then GH secrets and:

```
import os

IP = os.getenv("SERVER_IP")
```

krystian-hebel: Eventually it would be ideal to have this handled by GH Actions on merge, so the second option looks more appropriate. Thanks for the suggestion, I'll integrate it.

Before GHA would be possible, there are still things to be done though:
- expose DB so it is publicly accessible, at least for the parts required for normal operation (so just 3 main databases, and no admin panel etc.)
- make it use HTTPS instead of plain HTTP so the passwords aren't sent in plain text
  - this and above probably could be handled by nginx with no changes to existing DB VM
- validate content of JSON, there are at least some missing `changed_to` fields
- create an user for GHA with appropriate permissions (or use existing one after changing the password, if/when this script becomes the only, Proper Way™ of updating test cases in the database)
  - user/password would also be stored in GH secrets, which would require small change to the script to make it non-interactive
- (optional, but strongly suggested) pre-commit or another check that would ensure that all changes to .robot files are actually reflected in JSON as well

scripts/synchronize-db.py
```diff
@@ -0,0 +1,161 @@
...
+                print(f"{FAIL}Error connecting to the database: {repr(e)}{ENDC}")
+                sys.exit(1)
+
+            if response.status_code in (200, 201):
+                print(f"Test case {doc['doc']['_id']} successfully updated.")
+            else:
+                print(
+                    f"{FAIL}Failed to update {doc['doc']['_id']}. HTTP {response.status_code}: {response.text}{ENDC}"
+                )
+                # XXX: return here or keep going?
```
philipanda: The changes are not atomic, we are sending them doc by doc, so I'd say keep going, but record what failed. It's not like we can restore the changes that happened up to an error to make it feel like a transaction.

scripts/synchronize-db.py
```diff
@@ -0,0 +1,161 @@
...
+    try:
+        confirm = input(f"\nIs that correct? [N/y] ")
+    finally:  # to handle ^C (KeyboardInterrupt) and ^D (EOFError) gracefully
+        if confirm != "y" and confirm != "Y":
+            print(f"\nAborting")
+            sys.exit(0)
+
+    try:
+        user = input(f"Enter user name: ")
+        # TODO: check if this works without interactive terminal
```
philipanda, line 107: Passing anything using a pipe does not work, eg:
```yes 'y\nuser\npassword\n' | scripts/synchronize-db.py```
or
```scripts/synchronize-db.py <<<$'y\nuser\npass\n'```
or
```{echo "y"; echo "user"; echo "pass"} | scripts/synchronize-db.py```
The `y` and username are read fine. The password is not read from the stream.

scripts/synchronize-db.py
krystian-hebel: Added in https://github.com/Dasharo/open-source-firmware-validation/pull/852/commits/15871c798393c48ec2756711392145a2f1b13044, I've changed `SERVER_IP` to something a bit more specific and also included port number, no need to keep it public while IP is a secret. It can still be run locally with `DB_SERVER_IP=127.0.0.1:1234 ./scripts/synchronize-db.py`

scripts/synchronize-db.py
krystian-hebel: https://github.com/Dasharo/open-source-firmware-validation/pull/852/commits/15871c798393c48ec2756711392145a2f1b13044, failed cases are printed at the end now, e.g.:
image

artur-rs (approved):

scripts/synchronize-db.py
philipanda: That's not a problem that needs resolving, just a FYI, because I wanted to know :)

philipanda (approved):

krystian-hebel:
I've merged this PR to https://github.com/Dasharo/open-source-firmware-validation/pull/850, these two are closely related and rebasing one PR is faster than rebasing two of them.

PR #856, opened 2025-06-02
dasharo-security/cbnt.robot: add basic CBnT log parsing test

mkopec:


platform-configs/include/novacustom-common.robot
```diff
@@ -96,6 +96,7 @@ ${DTS_FIRMWARE_FLASHING_SUPPORT}=                   ${TRUE}
 ${DTS_EC_FLASHING_SUPPORT}=                         ${TRUE}
 ${ACPI_DRIVER_SUPPORT}=                             ${TRUE}
 ${ACPI_CAMERA_SWITCH_SUPPORT}=                      ${FALSE}
+${INTEL_CBNT_SUPPORT}=                              ${TRUE}
```
filipleple, line 99: IIRC you also have to add it to `platform-configs/include/default.robot`, or it will FAIL on other platforms instead of a nice skip

filipleple (changes requested):

platform-configs/include/novacustom-common.robot
mkopec: good point, added to https://github.com/Dasharo/open-source-firmware-validation/pull/856/commits/8de23b5e1fcf9dccc4a82c214043744278342b6c

filipleple (approved):

dasharo-security/cbnt.robot
```diff
@@ -0,0 +1,49 @@
...
+    [Tags]    robot:private
+    [Arguments]    ${os_id}
+    Power On
+    Boot System Or From Connected Disk    ${os_id}
+    Login To Linux
+    Switch To Root User
+    ${out_cbmem}=    Execute Command In Terminal    cbmem -1
+    Should Contain    ${out_cbmem}    FACB:${SPACE*20}1
+    Should Contain    ${out_cbmem}    measured boot:${SPACE*11}1
+    Should Contain    ${out_cbmem}    verified boot:${SPACE*11}1
```
SergiiDmytruk: This looks fragile so I looked up the check in [the docs](https://robotframework.org/robotframework/latest/libraries/BuiltIn.html#Should%20Contain) and there is `collapse_spaces=True` which should allow dropping `${SPACE*X}`.

Probably not necessary to do it here if there won't be other similar checks, but I see no uses of this parameter in the repository and wanted to mention it.

SergiiDmytruk (approved):
As I understand, this isn't meant to be an exhaustive test of IBG's configuration, just a quick smoke test, so LGTM.

dasharo-security/cbnt.robot
macpijan: Or https://robotframework.org/robotframework/latest/libraries/BuiltIn.html#Should%20Match%20Regexp ?

SergiiDmytruk: I actually went to look for a regexp match when I saw that parameter.  `Should Match Regexp` can allow matching the whole value (e.g., `FACB:\\S+1\\n`) and avoid false-positive matches.

filipleple:
@mkopec needs a rebase

dasharo-security/cbnt.robot
mkopec: https://github.com/Dasharo/open-source-firmware-validation/pull/856/commits/9af03df13906d56ac30ee7d9c8cba4bbec5a75b4

mkopec:
@filipleple rebased

PR #859, opened 2025-06-04
setup-and-boot-menus: Mount a bootable drive instead of UEFI shell

philipanda:
Fixes the test always FAILing because UEFI Shell is no longer distributed with Dasharo
[setup-and-boot-menus_log.zip](https://github.com/user-attachments/files/20589382/setup-and-boot-menus_log.zip)

### DIFF
--- a/.github/workflows/qemu-self-test.yml
+++ b/.github/workflows/qemu-self-test.yml
@@ -11,10 +11,12 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - name: Checkout repository and submodules
+    - name: Checkout repository
       uses: actions/checkout@v2
-      with:
-        submodules: 'recursive'
+
+    - name: Chceckout submodules
+      run: |
+        git submodule update --init --checkout --recursive
 
     - name: Set up QEMU
       run: |
@@ -43,6 +45,18 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install -r requirements.txt
+
+    - name: Set up git-annex
+      uses: jstritch/setup-git-annex@v1
+
+    - name: Set up osfv-test-data
+      working-directory: osfv-test-data
+      run: |
+        set +e
+        git config --local user.email "github-actions[bot]@users.noreply.github.com"
+        git config --local user.name "github-actions[bot]"
+        git annex pull
+        ./setup.sh
 
     - name: Start keywords self-tests with QEMU
       run: |
--- a/self-tests/dasharo-system-features-menus.robot
+++ b/self-tests/dasharo-system-features-menus.robot
@@ -124,7 +124,7 @@ Parse Power Management Options
     Skip If    not ${DASHARO_POWER_MGMT_MENU_SUPPORT}
     ${power_menu}=    Parsing Dasharo Submenu Verification    Power Management Options
     ${power_entries}=    Get Length    ${power_menu}
-    Should Be Equal As Integers    ${power_entries}    6
+    Should Be Equal As Integers    ${power_entries}    5
     Should Match Regexp    ${power_menu}[0]    ^Fan profile <.*>.*$
     Menu Construction Should Not Contain Control Text    ${power_menu}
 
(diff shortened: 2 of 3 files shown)


macpijan (changes requested):
Modified test case fails in this pipeline:

```
Parse One Time Boot Menu :: Test entering into User Password Manag... | FAIL |
[ No bootable media found ] does not contain value 'QEMU QEMU USB HARDDRIVE'.
```

If it depends on some external PR, please reference it.

philipanda:
> Modified test case fails in this pipeline:
> 
> ```
> Parse One Time Boot Menu :: Test entering into User Password Manag... | FAIL |
> [ No bootable media found ] does not contain value 'QEMU QEMU USB HARDDRIVE'.
> ```
> 
> If it depends on some external PR, please reference it.

It requires `osfv-test-data` to be initialized. Have you run `osfv-test-data/setup.sh`? The image `osfv-test-data/dts/dts-base-image-v2.1.3.wic` is used here and without initializing the osfv-test-data the file does not exist as it needs to be extracted from `osfv-test-data/dts/dts-base-image-v2.1.3.wic.gz`

\* edit
Oh, you meant the pipeline, right. It might be that the osfv-test-data is not initialized and so the bootable image is not available. I'll check that

macpijan:
Yes, I mean pipeline, I do not run anything myself here.

I just want to ensure that we do not accustomed to merging PRs with CI checks failing. I have already enabled the requirements for the CI to pass prior merging. So we need to (preferably) either fix it or disable this failing case until resolved.

philipanda:
@macpijan
The test now passes after correctly setting up `osfv-test-data`.

The `actions/checkout@v2` does not initialize `osfv-test-data` correctly, even with `submodules: recursive`. Instead `git submodule update --init --checkout --recursive` had to be run manually in a separate job.

Additionally a `user.email` and `user.name` needs to be set up in order to use `git annex pull`, and `-e` bash flag has to be unset using `set +e` to allow `git annex pull` to fail. That's because one of the files in `osfv-test-data` is broken right now https://github.com/Dasharo/osfv-test-data/issues/11. Other files will be downloaded just fine despite that one failing.

Not every test case had been run by the time I'm writing this. Especially the `Get Menu Construction Stress Test` will [take a very very long time ](https://github.com/Dasharo/open-source-firmware-validation/issues/870).

philipanda:
Only `Parse Power Management Options` tests fails, but it passes locally
[dasharo-system-features-menus_log.zip](https://github.com/user-attachments/files/20649834/dasharo-system-features-menus_log.zip)

philipanda:
@macpijan I am a bit troubled with the results of CI.
The CPU Throttling option really is missing from the Power Management Options menu, and so the fail is expected.
I've tested it manually on release v0.2.0 and v0.2.1.
image


What's worse is that some PRs, the newest one being #842, actually have this test PASS. It's like the option appears and disappears randomly on both v0.2.0 and v0.2.1 release for QEMU.
I can't confirm that, because I am yet to experience the option actually appearing in the menu.

philipanda:
That's the first Action to fail because of missing UEFI Shell https://github.com/Dasharo/open-source-firmware-validation/actions/runs/15215838318. That also is the first Action to fail because of the missing CPU throttling option too. Everything seems to indicate that it's the v0.2.1 release that removed the throttling option as well as UEFI shell, but nothing like that changed in the configs, and when booting v0.2.0 manually I can't see the option too.

philipanda:
@macpijan  I've been searching through the changes since v0.2.0 and it seems that the throttling menu options were reworked by @miczyg1 about a year ago. [This commit](https://github.com/Dasharo/edk2/commit/a5b3b31da521f#diff-fda7be7a2a48a17bb0f3f3b1c3b52b4bc3827f711badc19793e0e840176fbbe7L76-R76) made the cPU throttling disabled by default, and QEMU was never actually explicitly adding it in its coreboot configs. Due to this change the option has been removed from the Setup menu.

By adding a config option the entry in the menu can be restored, but the menu has changed. It now shows two lines, one of which is readonly. The test has to be modified anyway.

philipanda:
The reworked Power Management menu is too big to fit on one screen. I've tried implementing parsing multiscreen menus, but that seems a little too complex for now, so I've just skipped the last line check for this one menu.

macpijan (approved):

PR #872, opened 2025-06-09
Restore docking station agnostic utc cases

philipanda:
The changes are:
- USB Type-C Display output
  - was UTCX08.XXX USB Type-C Display output (<OS>) (ME: <STATE>) (<DOCKING STATION>)
  - is     UTC007.XXX USB Type-C Display output (<OS>) (ME: <STATE>)
  - Changed from automated to manual as the implementation was wrong. It was detecting a docking station, not a display. A docking station should not even be used for this test.
- USB Type-C PD current limiting
  - was UTC{1,2,3}33.{201,202,301} USB Type-C PD current limiting (<OS>) (ME: <STATE>) (<DOCKING STATION>)
  - is     UTC033.001 USB Type-C PD current limiting (Firmware) (ME: <STATE>)
- USB Type-A charging capability
  - was UTCX02.001 USB Type-A charging capability (Firmware) (ME: <STATE>) (<DOCKING STATION>)
  - is    `UTC002.001 USB Type-A charging capability (Firmware) (ME: <STATE>)
- Thunderbolt 4 USB Type-C power output
  - was UTCX04.001 Thunderbolt 4 USB Type-C power output (Firmware) (ME: <STATE>) (<DOCKING STATION>)
  - is     UTC002.001 USB Type-A charging capability (Firmware) (ME: <STATE>)

### DIFF
--- a/dasharo-compatibility/usb-type-c.robot
+++ b/dasharo-compatibility/usb-type-c.robot
@@ -22,16 +22,112 @@ Suite Teardown      Run Keyword
 
 *** Test Cases ***
 # Not automated
-# UTC101.001 USB Type-A charging capability (Firmware) (ME: Enabled) (WL-UMD05 Pro Rev.E)
+# UTC001.001 USB Type-A charging capability (Firmware) (ME: Enabled)
 #    [Documentation]    This test verifies that the USB-A ports are able to provide
 #    ...    charging to a connected smartphone.
-#    Usb Type-A Charging Capability    001    Enabled    WL-UMD05 Pro Rev.E
+#    Usb Type-A Charging Capability    001    Enabled    None
 
 # Not automated
-# UTC103.001 Thunderbolt 4 USB Type-C power output (Firmware) (ME: Enabled) (WL-UMD05 Pro Rev.E)
+# UTC003.001 Thunderbolt 4 USB Type-C power output (Firmware) (ME: Enabled)
 #    [Documentation]    This test verifies that the Thunderbolt 4 port is able
 #    ...    to provide charging to a USB Type-C accessory.
-#    Thunderbolt 4 Usb Type-C Power Output    001    Enabled    WL-UMD05 Pro Rev.E
+#    Thunderbolt 4 Usb Type-C Power Output    001    Enabled    None
+
+# Not automated
+# UTC007.201 USB Type-C Display output (Ubuntu) (ME: Enabled)
+#    [Documentation]    Check whether a display connected to the USB Type-C port
+#    ...    is recognized by the OPERATING_SYSTEM.
+#    Skip If    not ${USB_TYPE_C_DISPLAY_SUPPORT}    UTC007.201 not supported
+#    Skip If    not ${TESTS_IN_UBUNTU_SUPPORT}    UTC007.201 not supported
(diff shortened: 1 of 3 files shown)


macpijan:
Does it need some changes in https://github.com/Dasharo/open-source-firmware-validation/blob/develop/test_cases.json ?

philipanda:
> Does it need some changes in https://github.com/Dasharo/open-source-firmware-validation/blob/develop/test_cases.json ?

Right, some changes should happen there.
I am not sure what's the correct way of documenting the changes there though. The test cases that were restored in this PR shouldn't be split into docking station variants altogether. I suppose we should just add another set of mappings, but will our current tools be able to handle a sequence of updates? Like A->B->C? Tests other than the mentioned in the description weren't affected.

krystian-hebel:
Tools will handle A->B->C (assuming `Previous IDs` has correct format, as described in documentation). They wouldn't handle A->(B,C), and the way I understand what happened for UTC is something like A->(B,C,D,E)->F. None of the UTC cases was deprecated in the database, and if my understanding is correct, now it should be enough to do A->F, B->F, C->F etc.

I'm more worried about 1-to-N cases for other UTC that those mentioned here. There is no future-proof way that I can think of, even if we were to change the logic to support such relations, each case would point to 3 other cases (or however many docking stations we currently test), but it won't account for the stations added later. In my opinion, in such case adding a new set of tests for a new dock to scope is an action, and shouldn't be done by the tool after inaction of person using this tool. After all, new stations aren't added often, and if they are, then next releases can be copied from a fresh one, and not from a very old one.

Just to confirm, the changes mentioned here would be A->(B,C...)->Z, and not A->(B,C...)->A, right?

philipanda:
> Tools will handle A->B->C (assuming `Previous IDs` has correct format, as described in documentation). They wouldn't handle A->(B,C), and the way I understand what happened for UTC is something like A->(B,C,D,E)->F. None of the UTC cases was deprecated in the database, and if my understanding is correct, now it should be enough to do A->F, B->F, C->F etc.
> 
> I'm more worried about 1-to-N cases for other UTC that those mentioned here. There is no future-proof way that I can think of, even if we were to change the logic to support such relations, each case would point to 3 other cases (or however many docking stations we currently test), but it won't account for the stations added later. In my opinion, in such case adding a new set of tests for a new dock to scope is an action, and shouldn't be done by the tool after inaction of person using this tool. After all, new stations aren't added often, and if they are, then next releases can be copied from a fresh one, and not from a very old one.
> 
> Just to confirm, the changes mentioned here would be A->(B,C...)->Z, and not A->(B,C...)->A, right?

The changes would be like

A->(B,C,D,E,F,G)->(H,I)
- the docking station model is not relevant, but the ME state is, so there still have to be two variants for that

macpijan:
Blocker: we are not sure how we would set the changed_to flags.

We consider regenerating the tests DB, we could then simply add these tests to the DB.

philipanda:
@BeataZdunczyk @macpijan The issue this PR is supposed to fix is still relevant and affects the v1.0.0-rc4 tests.

BeataZdunczyk:
@philipanda his PR isn’t needed anymore, right? Should we close it, or do you think it’s still relevant and worth finishing?

philipanda:
> @philipanda his PR isn’t needed anymore, right? Should we close it, or do you think it’s still relevant and worth finishing?

The changes from this PR were merged in some other PR it seems. No longer relevant

PR #876, opened 2025-06-09
scripts/list-tests-from-{json,robot}.sh: add new scripts

krystian-hebel:
Can be used to check whether test cases in source files and their copy in `test_cases.json` are in sync.

### DIFF
--- a/docs/adding-and-naming-test-cases.md
+++ b/docs/adding-and-naming-test-cases.md
@@ -53,7 +53,8 @@ loosely. `environment_id`s with leading `0` don't use the convention above.
 
 When updating a test case ID to follow the convention,
 append the old ID in the keyword documentation after `Previous IDs:`
-to help with identifying the test cases during transition. Example:
+to help with identifying the test cases during transition. Example (not from
+actual test case):
 
 ```robot
 DSP002.201 - External HDMI display in OS (Ubuntu)
@@ -61,13 +62,15 @@ DSP002.201 - External HDMI display in OS (Ubuntu)
     ...    Linux OS. An external HDMI display must be provided in
     ...    the platform config.
     ...
-    ...    Previous IDs: DSP002.001
+    ...    Previous IDs: ABC042.001 DSP002.001
     (...)
 ```
 
 If the ID is changed again for the same test, the newer `Previous ID` is added
 to the end of line, with single space as a separator (i.e. no comma or semicolon
 between them). This format is expected by a script described [later](#validate).
+In the example above, `ABC042.001` was changed to `DSP002.001`, which was later
+changed to `DSP002.201`.
 
 # Synchronization with database
 
@@ -174,7 +177,7 @@ as those with deprecated IDs:
   (i.e. test cases in all Dasharo `*.robot` files), and prints a sorted list of
   all of them.
 - `scripts/list-tests-from-json.sh` - finds active (without `changed_to` field)
-  and obsolete (without that field) IDs in `test_cases.json` file, and prints a
+  and obsolete (with that field) IDs in `test_cases.json` file, and prints a
   list of all of them. The script doesn't sort the output, which indirectly
   checks that JSON file is properly sorted.
 
--- a/docs/adding-and-naming-test-cases.md
+++ b/docs/adding-and-naming-test-cases.md
@@ -66,6 +66,10 @@ DSP002.201 - External HDMI display in OS (Ubuntu)
     (...)
 ```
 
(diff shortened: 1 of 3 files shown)


krystian-hebel:
PR marked as a draft since the task includes fixes to the test cases, but scripts and docs can already be reviewed.

docs/adding-and-naming-test-cases.md
```diff
@@ -66,6 +65,10 @@ DSP002.201 - External HDMI display in OS (Ubuntu)
     (...)
 ```
 
+If the ID is changed again for the same test, the newer `Previous ID` is added
+to the end of line, with single space as a separator (i.e. no comma or semicolon
+between them). This format is expected by a script described [later](#validate).
```
philipanda, line 71: An example is worth 1000 words

docs/adding-and-naming-test-cases.md
```diff
@@ -158,6 +161,50 @@ existing one by pointing to the new ID.
...
+Two scripts are used for listing existing test cases, both active ones as well
+as those with deprecated IDs:
+
+- `scripts/list-tests-from-robot.sh` - finds active (lines starting with
+  something resembling an ID) and obsolete (`Previous IDs`, see
+  [Transitioning](#transitioning)) IDs in files in `dasharo-*` directories
+  (i.e. test cases in all Dasharo `*.robot` files), and prints a sorted list of
+  all of them.
+- `scripts/list-tests-from-json.sh` - finds active (without `changed_to` field)
+  and obsolete (without that field) IDs in `test_cases.json` file, and prints a
```
philipanda:
```suggestion
  and obsolete (with that field) IDs in `test_cases.json` file, and prints a
```

docs/adding-and-naming-test-cases.md
```diff
@@ -158,6 +161,50 @@ existing one by pointing to the new ID.
...
+  list of all of them. The script doesn't sort the output, which indirectly
+  checks that JSON file is properly sorted.
```
philipanda, line 182: So maybe the script should check if that's the case?

scripts/list-tests-from-robot.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
+# SPDX-FileCopyrightText: 2024 3mdeb <contact@3mdeb.com>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# Get lines that start with ID, store each line as array element
+mapfile -t active < <(grep -Rh "^[A-Z]\{1,6\}[0-9]\{3\}\.[0-9]\{3\}" dasharo-*)
```
philipanda: ETHPERF has 7 letters and will be left out by this regular expression.

What if someone creates an invalid test case, like one containing 4 numbers after the dot, and forgets to add it to the json? That case won't be catched.

I think the expressions shouldn't be that restrictive when it comes to the number of characters

philipanda (changes requested):

docs/adding-and-naming-test-cases.md
krystian-hebel: https://github.com/Dasharo/open-source-firmware-validation/pull/876/commits/a910152d3489fb765b38b5db6368b45a07b160b7

docs/adding-and-naming-test-cases.md
krystian-hebel: https://github.com/Dasharo/open-source-firmware-validation/pull/876/commits/a910152d3489fb765b38b5db6368b45a07b160b7

docs/adding-and-naming-test-cases.md
krystian-hebel: `Possible use cases` do that, by comparing with always sorted list from `.robot` files. This script is for printing only, and the checks are to be run by the caller. If (when) they are used by CI, then it would be checked there, possibly with another tests. Right now (at least for me) the more common use case is _showing_ the differences, not just checking if there are any.

scripts/list-tests-from-robot.sh
krystian-hebel: Nice catch. I was wondering why I only saw `DEPRECATED` tests there, but I wrongfully assumed that they were moved to NET.

I've changed it to less restrictive `^[A-Z]*[0-9]\{1,10\}\.[0-9]\{1,10\}` and checked that it doesn't give any false-positives: https://github.com/Dasharo/open-source-firmware-validation/pull/876/commits/7c9ba52aabda497c5d0519679459959bb1c2494e

philipanda (approved):

PR #879, opened 2025-06-11
Add manual CUP tests cases to matrix

JakubRed:


dasharo-performance/manual_cpu_performance.robot
```diff
@@ -38,7 +38,7 @@ UPP001.301 Manual Single Threaded CPU Benchmark (Windows) (AC)
     Run A Test Manually    smallpt    ${SMALLPT_TEST_SCORE}    # run smallpt
     Run A Test Manually    crafty    ${CRAFTY_TEST_SCORE}    # run crafty
 
-UPP001.302 Manual Multi Threaded CPU Benchmark (Windows) (AC)
+UPP002.301 Manual Multi Threaded CPU Benchmark (Windows) (AC)
```
krystian-hebel, line 41: Normally this is forbidden, but since it is clearly not following the naming convention, and it is used only by one release so far (which I already fixed), I'll allow it this time, and this time only.

Normally it should be handled by listing `Previous IDs`, but in this case it would block the ID for whatever may get `302` suffix in the future.

krystian-hebel (approved):

Issue #674, opened 2025-01-16
Rework / remove configs in variables.robot


macpijan:
Every time I need to go through variables.robot and backtrack to configs and RTE setups to find out which setup is which, connected where, part of me dies inside.

We have already reduced dependency on this, 

If we want to keep stands configs in similar form, we can keep it, but must be completely ed reworked - e.g. use yaml to define stands.

This issue aims to gather all threads related to this effort.

Relevant milestones:
- [x]  drop RTE's cpuid - it was checked with no obvious benefit - it has been used some time ago already
- [ ] drop USB 
  - [ ] issue: https://github.com/Dasharo/open-source-firmware-validation/issues/47
- [ ] TBD

PR #881, opened 2025-06-11
Odroid h4 ultra tests

miczyg1:
Introduce to new tests: IBECC and ODROID netcard.

The netcard test is pretty straightforward. It checks the performance of two pairs of the network ports as in regular ETHPERF test.

IBECC simply verifies whether the error in dmesg appears when IBECC is disabled/enabled or not.

I also prepared the ODROID H4 Ultra target (the IPs are temporary, they will get updated when the platform gets installed in the lab).


miczyg1:
test results:
https://github.com/Dasharo/coreboot/pull/682#issuecomment-2962095821
https://github.com/Dasharo/coreboot/pull/680#issuecomment-2962543047

miczyg1:
I am struggling a bit with suspend tests. Manually it works, but in tests the output is read in wrong places and the test assume that SSH is used to connect, because after waking up `Login To Linux` is used. But with telnet we still are in the logged system... This breaks many tests it seems, because fastboot tests assumed similar things...

```diff
diff --git a/lib/sleep-lib.robot b/lib/sleep-lib.robot
index 71a200c2..c093116e 100644
--- a/lib/sleep-lib.robot
+++ b/lib/sleep-lib.robot
@@ -93,10 +93,12 @@ Perform Suspend Test Using FWTS
     ...    test by using Firmware Test Suite tool
     [Arguments]    ${test_duration}=40
     ${is_suspend_performed_correctly}=    Set Variable    ${FALSE}
+    Read From Terminal
     Write Into Terminal    fwts s3 -f -r /tmp/suspend_test_log.log
     Sleep    ${test_duration}s
-    Login To Linux
-    Switch To Root User
+    Read From Terminal
+    Write Bare Into Terminal    ${ENTER}
+    Read From Terminal Until Prompt
     ${test_result}=    Execute Linux Command    cat /tmp/suspend_test_log.log
     TRY
         Should Contain    ${test_result}    0 failed
```

macpijan:
> I am struggling a bit with suspend tests.

Still struggling here?

Generally, changes look fine. We should:
- [ ] Get rid of WIP commits prior merging
- [ ] make a decision on sleep tests - I am sure some of us have been running these as well, we should discuss how to apporach it

dasharo-compatibility/efibootmgr.robot
```diff
@@ -31,10 +31,18 @@ EBM001.201 Network Boot enable
...
     Skip If    not ${DASHARO_NETWORKING_MENU_SUPPORT}    EBM001.001 not supported
 
+    Set UEFI Option    NetworkBoot    ${FALSE}
+
     ${boot_menu}=    Get UEFI Boot Manager Entries
     Should Not Contain    ${boot_menu}    ${IPXE_BOOT_ENTRY}
 
     Set UEFI Option    NetworkBoot    ${TRUE}
+
+    IF    '${ENV_ID_FEDORA}' in ${TESTED_LINUX_DISTROS}
```
philipanda: In `EBM001.201` the `.201` means that this test is performed on Ubuntu (Check `docs/adding-and-naming-test-cases.md`). If the test should be performed on Fedora or any other OS, then a new test case must be created

dasharo-compatibility/memory-ibecc.robot
```diff
@@ -0,0 +1,72 @@
...
+    # Restore default IBECC state and let the memory be trained again
+    Set IBECC State    ${FALSE}
+    Enter Setup Menu Tianocore
+
+
+*** Keywords ***
+Set IBECC State
+    [Arguments]    ${target_state}
+    Power On
+    Enter Setup Menu Tianocore
```
philipanda: I think you could use `Enter Setup Menu Tianocore And Return Construction` instead of a couple of the following lines.

dasharo-compatibility/memory-ibecc.robot
```diff
@@ -0,0 +1,72 @@
...
+    ${out}=    Read From Terminal Until    <Enter>=Select Entry
+    # Get IBECC state and change it if necessary
+    ${setup_menu}=    Parse Menu Snapshot Into Construction    ${out}    3    1
+    ${dasharo_menu}=    Enter Dasharo System Features    ${setup_menu}
+    ${memory_menu}=    Enter Dasharo Submenu    ${dasharo_menu}    Memory Configuration
+    ${status}=    Set Option State    ${memory_menu}    Memory In-Band ECC    ${target_state}
```
philipanda: If the intention is only to set a new Setup menu option, then I would suggest describing the path to it in `lib/bios/menus.py` and using the `Set UEFI Option` keyword instead of re-implementing this logic again, unless I can't see some non-standard step that is required here

dasharo-compatibility/nvme-support.robot
```diff
@@ -34,7 +34,16 @@ NVM001.001 NVMe support in firmware
     Depends On    ${TESTS_IN_FIRMWARE_SUPPORT}
     Power On
     ${out}=    Enter Boot Menu Tianocore And Return Construction
-    Should Contain    ${out}    ${CLEVO_DISK}
+    ${ssd_list}=    Get Current CONFIG List Element    Storage_SSD
```
philipanda, line 37: `variables.robot` is being deprecated https://github.com/Dasharo/open-source-firmware-validation/issues/674
I would suggest not adding new dependencies on it if possible.
I appreciate the more verbose messaging during this test though

keywords.robot
```diff
@@ -1522,3 +1540,13 @@ Should Contain All
     FOR    ${substring}    IN    @{substrings}
         Should Contain    ${string}    ${substring}
     END
+
+Deploy Uefi Shell
```
philipanda, line 1551: UEFI Shell should is installed on the system drive on every DUT by the test suite `util/basic-platform-setup.robot`. It should always be run prior to any testing to ensure that all test dependencies are installed on the device. Check https://github.com/dasharo/open-source-firmware-validation?tab=readme-ov-file#basic-platform-setup

I suggest running the suite on your DUT to ensure everything is prepared for testing instead of implementing it in test cases

lib/sleep-lib.robot
```diff
@@ -102,7 +102,11 @@ Perform Suspend Test Using FWTS
         Login To Linux
         Switch To Root User
     END
-    ${test_result}=    Execute Command In Terminal    cat /tmp/suspend_test_log.log
+    # Clean up console before reading file
+    Read From Terminal
```
philipanda: `Execute Command In Terminal` already flushes the terminal before executing the requested command via serial console using `Telnet.Read`

os-config/ansible/os-specific-tasks/201.yaml
```diff
@@ -13,3 +13,34 @@
 
 - name: Install UEFI Shell
   ansible.builtin.command: /tmp/deploy-shell-efi.sh /tmp/Shell.efi
+
+- name: Install docker required packages
```
philipanda, line 17: Could you also add this to the tasks for Fedora? (202.yaml)

platform-configs/odroid-h4-plus.robot
```diff
@@ -42,6 +42,7 @@ ${AUDIO_SUBSYSTEM_SUPPORT}=                     ${TRUE}
 ${EXTERNAL_HEADSET_SUPPORT}=                    ${TRUE}
 ${EXTERNAL_HDMI_DISPLAY_SUPPORT}=               ${TRUE}
 ${EXTERNAL_DISPLAY_PORT_SUPPORT}=               ${TRUE}
+${FAN_SPEED_MEASURE_SUPPORT}=                   ${FALSE}
```
philipanda: It's already set to ${FALSE} in `default.robot`

platform-configs/odroid-h4-ultra.robot
```diff
@@ -0,0 +1,149 @@
...
+${USB_DISKS_DETECTION_SUPPORT}=                 ${TRUE}
+${USB_KEYBOARD_DETECTION_SUPPORT}=              ${TRUE}
+${NVME_DISK_SUPPORT}=                           ${FALSE}
+${SATA_SUPPORT}=                                ${TRUE}
+${EMMC_SUPPORT}=                                ${TRUE}
+${AUDIO_SUBSYSTEM_SUPPORT}=                     ${TRUE}
+${EXTERNAL_HEADSET_SUPPORT}=                    ${TRUE}
+${EXTERNAL_HDMI_DISPLAY_SUPPORT}=               ${TRUE}
+${EXTERNAL_DISPLAY_PORT_SUPPORT}=               ${TRUE}
+${FAN_SPEED_MEASURE_SUPPORT}=                   ${FALSE}
```
philipanda: Already in `default.robot`

philipanda (changes requested):

dasharo-compatibility/memory-ibecc.robot
miczyg1: Yes, I probably wrote this when these keywords were not available.

dasharo-compatibility/nvme-support.robot
miczyg1: I see. Any suggestion how to get rid of `CLEVO_DISK` used on every platform just to validate NVMe visibility in firmware? Hwo do you plan to support multiple boards that will source the same configs (e.g. two ODROIDs H4+ ) but with different disks connected?

Is it even good to have a variable with clevo in name? Who made this up?

keywords.robot
miczyg1: That doesn't work.

I know that BPS deploys the UEFI Shell, but as soon as BPS finishes and next suite runs, e.g. dasharo-compatibility, the suite's  `__init__.robot` flashes the firmware again, and the Shell option is already gone.

tl;dr

BPS deploying the  UEFI Shell is useless during regression...

lib/sleep-lib.robot
miczyg1: It didn't before and I had output from previous command(s) when testing it...

I would have to check again, but right now the suspend tests won't work on Ubuntu, because of upstream ubuntu bug causing the fwts to fail to install: https://bugs.launchpad.net/ubuntu/+source/fwts/+bug/2111123

os-config/ansible/os-specific-tasks/201.yaml
miczyg1: I have no way to test it on ODROID, sorry. So to avoid breaking something, I won't add it.

keywords.robot
miczyg1: Also `__init__.robot` isn't the only place where the firmware is flashed, so it is very easy for the UEFi Shell to disappear from boot option list...

dasharo-compatibility/nvme-support.robot
philipanda: It probably was added by someone who at the time did not fully understand what the variables.robot file was used for? I think ideally the information about hardware configuration of a given DUT should be outsourced to snipeit and allow to override it using environment variables.

There are multiple platform-configs variables that really shouldn't be there, because as you said, another DUT might very well have a different disk model attached during testing. Unless we are enforcing the tests to be run using a well defined configuration the platform-configs need to be modified on every release because of that.

Helping with this is not the reason for this PR, so from my side using the variables.robot here is fine. Unless you have a vision of how to solve this better without diving deeper into the variables.robot/platform-configs situation, then that would be great

keywords.robot
philipanda: Hmm, it makes sense. Afterall the bootentry is manually created using efibootmgr, so it won't persist after flashing and the firmware is not discovering this bootable file on its own.

I am not sure what would be needed for the automatic discovery to work. Maybe a new tiny partition for the UEFI shell should be created on the system drive? Ideally all the dependencies should be installed as a preparation step to free the tests from this chore. Maybe we should create an issue to make the installation of UEFI shell during BPS persistent? Until that's solved I think your solution is good.

os-config/ansible/os-specific-tasks/201.yaml
philipanda: So maybe let's create an issue for that? Docker is definitely a dependency of some tests and it's clearly an oversight that it's not installed during BPS

lib/sleep-lib.robot
miczyg1: I had to change this anyways https://github.com/Dasharo/open-source-firmware-validation/pull/881/commits/ef4592f2d9ef483fe90f344f300c2b19af8411c7

The problem in fact was that the `Execute Command In Terminal` expects the promtp to appear after command is executed. However, the FWTS S3 test will not return the prompt after platform resumes. One has to additionally press ENTER for the prompt to appear again.

keywords.robot
miczyg1: https://github.com/Dasharo/open-source-firmware-validation/issues/1059

os-config/ansible/os-specific-tasks/201.yaml
miczyg1: https://github.com/Dasharo/open-source-firmware-validation/issues/1060

platform-configs/odroid-h4-plus.robot
miczyg1: 6012420f

platform-configs/odroid-h4-ultra.robot
miczyg1: 6012420f

dasharo-compatibility/nvme-support.robot
miczyg1: >  Unless we are enforcing the tests to be run using a well defined configuration the platform-configs need to be modified on every release because of that.

I still believe we are running tests na a well defined configuration, no? Otherwise, why would we describe HW matrix on dasharo docs?

dasharo-compatibility/memory-ibecc.robot
miczyg1: 61e150cc

dasharo-compatibility/memory-ibecc.robot
miczyg1: 61e150cc

lib/sleep-lib.robot
philipanda: I wonder if it's specific to the version of FWTS you've had installed here? I had no issue with running suspend tests for NUCBOX v0.9.0 via serial console. The changes look like they won't break anything should the prompt appear normally, so I think it's fair to use them

dasharo-compatibility/nvme-support.robot
philipanda: > I still believe we are running tests na a well defined configuration, no? 

That's what I would like to believe, but I don't think everyone makes sure to follow it 100%.
Maybe if the DUT's weren't shuffled regularily and the precise set of models of resources like keyboards and usb-drives was easier to find and complete, then it wouldn't be an issue. In practice, I doubt any tester would invalidate a test result just because the model name of used usb-keyboard was different than expected.
Maybe they should?
Anyhow that's not related to the changes, so from me the conversation is resolved. We might continue the topic of the hardware matrix and variables.robot somewhere else

dasharo-compatibility/device-detection.robot
```diff
@@ -19,11 +19,15 @@ Resource            ../keys.robot
 Suite Setup         Run Keywords
 ...                     Prepare Test Suite
 ...                     AND
+...                     Skip If    '${OPTIONS_LIB}' != 'options-lib_dcu'
+...                     AND
+...                     Skip If    '${DUT_CONNECTION_METHOD}' == 'telnet'
+...                     AND
```
philipanda, line 27: Why isn't it possible to perform the tests using the Setup Menu?
The option to disable the USB stack should be perfectly possible to be changed using `options-lib_uefi-setup-menu`
```
    "UsbDriverStack": [
        "Dasharo System Features",
        "USB Configuration",
        "Enable USB stack",
    ],
    ```

Comment (philipanda) [reply dasharo-compatibility/device-detection.robot:27]: I see, it would be possible had the test been written correctly:
```
Login To Linux With Root Privileges
Set UEFI Option UsbDriverStack Disabled
Execute Reboot Command
Sleep 10s
```
https://github.com/Dasharo/open-source-firmware-validation/issues/1061

File: dasharo-compatibility/odroid-netcard.robot [hunk 11]
```diff
@@ -22,12 +23,21 @@ Suite Setup         Run Keyword
 Suite Teardown      Run Keyword
 ...                     Log Out And Close Connection
 
-Default Tags        automated
-
 
 *** Test Cases ***
-STUB001.001 External flash explicit
-    Flash Device Via External Programmer
+NETCARD001.201 Check Performance of 2.5G Wired Network Interface On Netcard (Ubuntu)
```
philipanda: That's some wild diff from GitHub, I must say :laughing: 
The STUB test suite appears at the osfv-dashboard, so we can't delete it just like that, even if it is arguably useful, and even if so, then shouldn't be a part of dasharo-compatibility.

Why not adjust the ETHPERF test suite and create a new one? They seem to do  the same thing apart from some manual preparation in ETHPERF001.201, which could be be skipped conditionally  on odroid if that's not needed

dasharo-performance/fast-boot.robot
```diff
@@ -20,15 +20,16 @@ FBT001.201 Fast Boot Reduces Boot Time
...
     Login To Linux
     Switch To Root User
     Set Fast Boot State    on
     ${fast_min}    ${fast_max}    ${fast_avg}    ${fast_stddev}=
     ...    Measure FW Boot Time On Linux    ${ITERATIONS}
 
     Log To Console    \nMeasuring boot time without Fast Boot\n
-    Login To Linux
-    Switch To Root User
+    # We should still be in the OS as root after Measure FW Boot Time On Linux
```
philipanda: That should be a commit message, not a comment right?
Without context I wouldn't understand why this comment is here

lib/sleep-lib.robot
```diff
@@ -89,15 +89,19 @@ Perform Suspend Test Using FWTS
...
+        Read From Terminal
+        Write Bare Into Terminal    ${ENTER}
+        Read From Terminal Until Prompt
+    END
+    IF    '${DUT_CONNECTION_METHOD}' == 'SSH'
         Login To Linux
         Switch To Root User
     END
-    ${test_result}=    Execute Command In Terminal    cat /tmp/suspend_test_log.log
+    ${test_result}=    Execute Linux Command    cat /tmp/suspend_test_log.log
```
philipanda: Why was that required?
We would like to transfer from using `keywords.robot` keywords to the ones defined in `lib` and ideally remove `keywords.robot` altogether

platform-configs/odroid-h4-plus.robot
```diff
@@ -10,24 +10,24 @@ ${FLASH_LENGTH}=                                ${TBD}
 ${MANUFACTURER}=                                Hardkernel
 ${CPU}=                                         Intel(R) N97
 ${POWER_CTRL}=                                  RteCtrl
-${FLASH_VERIFY_METHOD}=                         tianocore-shell
+${FLASH_VERIFY_METHOD}=                         none
```
philipanda, line 13: Why not leave `tianocore-shell` if you've made a workaround that restores UEFI Shell?
That way we are not verifying if the correct binary is really flashed at  `Prepare Platform`

scripts/regression-rerun-failed.sh
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env bash
+
+# SPDX-FileCopyrightText: 2024 3mdeb <contact@3mdeb.com>
```
philipanda: It's 2025

philipanda (changes requested):

platform-configs/odroid-h4-plus.robot
miczyg1: The deploy shell keyword assumes there is an OS installed on the platform. However, the tianocore-shell flash verify will instantly try to boot the shell right after flashing. So that method is not suitable IMHO. It was good as long as Shell was included by default.

lib/sleep-lib.robot
miczyg1: TBH, I don't know why this keyword changed. Reverted: 441aa43c4e04f11ba2c61f75aa949972f750a7e0

dasharo-compatibility/odroid-netcard.robot
miczyg1: > Why not adjust the ETHPERF test suite and create a new one? They seem to do the same thing apart from some manual preparation in ETHPERF001.201, which could be be skipped conditionally on odroid if that's not needed

The way it is tested resembles ETHPERF, but it just uses a performance test to prove compatibility (i.e. that the netcard works correctly). That is why I decided for it to be separate.

dasharo-performance/fast-boot.robot
miczyg1: Extended the commit message: 8613c268

dasharo-compatibility/efibootmgr.robot
miczyg1: Yeah, got a bit ahead of myself here. Changed: b8e80db2

scripts/regression-rerun-failed.sh
miczyg1: 6c8084c1

dasharo-compatibility/odroid-netcard.robot
miczyg1: Dropped the commit that removed flash stub.

philipanda: > That is why I decided for it to be separate.

> The way it is tested resembles ETHPERF, but it just uses a performance test to prove compatibility 

Yet it's name's "Check Performance of..."?
While I perfectly understand making one test for one thing, I hate how much time that often wastes in regressions. The main difference in the suites is the name. Why couldn't we be assuming, that if the netcard performance is fine, then the netcard must surely work properly?

lib/sleep-lib.robot
philipanda: Maybe you're using some linter and the `terminal` lib was not imported, so it changed it to a similar keyword from `keywords.robot`?

dasharo-compatibility/odroid-netcard.robot
miczyg1: > While I perfectly understand making one test for one thing, I hate how much time that often wastes in regressions. The main difference in the suites is the name. Why couldn't we be assuming, that if the netcard performance is fine, then the netcard must surely work properly?

I guess we can. Currently the only problem is that ETHPERF tests can verify only 1 ETH pair. But I need to test 2 pairs. Any suggestions how to do this so we could resolve this thread? @philipanda

miczyg1: Also it seems the check for empty list of ETH ports fails:

image

philipanda: >  Currently the only problem is that ETHPERF tests can verify only 1 ETH pair. But I need to test 2 pairs. Any suggestions how to do this so we could resolve this thread?

I think it would be enough to add more pairs to this `${ETH_PERF_PAIR_10_G}`  variable and iterate over pairs in the list
```robot
    ${eth_1}=    Get From List    ${ETH_PERF_PAIR_10_G}    0
    ${eth_2}=    Get From List    ${ETH_PERF_PAIR_10_G}    1
 ```
 something like:
 ```
 for i=0; i<len(ETH_PERF_PAIR_10_G); i+=2 {
     perform_test(ETH_PERF_PAIR_10_G[i], ETH_PERF_PAIR_10_G[i+1]
 }
 ```
 
 The `${target_bitrate}` which is currently hardcoded in the test could also get parametrized. One ETHPERF test uses `2.35`, just like your `NETCARD`, but another one uses `9.35` for no reason apparent to me

Comment (philipanda) [reply dasharo-compatibility/odroid-netcard.robot]: > Also it seems the check for empty list of ETH ports fails:

Seems like an issue caused by the quotes surrounding the variables. Their values are expanded into literals before being passed as a string parameter to `Depends On`. This string representation, which contains the additional quotes is then `Evaluate`'d, which fails, because the pairing of the quotes is now broken as i your screenshot. Any test using `Depends On` previously has been passing plain variables, not logic expressions, so it worked. I think removing the quotes should fix this bug
```robot
    Depends On    ${ETH_PORTS} != @{EMPTY}
```

miczyg1: Yeah I have fixed ETH_PORTS array the way you explained. I wonder how it even worked before.

miczyg1: @philipanda moved the net card tests to ETHPERF: 16483ed4
Created a new case because the first one is semiauto, while ODROID with netcard will always be able to do fully automated ETHPERF test.

But if you ask me what should be where in this testing framework, I would definitely not add "performance" test to compatibility suite (we have performance suite). And certainly not add `platform-stability.robot` to performance suite (we have stability suite). Someone's had a really good decision process here.

philipanda: Does ODROID ULTRA have other network interfaces that we would like to test other than the Net Card 2 ones?
If `ETHPERF003` and `ETHPERF001` tests are mutually exclusive, I'd definitely merge them. Otherwise, why not create separate tests for ODROID ULTRA for every ethernet-related suite?

I think the `NETCARD` tests were not pushed to osfv-dashboard, so a simple replacement in `test_cases.json` should also be fine instead of using `changed_to` fields.

dasharo-compatibility/network-speed.robot
```diff
@@ -76,12 +77,19 @@ ETHPERF001.201 Check Performance of 2.5G Wired Network Interface (Ubuntu)
         Lists Should Be Equal    ${manual_eth_restore}    ${dut_restore_values}
     END
 
+    # ODROID
+    IF    ${ODROID_NETCARD_SUPPORT}
+        # Test second pair of network ports on netcard
+        Configure Network Interfaces For Testing    enp6s0    enp7s0
```
philipanda: Why hardcode the interface names?

dasharo-compatibility/device-detection.robot
```diff
@@ -21,13 +21,19 @@ Suite Setup         Run Keywords
...
 Suite Teardown      Run Keywords
-...                     Skip If    '${SUITE_STATUS}' == 'SKIP'    Skipping Teardown since Suite was skipped as well
-...                     AND
 ...                     Log Out And Close Connection
 ...                     AND
+...                     Skip If    '${SUITE_STATUS}' == 'SKIP'    Skipping Teardown since Suite was skipped as well
+...                     AND
 ...                     Set UEFI Option    UsbDriverStack    Enabled
+...                     AND
+...                     Log Out And Close Connection
```
philipanda, line 34: Won't double `Log Out And Close Connection` break the teardown? It's already there a couple lines above

dasharo-compatibility/network-speed.robot
miczyg1: Removed: 8ac0fa8a

dasharo-compatibility/device-detection.robot
miczyg1: Good catch. Probably bad rebase from me. I have to fix it. But notice how the close connection happens before `Set UEFI Option`. The keyword had no chance to pass after closing connections..

dasharo-compatibility/odroid-netcard.robot
miczyg1: Merged into ETHPERF001: 8ac0fa8a

dasharo-compatibility/device-detection.robot
miczyg1: dee1fbc0

dasharo-compatibility/odroid-netcard.robot
philipanda: Great, that's much better :+1:

philipanda (approved):

PR #887, opened 2025-06-13
CI: Save robot logs even if it fails

philipanda:
I think the logs from fails are much more important than from passes...


macpijan (approved):
No idea how it works, but I assume you have checked it

PR #889, opened 2025-06-16
Test id ci

philipanda:


philipanda:
@krystian-hebel After reviewing this PR, force-push removing the commit named ' [REMOVE ME] Test CI triggering on changes '. It shouldn't be merged. It's only used to trigger the CI in this PR. With no changes to any `.robot file` nor the `test_cases.json` the CI shouldn't be triggered.

.github/workflows/test_ids_check.yml
```diff
@@ -0,0 +1,43 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+name: Test IDs check
+
+# Only try to run if a robot file or the test cases mapping file were modified
+on:
+  pull_request:
+    paths:
+      - '**.robot'
```
krystian-hebel: The scripts only consider `dasharo-*` directories, perhaps CI should also.

At some point Zarhus may want to have their tests checked as well, but as of now there is no clear distinction as to who should add them, whether it would use the same dashboard, and by extension, if the same `test_cases.json` file should contain both Dasharo and Zarhus tests.

.github/workflows/test_ids_check.yml
```diff
@@ -0,0 +1,43 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repository and submodules
+      uses: actions/checkout@v4
+      with:
+        submodules: 'recursive'
```
krystian-hebel: Not sure how much time it would actually save, but for this check submodules aren't used.

.github/workflows/test_ids_check.yml
philipanda: https://github.com/Dasharo/open-source-firmware-validation/compare/9b1eea4455261bf6bd6dbd174c95878ab29d2d0d..a3b917b7868ccb3c692f61c83bc8d1310ef8f5d3#diff-7689e88fbb7acba23993f9e385d8091fd023233d19b95984e76ad9646c82ac7aR11-R14

.github/workflows/test_ids_check.yml
philipanda: https://github.com/Dasharo/open-source-firmware-validation/compare/9b1eea4455261bf6bd6dbd174c95878ab29d2d0d..a3b917b7868ccb3c692f61c83bc8d1310ef8f5d3#diff-7689e88fbb7acba23993f9e385d8091fd023233d19b95984e76ad9646c82ac7aL21-L22

philipanda:
@krystian-hebel
I've moved all of the checks to pre-commit as it is run on every PR anyway.
Pre-commit now also checks if the IDs look like expected and if the OS name is included in tests performed on OSes - so a `*.\.201` must contain `(Ubuntu)` in its name.

I am not happy that the OS ids are hardcoded in the scripts. Maybe we should modify the `os-config/environment-test-ids.py` so that it defines a friendly name for each OS to be used in test names? Then all of the envs could be read from that file instead of being hardcoded

scripts/ci/test_ids/test_id_checking_lib.py
```diff
@@ -0,0 +1,82 @@
...
+    finder = TestCasesFinder()
+    testsuite.visit(finder)
+
+    list_of_tests = finder.tests
+
+    return list_of_tests
+
+
+def get_id(test):
+    res = re.search("^([A-Z]{0,9}[0-9]{3}\\.[0-9]{3}).*", test.name)
```
krystian-hebel: I don't think we want to have shorter identifiers.

```suggestion
    res = re.search("^([A-Z]{3,9}[0-9]{3}\\.[0-9]{3}).*", test.name)
```

scripts/ci/test_ids/test_id_checking_lib.py
```diff
@@ -0,0 +1,82 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+import difflib
+import re
+import subprocess
+
+from robot.model import SuiteVisitor
+from robot.running import TestSuiteBuilder
+
+# Maybe use os-config/environment-test-ids.py instead???
```
krystian-hebel: I think this would be better idea, especially since that file is already mentioned in the documentation. We would also need the same tests for firmware, and in the future for other OSes than just those below.

krystian-hebel (commented):
I tried running `pre-commit run --all` which caused ShellCheck to complain about a line in my script, namely line 18 in `list-tests-from-robot.sh`. Changing that line to `mapfile -t all < <(sort <<<"${all[*]}")` silences the error and doesn't change the overall result of further tests. If you want, feel free to change it in this PR, if not, I can do this in a follow-up.

Overall the solution is sound and tests what must be tested, but it would be best to use `environment-test-ids.py` as you noted. We shouldn't expect developers to dig into CI scripts in search for proper name.

scripts/ci/test_ids/test_id_checking_lib.py
```diff
@@ -0,0 +1,82 @@
...
+        "./scripts/list-tests-from-json.sh", capture_output=True, text=True
+    )
+    diff = difflib.unified_diff(
+        str.splitlines(robot.stdout),
+        str.splitlines(json.stdout),
+        fromfile="robot",
+        tofile="json",
+        lineterm="",
+    )
+    return [d for d in diff if not d.startswith(" ")]
```
krystian-hebel: There are also lines that start with `@`, like this one:

```
@@ -1,55 +1,110 @@
```

Do you think these should also be filtered out? IDs are sorted, and line references may be misleading when there are different and identical lines intertwined.

scripts/ci/test_ids/test_id_checking_lib.py
philipanda: Added the friendly names here:
https://github.com/Dasharo/open-source-firmware-validation/pull/889/
commits/907bc426c3934cac57ca5d803a44051f34b1b9d3

Parsing the os-config:
https://github.com/Dasharo/open-source-firmware-validation/pull/889/commits/9dcf2825239c525968d1233a28dd49b1e491cb2f

scripts/ci/test_ids/test_id_checking_lib.py
philipanda: Yes, they do look misleading. Let's remove them

philipanda: https://github.com/Dasharo/open-source-firmware-validation/compare/2f4d31d8a6c7af594ef354366f9d7f62d3fb2f68%5E..2f4d31d8a6c7af594ef354366f9d7f62d3fb2f68

philipanda:
@krystian-hebel When I run pre-commit locally, the last check is not triggered:
```text
$ pre-commit run --from-ref origin/develop --to-ref origin/test_id_ci
check for added large files..............................................Passed
check for merge conflicts................................................Passed
check for broken symlinks............................(no files to check)Skipped
detect private key.......................................................Passed
fix end of files.........................................................Passed
trim trailing whitespace.................................................Passed
mixed line ending........................................................Passed
codespell................................................................Passed
markdownlint.............................................................Passed
markdownlint-fix.........................................................Passed
ShellCheck v0.10.0.......................................................Passed
isort (python)...........................................................Passed
black....................................................................Passed
robotidy.............................................(no files to check)Skipped
robocop..............................................(no files to check)Skipped
reuse lint-file..........................................................Passed
platform-configs.....................................(no files to check)Skipped
test-ids-check.......................................(no files to check)Skipped
test-os-ids-check....................................(no files to check)Skipped
test-ids-mappings....................................(no files to check)Skipped
```
I have no more ideas why it does trigger in this PR.

macpijan:
> @krystian-hebel When I run pre-commit locally, the last check is not triggered:

This is my local execution if that helps, that is how it should look like?

`pre-commit run  --all-files > out.txt`


```
check for added large files..............................................[42mPassed[m
check for merge conflicts................................................[42mPassed[m
check for broken symlinks............................(no files to check)[46;30mSkipped[m
detect private key.......................................................[42mPassed[m
fix end of files.........................................................[42mPassed[m
trim trailing whitespace.................................................[42mPassed[m
mixed line ending........................................................[42mPassed[m
codespell................................................................[42mPassed[m
markdownlint.............................................................[42mPassed[m
markdownlint-fix.........................................................[42mPassed[m
ShellCheck v0.10.0.......................................................[42mPassed[m
isort (python)...........................................................[42mPassed[m
black....................................................................[42mPassed[m
robotidy.................................................................[42mPassed[m
robocop..................................................................[42mPassed[m
reuse lint-file..........................................................[42mPassed[m
platform-configs.........................................................[42mPassed[m
test-ids-check.......................................(no files to check)[46;30mSkipped[m
test-os-ids-check....................................(no files to check)[46;30mSkipped[m
test-ids-mappings........................................................[41mFailed[m
[2m- hook id: test-ids-mappings[m
[2m- exit code: 1[m

--- robot
+++ json
-ACPI001.201 ACPI driver test (Ubuntu)
-ACPI001.202 ACPI driver test (Fedora)
+ACPI001.001 ACPI driver test (Ubuntu)
```

krystian-hebel:
@macpijan are those all reported differences? I've had few hundreds of them.

I think that with `--all-files` none of the last 3 checks should be skipped, and that was the case in previous version I tested.

philipanda:
> This is my local execution if that helps, that is how it should look like?

Ah! That actually helped a lot. I though the pre-commit.ci runs a `pre-commit run --from-ref origin/develop --to-ref origin/test_id_ci` but actually in this repo it is configured to run a `pre-commit run --all-files`. That means that its the first two rules that were malformed.
I've missed a dot in the regex, it was `dasharo-*/.*\.robot`, but should have been `dasharo-.*/.*\.robot`

philipanda:
> @macpijan are those all reported differences? I've had few hundreds of them.
> 
> I think that with `--all-files` none of the last 3 checks should be skipped, and that was the case in previous version I tested.

Yes, with --all-files all the checks should run as if every file in the repository was just added. That's the case now. I suppose you've checked the pre-commit before I've added the malformed file restrictions. Every test now runs, because our pre-commit-ci runs an `--all-files` run instead of `--from-ref <base> --to-ref <branch>`. 

If you locally run `pre-commit run -
-from-ref origin/develop --to-ref origin/test_id_ci` none of the tests added in this PR are tested, because no changes to the target files were done in this PR

macpijan:
> are those all reported differences? I've had few hundreds of them.

No, not all.

os-config/environment-test-ids.py
```diff
@@ -27,3 +27,11 @@
     ENV_ID_TRENCHBOOT: "trenchboot",
     ENV_ID_ESXI: "ESXi",
 }
+
+ENV_ID_OS_FRIENDLY_NAMES = {
```
krystian-hebel: I think we should drop `OS` from name and include strings for firmware, even though they are rarely used right now.

os-config/environment-test-ids.py
```diff
@@ -27,3 +27,11 @@
...
+    ENV_ID_UBUNTU: "Ubuntu",
+    ENV_ID_FEDORA: "Fedora",
+    ENV_ID_WINDOWS: "Windows",
+    ENV_ID_TRENCHBOOT: "Trenchboot",
```
krystian-hebel: The proper spelling is `TrenchBoot`.

scripts/ci/test_ids/test_id_checking_lib.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def visit_test(self, test):
+        self.tests.append(test)
+
+
+class OsConfig:
+    def __init__(self):
+        """Parse the os-config file contents into a dictionary"""
+        self.os_ids = {}
+        self.bootmenu_names = {}
```
krystian-hebel: IIUC this is unused now, so please remove.

krystian-hebel (changes requested):

os-config/environment-test-ids.py
philipanda: https://github.com/Dasharo/open-source-firmware-validation/compare/fcf3d5675ea8eeba43b30781a79987b98fee76a6..3131d4ee55b29516be58311bc6e321890a50716a

os-config/environment-test-ids.py
philipanda: https://github.com/Dasharo/open-source-firmware-validation/compare/fcf3d5675ea8eeba43b30781a79987b98fee76a6..3131d4ee55b29516be58311bc6e321890a50716a

scripts/ci/test_ids/test_id_checking_lib.py
philipanda: https://github.com/Dasharo/open-source-firmware-validation/compare/3131d4ee55b29516be58311bc6e321890a50716a..7ce0820c9891bd7c748f9abc9395cfd895c3d3bc

scripts/ci/test_ids/test_id_checking_lib.py
```diff
@@ -0,0 +1,148 @@
...
+                and len(node.targets) == 1
+                and isinstance(node.targets[0], ast.Name)
+            ):
+                name = node.targets[0].id
+                vars[name] = _parse_ast_node(node.value)
+
+        for name, value in vars.items():
+            if name == "ENV_ID_OS_FRIENDLY_NAMES":
+                self.friendly_names = value
+            else:
```
krystian-hebel:
```suggestion
            elif not isinstance(value, dict):
```

Otherwise `os_id` in line 84 is unhashable and can't be compared against `friendly_names` key.

krystian-hebel (commented):
I still don't understand why the tests are run multiple times, e.g. there are 4 `+++ json` lines in CI log, and 15 when running `pre-commit run --all-files` locally. Anyway, assuming that these tests will eventually pass, it isn't a problem that needs to be resolved here and now.

krystian-hebel (approved):

philipanda:
> I still don't understand why the tests are run multiple times, e.g. there are 4 +++ json lines in CI log, and 15 when running pre-commit run --all-files locally. 

I'll investigate that and fix it before we merge it

philipanda:
@krystian-hebel 
Pre commit is passing the filenames to the hooks. The thing is there must be some config option that makes it pass them in portions. On GH it executes the scripts 4 times with N/4 files passes. Locally on my device it runs the scripts 15 times with smaller portions of filenames each. The solution could be to either disable this splitting, or just tell pre-commit that we don't want the filenames at all, which is the case here.
This commit does exactly that: https://github.com/Dasharo/open-source-firmware-validation/pull/889/commits/dfa1909666fa17ae8e63e60ff9db358742bb11d5

krystian-hebel (approved):

macpijan:
@philipanda We hold-off with merging until we clean up IDs and make this pre-commit check pass, right?

philipanda:
> @philipanda We hold-off with merging until we clean up IDs and make this pre-commit check pass, right?

Yes, if we do now, then every PR will have failed CI checks. The idea is to branch off these changes and fix all the IDs there, so that any active PRs will only need to fix their changes to have the CI pass

PR #892, opened 2025-06-16
Draft: Deprecated ids

krystian-hebel:
**Note before merging:** CFC001.002 has to be manually removed on server.


krystian-hebel:
Todos:
- manual tests - whole suites and specific OSes (Qubes OS, XCP-NG)
- missing OS names - can be automated
- AUD - deprecation, unclear
- CBP - not used in releases, dev-tests?
- DIO - inconsistent numbering
- DMI - missing Fedora tests in the database
- ECR - ID clashes
- MBO/MNE FUBAR
- RTD - reset to X instead of reset to default, should be changed on OSFV side?
- USB - deprecation, unclear
- UTC - deprecation
- VBO - multiple tests with the same ID in OSFV (semi-auto)
- WBS - disable became enable and vice versa, explain

These changes increased number of synchronized cases from 842 to 933. A large part of remaining differences comes from manual tests, and some more from tests without OS in name.

macpijan:
We still wanted to fill in missing OS info here

PR #895, opened 2025-06-18
workflows/pre-commit.yml: Add

philipanda:


philipanda:
I wonder why is not `pre-commit.ci` running the shellcheck. I thought it runs `--all-files`, but it clearly is not the case.
The markdown related checks are performed, but the shellcheck is not. In that case maybe using GH actions would be better because at least we are sure what exactly it runs?

macpijan:
https://github.com/Dasharo/open-source-firmware-validation/blob/develop/.pre-commit-config.yaml#L13

Yes, in this case the GH actions wins.

macpijan:
I disabled the repo in pre-commit.ci, let's test the GH actions approach for pre-commit then.

PR #900, opened 2025-06-24
202-packages.yaml: Update ffmpeg package name

philipanda:
It's now called `ffmpeg-free`. You can still install it manually using `dnf install ffmpeg`, which will result in installing `ffmpeg-free`, but when running it via ansible the old name won't work.


macpijan (approved):

macpijan:
@philipanda I encourage you to insert these useful information not only in GH PR but also in the commit body. This way it will be part of the git repository, not matter what happens with GitHub.

PR #901, opened 2025-06-24
Flashing ready-to-use disk images onto devices

philipanda:


util/preseeds.robot
```diff
@@ -0,0 +1,135 @@
...
+# Test suite used to quickly flash disk images with preinstalled OSes.
+# Two environment variables must be set prior to running it:
+# - SOURCE_IMAGE
+# - TARGET_DISK
+# Optional:
+# - CLONEZILLA_TTY
+#
+# Example usage:
+# - Flash the `1_windows_ubuntu` image onto the `nvme0n1` disk on the device using ${RTE_IP}:
+#    SOURCE_IMAGE=1_windows_ubuntu TARGET_DISK=nvme0n1 ./scripts/run.sh util/preseeds.robot -- -t "Restore Disk*"
```
macpijan, line 30: @philipanda It is the disk path as seen in the clonezilla OS?

Are there also other ways of pointing to the target disk?

These paths are not fully reproducible, especially if we have multiple disk of the same type (e.g. 2xNVMe disk, or SATA disk + some USB drives).

util/preseeds.robot
```diff
@@ -0,0 +1,135 @@
...
+
+*** Test Cases ***
+Restore Disk Clonezilla
+    ${clonezilla_tty}=    Get Envvar    CLONEZILLA_TTY    ${TRUE}
+    ${source_image}=    Get Envvar    SOURCE_IMAGE
+    ${target_disk}=    Get Envvar    TARGET_DISK
+
+    Power On
+    ${ipxe_entered}=    Run Keyword And Return Status    Enter IPXE
+    IF    not ${ipxe_entered}    # It might just be disabled
```
macpijan, line 56: We also have `Make Sure That Network Boot Is Enabled` - maybe it could be used here prior entering iPXE?

util/preseeds.robot
```diff
@@ -0,0 +1,135 @@
+*** Settings ***
```
macpijan, line 1: This test/automation flow does not have much in common with preseeds. Maybe just call it `clonezilla` or `restore-disk`, or something like that?

util/preseeds.robot
```diff
@@ -0,0 +1,135 @@
...
+#
+# Example usage:
+# - Flash the `1_windows_ubuntu` image onto the `nvme0n1` disk on the device using ${RTE_IP}:
+#    SOURCE_IMAGE=1_windows_ubuntu TARGET_DISK=nvme0n1 ./scripts/run.sh util/preseeds.robot -- -t "Restore Disk*"
+# - The same, but redirect the Clonezilla's output onto ttyS0 if video output is not available:
+#    SOURCE_IMAGE=1_windows_ubuntu TARGET_DISK=nvme0n1 CLONEZILLA_TTY=ttyS0 ./scripts/run.sh util/preseeds.robot -- -t "Restore Disk*"
+# - Launch clonezilla in manual mode over chosen serial console (for restore or upload):
+#    CLONEZILLA_TTY=ttyUSB0 ./scripts/run.sh util/preseeds.robot -- -t "Manual*"
+#
+#    Important: Don't run the whole test suite. Every test case is a
```
macpijan, line 36: After going through this, I think it's a bit counterintuitive and might cause problems by not following docs or this instruction precisely.

I think it'd be less error-prone and easier to use if we split it into something like `restore-disk-auto.robot` and `restore-disk-manual.robot`

macpijan:
[preseeds_log.html](https://github.com/user-attachments/files/21900781/preseeds_log.html)

I think I am running into rather unexpected error when trying to run on VP2430. Am I missing something here?

util/preseeds.robot
```diff
@@ -0,0 +1,135 @@
...
+Resource        ../keys-and-keywords/ubuntu-keywords.robot
+
+Suite Setup     Run Keywords
+...                 Prepare Test Suite
+# Test suite used to quickly flash disk images with preinstalled OSes.
+# Two environment variables must be set prior to running it:
+# - SOURCE_IMAGE
+# - TARGET_DISK
+# Optional:
+# - CLONEZILLA_TTY
```
macpijan, line 26: Is it optional or mandatory? My test was skipped if not given.

macpijan: [Uploading preseeds_log.html…]()

macpijan:
> I think I am running into rather unexpected error when trying to run on VP2430. Am I missing something here?

Setting new venv did not help.

But rebasing, and then setting new venv - did help.

util/preseeds.robot
macpijan: Ok I think I know how it supposed to work now, put it puzzled me at first.

For me, if test was skipped, it was not executed, so I was confused and did not read the warning properly.

Make we should pass here and/or make this message more verbose (that the cloning process is in progress, and must be monitored via other means)?

util/preseeds.robot
philipanda: >  It is the disk path as seen in the clonezilla OS?

Not a full path, only the device name as seen by a Linux. If unsure it might be needed to boot some live image and check the disk name. Alternatively you can use the `Manual*` "test case" instead of `Restore*`, enter the shell in clonezilla, and check the name. 

I haven't seen any way to choose the target disk other than by this name in the clonezilla docs.

util/preseeds.robot
philipanda: All `Make Sure That Network Boot Is Enabled` does is just call `Set UEFI Option`, and it fails silently instead of FAILing if it is not possible, so it's not equivalent.

I think a better alternative would be to extend `Enter IPXE` a bit so that it performs this check and optionally enables Network Boot if disabled.

util/preseeds.robot
philipanda: That makes sense

philipanda:
> [preseeds_log.html](https://github.com/user-attachments/files/21900781/preseeds_log.html)
> 
> I think I am running into rather unexpected error when trying to run on VP2430. Am I missing something here?

Old version of our robotframework fork with a bug. Either this branch needs a rebase. or you've used an old venv.

util/preseeds.robot
macpijan: > Not a full path, only the device name as seen by a Linux

Path or name - whatever it's basically the same. What I am simply saying is that these are not repeatable in Linux, and if having multiple drives, we can have one disk named sda, the other time - sdb. Etc. If there is no way around that, we have to stick to that and keep in mind that at some point it maybe cause some issue. Perhaps not that probbale in our setups.

util/preseeds.robot
macpijan: Let's do it or at least note it down and link here.

philipanda: https://github.com/Dasharo/open-source-firmware-validation/pull/901/commits/b37f7ba4b350e494e2037363cf6b8f948f530ea7
The change shouldn't collide with existing usages of this keyword, as it only does something different if entering the IPXE menu fails

util/preseeds.robot
philipanda: 86bca8257bd00ac9d4990ae0037c44c4dd0013d0

util/preseeds.robot
philipanda: https://github.com/Dasharo/open-source-firmware-validation/pull/901/commits/30c1db3aa02d72b173fc84c191b22f28168e8278

util/preseeds.robot
philipanda: Added a manual step so that a popup will appear if clonezilla is run without a tty in https://github.com/Dasharo/open-source-firmware-validation/pull/901/commits/e503c25b019acfd4d2c5df97bf1ea1dea96b7a1f

Note that this does not stop the installation, it will run on the target device no matter the popup.
Made the message a bit more meaningful here: https://github.com/Dasharo/open-source-firmware-validation/pull/901/commits/22ecef038762e3dc17f707175b78cba8bdd575bc

macpijan: Thanks, that will be much more obvious.

macpijan:
LGTM. But pre-commit is unhappy ?

philipanda:
@macpijan I've made the pre-commit happy

macpijan (approved):

PR #903, opened 2025-06-25
Self tests separate stress tests

philipanda:
The stress tests which take more than half of the total self-tests time were separated into a new workflow, that runs every day on midnight instead of on every PR/push to save a whole lot of time, as stress-testing on every change shouldn't be necessary


macpijan (approved):

macpijan:
Pushed fix to the pre-commit issue: 1316b63048ef4eebf816ae1918c2c76ec47c1617

PR #908, opened 2025-06-27
Fix nightly stress tests fails

philipanda:
The tests were failing because robot was run on single files instead of the whole module, so it produced an error if no `stress-test` tag was found in the selected file. Using the `run.sh` wrapper the results will be closer to what a tester would experience and the self tests scripts can be simplified because they were doing pretty much the same thing as run.sh.

The qemu-self-tests.sh script was updated too to work the same way.

### DIFF
--- a/.github/workflows/qemu-stress-tests.yml
+++ b/.github/workflows/qemu-stress-tests.yml
@@ -1,12 +1,15 @@
-# SPDX-FileCopyrightText: 2024 3mdeb <contact@3mdeb.com>
+# SPDX-FileCopyrightText: 2025 3mdeb <contact@3mdeb.com>
 #
 # SPDX-License-Identifier: Apache-2.0
 
-name: Keywords self-tests with QEMU
+name: Keywords stress-tests with QEMU
 
 on:
   schedule:
     - cron: '0 0 * * *' # Every day at midnight
+  workflow_dispatch:
+  pull_request:
+
 
 jobs:
   qemu:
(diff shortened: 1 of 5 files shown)


philipanda:
Action runs with the changes:
https://github.com/Dasharo/open-source-firmware-validation/actions/runs/15928202030/job/44930547540?pr=908
https://github.com/Dasharo/open-source-firmware-validation/actions/runs/15928202041/job/44930549220?pr=908

I've run the stress test by making it run on pull_requests. It won't run on this PR now, because it should only run once per night.

macpijan (approved):

.github/workflows/qemu-stress-tests.yml
```diff
@@ -1,12 +1,15 @@
...
 # SPDX-License-Identifier: Apache-2.0
 
-name: Keywords self-tests with QEMU
+name: Keywords stress-tests with QEMU
 
 on:
   schedule:
     - cron: '0 0 * * *' # Every day at midnight
+  workflow_dispatch:
+  pull_request:
```
macpijan, line 11: @philipanda Do we want to leave both schedule and pull_request here?

philipanda: No, that's a mistake, `pull_request` should only be on the normal "quick" self-tests

philipanda: @macpijan 
https://github.com/Dasharo/open-source-firmware-validation/pull/917

macpijan: I thought so, but realized only after merging. Fixed her: faf7690788d12db1f40a7584e8a6ce115aa9e4bf

PR #911, opened 2025-06-27
qemu-selftest workflows: Only run on relevant changes

philipanda:
also added  `workflow_dispatch` to allow running them manually if someone would want to


macpijan:
Thanks @philipanda that is a very nice addition

Issue #910, opened 2025-06-27
Add option to skip manual/semi-manual tests in regression.sh


BeataZdunczyk:
### The problem you're addressing (if any)

We have been gradually creating all test cases, including manones ones into the osfv repo. Soon, we will also migrate any remaining manual tests from docs.dasharo.com to ensure consistency and to maintain all test cases in one place, osfv repo

However, we still want to be able to run regression tests without failures caused by manual or semi-manual test cases, so we need a way to skip them during execution.

### Describe the solution you'd like

Add an option to skip manual/semi-manual tests during the execution of regression.sh, possibly using tags.
This could be implemented similarly to how test IDs are handled, e.g.:
https://github.com/Dasharo/open-source-firmware-validation/blob/test_id_ci/dasharo-security/verified-boot.robot#L137

### Where is the value to a user, and who might that user be?

Be able to run regression tests without failures caused by manual or semi-manual test cases.

### Describe alternatives you've considered

-

### Additional context

-

PR #914, opened 2025-06-30
Test tags

wiktormowinski:
aims to resolve: http://github.com/Dasharo/open-source-firmware-validation/issues/910

### DIFF
--- a/dasharo-compatibility/audio-subsystem.robot
+++ b/dasharo-compatibility/audio-subsystem.robot
@@ -15,7 +15,7 @@ Resource            ../keys.robot
 Suite Setup         Prepare Test Suite
 Suite Teardown      Log Out And Close Connection
 
-Force Tags          automated
+Default Tags        automated
 
 
 *** Variables ***
--- a/dasharo-compatibility/audio-subsystem.robot
+++ b/dasharo-compatibility/audio-subsystem.robot
@@ -58,6 +58,7 @@ AUD002.201 Internal Audio playback
     ...    toplayback audio recordings. To do so, first determine presence
     ...    of audio sink. Audio sink must not be a dummy. After it was
     ...    verified, we verify that sound is not malformed.
+    [Tags]    semiauto
     Skip If    not ${TESTS_IN_UBUNTU_SUPPORT}    ${TEST_NAME} not supported
     Skip If    '${ENV_ID_UBUNTU}' not in ${TESTED_LINUX_DISTROS}    ${TEST_NAME} not supported
     Skip If    not ${INTERNAL_AUDIO_SUPPORT}    ${TEST_NAME} not supported
@@ -73,6 +74,7 @@ AUD003.201 Internal Audio capture
     [Documentation]    Check whether the audio subsystem is able to capture
     ...    audio on Ubuntu. To do so, we first determine presence of internal
     ...    capture device.
+    [Tags]    semiauto
     Skip If    not ${TESTS_IN_UBUNTU_SUPPORT}    ${TEST_NAME} not supported
     Skip If    '${ENV_ID_UBUNTU}' not in ${TESTED_LINUX_DISTROS}    ${TEST_NAME} not supported
     Skip If    not ${INTERNAL_AUDIO_SUPPORT}    ${TEST_NAME} requires internal Microphone
@@ -96,6 +98,7 @@ AUD004.201 External headset recognition
 AUD005.201 External headset audio playback
     [Documentation]    Check whether Ubuntu has capability to playback
     ...    sounds via external headset.
+    [Tags]    semiauto
     Skip If    not ${TESTS_IN_UBUNTU_SUPPORT}    ${TEST_NAME} not supported
     Skip If    '${ENV_ID_UBUNTU}' not in ${TESTED_LINUX_DISTROS}    ${TEST_NAME} not supported
     Skip If    not ${EXTERNAL_HEADSET_SUPPORT}    ${TEST_NAME} not supported
@@ -111,6 +114,7 @@ AUD005.201 External headset audio playback
 AUD006.201 External headset audio capture
     [Documentation]    Check whether Ubuntu has capability to capture sound
     ...    via external headset.
+    [Tags]    semiauto
     Skip If    not ${TESTS_IN_UBUNTU_SUPPORT}    ${TEST_NAME} not supported
     Skip If    '${ENV_ID_UBUNTU}' not in ${TESTED_LINUX_DISTROS}    ${TEST_NAME} not supported
     Skip If    not ${EXTERNAL_HEADSET_SUPPORT}    ${TEST_NAME} not supported
@@ -147,6 +151,7 @@ AUD002.202 Internal Audio playback
     ...    toplayback audio recordings. To do so, first determine presence
     ...    of audio sink. Audio sink must not be a dummy. After it was
     ...    verified, we verify that sound is not malformed.
+    [Tags]    semiauto
     Skip If    '${ENV_ID_FEDORA}' not in ${TESTED_LINUX_DISTROS}    ${TEST_NAME} not supported
     Skip If    not ${INTERNAL_AUDIO_SUPPORT}    ${TEST_NAME} not supported
     Audio Subsystem Detection Linux    ${ENV_ID_FEDORA}
@@ -160,6 +165,7 @@ AUD003.202 Internal Audio capture
     [Documentation]    Check whether the audio subsystem is able to capture
     ...    audio on Fedora. To do so, we first determine presence of internal
     ...    capture device.
+    [Tags]    semiauto
     Skip If    '${ENV_ID_FEDORA}' not in ${TESTED_LINUX_DISTROS}    ${TEST_NAME} not supported
     Skip If    not ${INTERNAL_AUDIO_SUPPORT}    ${TEST_NAME} requires internal Microphone
     Audio Subsystem Detection Linux    ${ENV_ID_FEDORA}
@@ -181,6 +187,7 @@ AUD004.202 External headset recognition
 AUD005.202 External headset audio playback
     [Documentation]    Check whether Fedora has capability to playback
     ...    sounds via external headset.
+    [Tags]    semiauto
     Skip If    '${ENV_ID_FEDORA}' not in ${TESTED_LINUX_DISTROS}    ${TEST_NAME} not supported
     Skip If    not ${EXTERNAL_HEADSET_SUPPORT}    ${TEST_NAME} not supported
     Audio Subsystem Detection Linux    ${ENV_ID_FEDORA}
@@ -195,6 +202,7 @@ AUD005.202 External headset audio playback
 AUD006.202 External headset audio capture
     [Documentation]    Check whether Fedora has capability to capture sound
     ...    via external headset.
+    [Tags]    semiauto
     Skip If    '${ENV_ID_FEDORA}' not in ${TESTED_LINUX_DISTROS}    ${TEST_NAME} not supported
     Skip If    not ${EXTERNAL_HEADSET_SUPPORT}    ${TEST_NAME} not supported
     Audio Subsystem Detection Linux    ${ENV_ID_FEDORA}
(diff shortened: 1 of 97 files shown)


scripts/regression.sh
```diff
@@ -21,10 +21,10 @@ export _REGRESSION_RUN
 
 check_test_station_variables
 
-if [ -z "$NO_SETUP" ]; then
-    execute_robot "util/basic-platform-setup.robot" "${@}"
-fi
+#if [ -z "$NO_SETUP" ]; then
```
philipanda: Did you forget to restore it?

wiktormowinski: exactly https://github.com/Dasharo/open-source-firmware-validation/pull/914/commits/5544c3557de6f001acf7af09292451117fd0c2c8

dasharo-compatibility/audio-subsystem.robot
```diff
@@ -15,6 +15,8 @@ Resource            ../keys.robot
 Suite Setup         Prepare Test Suite
 Suite Teardown      Log Out And Close Connection
 
+Force Tags          automated
```
philipanda: Some of the tests in this suite are manual or semi auto. I can't add a comment to them because I can only comment changed lines. Example: `AUD002.201 Internal Audio playback`, the last line is `Log \Internal speakers detected, check validity of sound playback manually\n`

wiktormowinski: the way i decided on where should i put auto and where semiauto is the presence of a stopper such as Dialogs lib popup windows that wait for user input,

in the example you provided the manual check is a bit too subtle, tbh. As a result, the tester might not even realize there was an additional step that granted them a trust credit, with the hope that they would genuinely report the outcome based on that 'extra' step. 

I think we should make the manual steps to always enforce a tester's input.

dasharo-compatibility/cpu-fan-speed-measure.robot
```diff
@@ -21,6 +21,8 @@ Resource            ../lib/performance/cpu.robot
 Suite Setup         Prepare CPU Fan Speed Measure Suite
 Suite Teardown      Log Out And Close Connection
 
+Force Tags          automated
```
philipanda: 003, 004, 005 are semi manual

dasharo-compatibility/audio-subsystem.robot
philipanda: Yes, we should. I am 100% sure that a tester would say this test is PASS without looking. That is not in the scope of this PR though.
It would be great to find all such test cases and create an issue for correcting such steps

dasharo-compatibility/dasharo-tools-suite.robot
```diff
@@ -27,6 +27,7 @@ Test Setup          Run Keyword If    ${TESTS_IN_FIRMWARE_SUPPORT}
 DTS001.001 Booting DTS from USB works correctly
     [Documentation]    This test aims to verify that DTS is properly booting
     ...    from USB.
+    [Tags]    automated
```
philipanda: You can use `Default Tags` instead of `Force Tags` It only gives the tag if no `[Tags]` is present in a test case, so a single `Default Tags    automated` and `[Tags]    semi-auto` on semi auto tests would be a bit cleaner

dasharo-compatibility/display-ports-and-lcd-support.robot
```diff
@@ -27,6 +27,7 @@ DSP001.201 - Internal display in OS (Ubuntu)
     [Documentation]    Check whether an internal display is visible in
     ...    Ubuntu.
     ...    Previous IDs: DSP001.002
+    [Tags]    automated
```
philipanda: You can use `Default Tags` here too

dasharo-compatibility/dasharo-tools-suite.robot
wiktormowinski: oh ok, i like that idea. I was upset with current solution. i am going to swap all of these and add to missing suites

dasharo-compatibility/platform-hibernation-and-resume.robot
```diff
@@ -14,6 +14,8 @@ Resource            ../keys.robot
 Suite Setup         Prepare Test Suite
 Suite Teardown      Log Out And Close Connection
 
+Force Tags          automated
```
philipanda: What about the comment saying  that the test case has to be done manually?
https://github.com/Dasharo/dasharo-issues/issues/485

dasharo-compatibility/usb-type-c.robot
```diff
@@ -60,6 +60,7 @@ Suite Teardown      Run Keyword
 
 UTC107.201 USB Type-C Display output (Ubuntu) (ME: Enabled) (WL-UMD05 Pro Rev.E)
     [Documentation]    Check whether the DUT can detect the USB Type-C hub.
+    [Tags]    automated
```
philipanda: Isn't every test in this suite automated? The manual ones are commented out

wiktormowinski: there were some tagged as semiauto already

philipanda: Right, then let's use the `Default Tags` here

dasharo-performance/platform-stability.robot
```diff
@@ -24,7 +24,7 @@ Suite Setup         Run Keywords
 Suite Teardown      Run Keyword
 ...                     Log Out And Close Connection
 
-Force Tags          automated
+Default Tags        automated
```
philipanda, line 27: STB002.201 will not get the tag `automated`, because the Default Tags are overwritten by `[Tags]   minimal-regression`. Try this to search for possible scenarios like this:
`grep -rlF "Default Tags" dasharo-performance | xargs grep -lF "[Tags]"`

scripts/regression.sh
```diff
@@ -25,6 +25,6 @@ if [ -z "$NO_SETUP" ]; then
     execute_robot "util/basic-platform-setup.robot" "${@}"
 fi
 
-execute_robot "dasharo-compatibility" "${@}"
-execute_robot "dasharo-security" "${@}"
-execute_robot "dasharo-performance" "${@}"
+execute_robot "dasharo-compatibility" -- --include automated "${@}"
```
krystian-hebel, line 28: This works around a bug that I noticed some time ago but forgot to push fixes for. I tried to fix in in `execute_robot` to catch all invocations, not only these ones. Unless I'm mistaken, this was it, at least at that time:

```diff
diff --git a/scripts/lib/robot.sh b/scripts/lib/robot.sh
index 4705b7a41574..8076f79d037b 100644
--- a/scripts/lib/robot.sh
+++ b/scripts/lib/robot.sh
@@ -50,7 +50,7 @@ execute_robot() {
   # scope and additional arguments to robot which need to be separated
   # when calling robot
   for ((i=0;i<_args_len;i++)); do
-    if [[ ${_args[$i]} == *"--"* ]]; then
+    if [[ ${_args[$i]} == "--" ]]; then
       _separator_idx=$i
       break
     fi
@@ -63,6 +63,9 @@ execute_robot() {
   if [[ $_separator_idx -gt 0 ]]; then
     _separator_idx=$_separator_idx+1
     for ((i=_separator_idx;i<_args_len;i++)); do
+      if [[ ${_args[$i]} == "--" ]]; then
+        continue
+      fi
       # Some arguments may contain spaces. Bash removes quotation marks
       # from command arguments. Because we need to pass them again to
       # another command the quotation marks need to be restored or the arguments
```

Either copy this to `robot.sh` or add `--` to all calls to that functions, I know about this file and `run.sh`, but maybe there are other. Having a workaround in just one place isn't enough.

krystian-hebel (changes requested):

scripts/regression.sh
philipanda: What was the bug?

krystian-hebel: First option with `--` was ignored, regardless of whether double dash were the first two characters or somewhere inside the option name (or arg).

wiktormowinski: I tried this with `-- --include semiauto` and it ran only semiauto tests from `dasharo-compatibility`. Shouldn't it get ignored then?

philipanda: `--include` runs only the given tag, `--exclude` runs everything but the given tag. For regression we want `--include automated` so only automated tests are run.

krystian-hebel: Have you tried with just `--include semiauto`? It would ignore `--include` but leave `semiauto`.

philipanda: `--include semiauto` without the preceeding `--` is invalid. The `--` separates the names of test modules to be run, of which there might be an arbitrary amount, from arguments that shall be passed to `robot`. 
https://github.com/dasharo/open-source-firmware-validation?tab=readme-ov-file#running-tests-with-additional-arguments

wiktormowinski: Yeah, that’s the official version we have in this PR. But to verify that it actually works, I tested it with semiauto, since it’s much more noticeable when the regression.sh starts from the DTS suite.

wiktormowinski: @krystian-hebel @philipanda I meant -- --, not --. Thanks for the correction, gentlemen.

dasharo-compatibility/audio-subsystem.robot
macpijan: @wiktormowinski can you please create issue and link to this discussion as one of the spotted case of such a problem?

dasharo-compatibility/usb-type-c.robot
wiktormowinski: alright done https://github.com/Dasharo/open-source-firmware-validation/pull/914/commits/6a2022b2d85baf68cff9c3ead48c1aa804fa3bea

dasharo-performance/platform-stability.robot
philipanda: https://github.com/Dasharo/open-source-firmware-validation/pull/914/commits/23cb2b3430aa51671f25675ea29df8e0b57eebf4

dasharo-compatibility/cpu-fan-speed-measure.robot
philipanda: https://github.com/Dasharo/open-source-firmware-validation/pull/914/commits/6a2022b2d85baf68cff9c3ead48c1aa804fa3bea#diff-d2248e1805002aa05b033a79fc4738cce1636818db0129db3d5ee48f48be96e7R59-R85

dasharo-compatibility/audio-subsystem.robot
wiktormowinski: @macpijan  something like that?: https://github.com/Dasharo/open-source-firmware-validation/issues/928

dasharo-compatibility/platform-hibernation-and-resume.robot
wiktormowinski: https://github.com/Dasharo/open-source-firmware-validation/pull/914/commits/a9f9bdbe8670f95151eb5d16e520bfe0df04b038

dasharo-compatibility/dasharo-tools-suite.robot
wiktormowinski: https://github.com/Dasharo/open-source-firmware-validation/pull/914/commits/c8e16c0b6ac591ffdacb5681196afebfee5f356a

dasharo-compatibility/display-ports-and-lcd-support.robot
wiktormowinski: https://github.com/Dasharo/open-source-firmware-validation/pull/914/commits/47955a15638b4cc8a864d745345a6bafcc46f207

docs/adding-and-naming-test-cases.md
```diff
@@ -217,3 +217,29 @@ permissions. Example usage, assuming database server running on localhost:
 ```shell
 DB_SERVER_IP=127.0.0.1:5984 ./scripts/synchronize-db.py
 ```
+
+# Test Workflow Tag Guidelines
```
philipanda, line 221: I feel that the documentation starts to become a little messy. The tags are also mentioned in https://github.com/Dasharo/open-source-firmware-validation/blob/562d229d2675466dbf632a1a40001ef733b92ae5/docs/contributing.md?plain=1#L153, so I've been thinking whether the two topics shouldn't be kept together. It shouldn't be necessary for now.
Moving this section to the linked file wouldn't help make the documentation cleaner though. It probably needs to be restructured a bit, and that is not within the scope of this PR.

philipanda (approved):
I my opinion everything's good now.

docs/adding-and-naming-test-cases.md
wiktormowinski: yeah but i like to have this step in "test-adding-guide" so that any new test gets done 100% ready to go, no redos.

PR #918, opened 2025-07-03
test_cases.json: correct id update

wiktormowinski:


krystian-hebel (changes requested):
[RTFM](https://github.com/Dasharo/open-source-firmware-validation/blob/develop/docs/adding-and-naming-test-cases.md#synchronization-with-database) please :slightly_smiling_face:

> Each time a test name, ID or module is **modified**, the change must be reflected in `test_cases.json` file. This includes previous IDs, which are saved in `changed_to` key. Note that there is only one ID in that key. Each ID change results in _creation_ of a new object in JSON file, and _modification_ of the existing one by pointing to the new ID.

There are no new objects added in this PR.

krystian-hebel (approved):

PR #938, opened 2025-07-15
Ci on hw fixes

macpijan:
Missing flags on some platforms is a common failure during regression testing.

The goal here is to have a pre-commit which checks if the newly added flags are defined in default.robot.

That was the script execution result here:  https://paste.dasharo.com/?bea619207b960492#8Kr1hUr4D37CXoaC5KpPZpy5UFaZ4Ujqy9W7cyeKgxfp

### DIFF
--- a/platform-configs/odroid-h4-plus.robot
+++ b/platform-configs/odroid-h4-plus.robot
@@ -34,7 +34,6 @@ ${DCU_UUID_SUPPORT}=                            ${TRUE}
 ${DCU_SERIAL_SUPPORT}=                          ${TRUE}
 ${CUSTOM_LOGO_SUPPORT}=                         ${TRUE}
 ${DCU_SUPPORTED_BOOLEAN_SMMSTORE_VARIABLE}=     ${EMPTY}
-${ETH_PORTS}=                                   ${EMPTY}
 ${USB_DISKS_DETECTION_SUPPORT}=                 ${TRUE}
 ${USB_KEYBOARD_DETECTION_SUPPORT}=              ${TRUE}
 ${NVME_DISK_SUPPORT}=                           ${TRUE}
--- a/scripts/ci/check_platform_configs_vars.py
+++ b/scripts/ci/check_platform_configs_vars.py
@@ -11,6 +11,7 @@ import sys
 
 from robot.api import get_model
 from robot.api.parsing import Variable, VariableSection
+from robot.running import ResourceFileBuilder
 
 # Define paths
 BASE_DIR = "platform-configs"  # Adjust this if needed
@@ -20,15 +21,8 @@ DEFAULT_ROBOT = os.path.join(INCLUDE_DIR, "default.robot")
 
 def get_defined_variables(file_path):
     """Extract variable names from a Robot Framework file using RF parser."""
-    with open(file_path, encoding="utf-8") as f:
-        model = get_model(f.read())
-
-    variables = set()
-    for section in model.sections:
-        if isinstance(section, VariableSection):
-            for var in section.body:
-                if isinstance(var, Variable):
-                    variables.add(var.name)
+    variables = ResourceFileBuilder().build(file_path).variables
+    variables = set([var.name for var in variables])
     return variables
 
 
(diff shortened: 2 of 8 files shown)


platform-configs/odroid-h4-plus.robot
```diff
@@ -34,7 +34,6 @@ ${DCU_UUID_SUPPORT}=                            ${TRUE}
 ${DCU_SERIAL_SUPPORT}=                          ${TRUE}
 ${CUSTOM_LOGO_SUPPORT}=                         ${TRUE}
 ${DCU_SUPPORTED_BOOLEAN_SMMSTORE_VARIABLE}=     ${EMPTY}
-${ETH_PORTS}=                                   ${EMPTY}
```
macpijan, line 37: @miczyg1 I assume it should not break anything for odroid?
It is already defined in default.robot, and this was the only platform cofnig defining is as a single variable ($), not a list (@).

Which triggered detection by a newly added script.

scripts/ci/check_platform_configs_vars.py
```diff
@@ -0,0 +1,71 @@
...
+from robot.api import get_model
+from robot.api.parsing import Variable, VariableSection
+
+# Define paths
+BASE_DIR = "platform-configs"  # Adjust this if needed
+INCLUDE_DIR = os.path.join(BASE_DIR, "include")
+DEFAULT_ROBOT = os.path.join(INCLUDE_DIR, "default.robot")
+
+
+def get_defined_variables(file_path):
```
philipanda, line 22: This whole function could be replaced with `variables = ResourceFileBuilder().build(file_path).variables`, but that would require changing the rest of the script a bit and doesn't really help in any way than simplifying this a bit.

philipanda: Or maybe not, that works
```py
def get_defined_variables(file_path):
    """Extract variable names from a Robot Framework file using RF parser."""
    variables = ResourceFileBuilder().build(file_path).variables
    variables = set([var.name for var in variables])
    return variables
```

macpijan: Thanks.

064d05601013e6035147e3f5400d475d40be7f96

macpijan: Confirmed it still works by adding some random variable:

```
⚠ Variables defined in files but missing in include/default.robot:

platform-configs/novacustom-ns70mu.robot:
  ${TESTS_IN_XYZ}

📌 Unique variables to consider adding to include/default.robot:

*** Variables ***

${TESTS_IN_XYZ}=                           ${TBD}
```

platform-configs/odroid-h4-plus.robot
philipanda: It's set in `default.robot` to `${EMPTY}` too, so that change does not impact the tests

philipanda: `dasharo-compatibility/check-ethernet-ports-order.robot` checks if the variable is equal to `${EMPTY}`, and that one test suite will break.

philipanda (commented):
Everything looks good to me apart from this topic https://github.com/Dasharo/open-source-firmware-validation/pull/938/files#r2206975056

platform-configs/odroid-h4-plus.robot
macpijan: @philipanda I think this was wrong and defining it here as a variable (not a list) was a workaround for this specific platform.

Fixed here: b648101adb1f47083792ce1894e991cf512a225c

Tests are properly skipped for odroid:

```
Logs will be saved at logs/odroid-h4-plus/network-speed_2025_07_16_12_27_40
Watch "logs/odroid-h4-plus/network-speed_2025_07_16_12_27_40/network-speed_debug.log" to monitor the progress of the test
Network-Speed                                                                 
ETHPERF001.201 Check Performance of 2.5G Wired Network Interface (... | SKIP |
ETHPERF001.201 not supported
ETHPERF002.201 Check Performance of 10G Wired Network Interface (U... | SKIP |
ETHPERF002.201 not supported
The asset has been checked out automatically. It is now automatically checked in as well. If you want to keep using it, reserve it manually first.
Network-Speed                                                         | SKIP |
2 tests, 0 passed, 0 failed, 2 skipped
```

philipanda (approved):

PR #946, opened 2025-07-18
scripts/lib/robot.sh: Prevent running on dirty tree

philipanda:


scripts/lib/robot.sh
```diff
@@ -139,6 +139,16 @@ execute_robot() {
     installed_dut_option=""
   fi
 
+  # Prevent executing tests on a dirty git tree
+  if ! git diff --quiet; then
+    if [[ -z "${ALLOW_DIRTY}" ]]; then
+        echo "Git tree is dirty!"
+        echo "Commit your changes before running tests!"
```
krystian-hebel:
```suggestion
        echo "Commit and push your changes before running tests!"
```

Having unknown revision from local commit would be worse than `-dirty`. Even though it isn't tested, having it printed may be enough. I also don't think that it needs to be tested, in case CI or some other script does a shallow checkout that may leave the repo in detached HEAD state.

macpijan:
@philipanda What's the origin story of this?

I am afraid it would cause more workarounds and skipping usage of the run.sh

Better place to put it would be regression.sh, not for every single test.

scripts/lib/robot.sh
macpijan: Do you really commit and **push** all of your code (not just test code) prior doing any test on it?

philipanda:
> What's the origin story of this?

OSFV dashboard logs the OSFV revision used for running the tests. It would be incredible to be able to actually reproduce the historic test results, but as of now nothing prevents the test to be run on dirty tree with arbitrary modifications to OSFV. I know that @krystian-hebel strongly insisted before on enforcing that, and I agree with him.

> I am afraid it would cause more workarounds and skipping usage of the run.sh

That's why I've already added a workaround - setting the `ALLOW_DIRTY` environment variable. I think it would be an inconvenience small enough to not cause people creating workarounds, but if someone runs the tests on a dirty git tree, they'll be immediately notified. All someone has to do is commit their changes locally, even with a commit message like "WIP", and they're fine.

> Better place to put it would be regression.sh, not for every single test.

Eventually, yes. As of now I don't think `regression.sh` is used that often, as it would require the tested to have the belief, that the whole regression would run unattended without issues. I personally use it very rarely, because when I see a fail I want to immediately analyze it, not wait till the whole module finishes.

scripts/lib/robot.sh
philipanda: > Do you really commit and push all of your code (not just test code) prior doing any test on it?

I think that in the case of release tests, that should be the only possibility.

For any other scenario, like testing **your code** not the **firmware** - `ALLOW_DIRTY`

philipanda: >  Even though it isn't tested, 

I think it might be possible to test it

macpijan: Let's then update to Krystian's suggestion and go ahead with this

krystian-hebel: > > Do you really commit and push all of your code (not just test code) prior doing any test on it?
> 
> I think that in the case of release tests, that should be the only possibility.
> 
> For any other scenario, like testing **your code** not the **firmware** - `ALLOW_DIRTY`

This is exactly what I would do. Every **published** test result should be reproducible, and for testing the tests it's OK to have a workaround.

philipanda: Added a check for pushing the changes to remote

https://github.com/Dasharo/open-source-firmware-validation/pull/946/commits/25aa46de8facd5373c1925c66e6907e0dfdffaa1

```
❯ git add scripts/
❯ git commit -sS -m "scripts/lib/robot.sh: Check if local commit pushed to remote"
check for added large files..............................................Passed
check for merge conflicts................................................Passed
check for broken symlinks............................(no files to check)Skipped
detect private key.......................................................Passed
fix end of files.........................................................Passed
trim trailing whitespace.................................................Passed
mixed line ending........................................................Passed
codespell................................................................Passed
markdownlint.........................................(no files to check)Skipped
markdownlint-fix.....................................(no files to check)Skipped
ShellCheck v0.10.0.......................................................Passed
isort (python).......................................(no files to check)Skipped
black................................................(no files to check)Skipped
robotidy.............................................(no files to check)Skipped
robocop..............................................(no files to check)Skipped
reuse lint-file..........................................................Passed
platform-configs.....................................(no files to check)Skipped
test-ids-check.......................................(no files to check)Skipped
test-ids-sort........................................(no files to check)Skipped
Check Platform Config Variables......................(no files to check)Skipped
Conform..................................................................Passed
[prevent-dirty-runs 25aa46de8fac] scripts/lib/robot.sh: Check if local commit pushed to remote
 1 file changed, 16 insertions(+), 3 deletions(-)
❯ git push --force-with-lease
❯ ./scripts/run.sh util/test.robot
Checking if running the test from a reproducible revision.
If NECESSARY, set ALLOW_DIRTY to skip the check.
Local branch prevent-dirty-runs is ahead of origin/prevent-dirty-runs by 2 commits!
Push your changes before running any tests!
❯ 
```

philipanda: I don't think we can reasonably do much about changing the git history though. When someone does a rebase, all the commit IDs will change and GH will eventually remove those orphan commits from the history.

macpijan: But the test still starts despite warning, was that your goal?

```
NO_SETUP="true" FW_FILE=protectli_vp46xx_v1.2.0.rom DEVICE_IP=192.168.22 RTE_IP=192.168.10.203 CONFIG=protectli-vp4650 ./scripts/run.sh ./util/basic-platform-setup.robot 
Checking if running the test from a reproducible revision.
If NECESSARY, set ALLOW_DIRTY to skip the check.
Logs will be saved at logs/protectli-vp4650/basic-platform-setup_2025_07_18_14_50_11
Watch "logs/protectli-vp4650/basic-platform-setup_2025_07_18_14_50_11/basic-platform-setup_debug.log" to monitor the progress of the test
Basic-Platform-Setup                                                          
BPS001.001 Power Control - PSU ON and serial output :: Verifies if... | FAIL |
Parent suite setup failed:
Error checking out asset 44. Response data: {'status': 'error', 'messages': 'That asset is not available for checkout!', 'payload': {'asset': '00044'}}
```

macpijan: Sorry, it does not. But in such case, `If NECESSARY, set ALLOW_DIRTY to skip the check.` should not be displayed if there is nothing to skip?

scripts/lib/robot.sh
```diff
@@ -139,6 +139,29 @@ execute_robot() {
     installed_dut_option=""
   fi
 
+  # Prevent executing tests on a dirty git tree
+
+  if [[ -z "${ALLOW_DIRTY}" ]]; then
+    echo "Checking if running the test from a reproducible revision."
+    echo "If NECESSARY, set ALLOW_DIRTY to skip the check."
+
+    if ! git diff --quiet || ! git diff --staged --quiet; then
```
macpijan, line 150: Just a note that it does not include untracked files. Not sure if intended or not. Typically, we call a repo -dirty if there are new untracked files as well.

scripts/lib/robot.sh
macpijan: After reflection, I would also try to explain the rationale behind this to the reader. E.g.:

```
echo "Please commit your changes before running tests to ensure reproducibility. (Set ALLOW_DIRTY=1 to override)"
```

Understanding the reason (e.g., ensuring reproducibility via clean git state) makes it more likely users will follow the rule properly, rather than bypassing it out of frustration.

scripts/lib/robot.sh
philipanda: Won't that be too much? I personally keep the binaries I use at the time inside the osfv repo directory. It doesn't affect the test at all, as the files that are tracked and a part of the repo were not modified.

Would that help with reproducibility in any way? Maybe it could be an issue if someone develops a new test and forgot to add some important file. I can't really think of any other reason for that.

scripts/lib/robot.sh
philipanda: https://github.com/Dasharo/open-source-firmware-validation/pull/946/commits/e3dc156b111c2e09e45dc076a8e4171950d04796

scripts/lib/robot.sh
philipanda: I think it might be better to implement a way to give an URL to the used binary, instead of a local path (FW_FILE). It could link to cloud, dl.3mdeb.com, or GH CI. That way even the used binary would be 100% reproducible. That's if we don't take into account the URLs breaking, but that would still be a great improvement and using an URL instead of a local path could be enforced here the same way as keeping the tree clean.

macpijan: Storing binaries locally in repo is fine, it simply means that the .gitignore file might need to be extended if this is a common case.

macpijan: Anyway, let's start with this and see how much of a problem it is first.

krystian-hebel: This doesn't seem to catch changes in submodules.

krystian-hebel: Never mind, it just doesn't catch stuff like `modified:   3rdparty/dasharo-blobs (untracked content)`, which is the same as above, but in submodule. `(modified content)` is caught properly.

PR #953, opened 2025-07-23
CBnT: test reconstruction of PCR-0

SergiiDmytruk:
*ref: ncm-1811*

### DIFF
--- a/dasharo-security/cbnt.robot
+++ b/dasharo-security/cbnt.robot
@@ -60,12 +60,19 @@ Check CBnT Profile 5
     ...    profile 5
     [Tags]    robot:private
     [Arguments]    ${os_id}
-    Power On
-    Boot System Or From Connected Disk    ${os_id}
-    Login To Linux
-    Switch To Root User
+    Boot OS And Enter Root Shell    ${os_id}
     ${out_cbmem}=    Execute Command In Terminal    cbmem -1
     Should Match Regexp    ${out_cbmem}    FACB:\\S+1\\n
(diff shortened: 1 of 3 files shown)


dasharo-security/cbnt.robot
```diff
@@ -53,19 +54,63 @@ CBNT002.101 Converged Boot Guard and TXT Status Menu is visible
     Read From Terminal Until    Revoked
     Read From Terminal Until    Boot Guard Capability
 
+CBNT003.201 Converged Boot Guard and TXT - PCR-0 is reconstructed correctly (Ubuntu)
```
miczyg1, line 57: Does this test have to be CBNT specific? I mean, every paltform shoudl always reconstruct PCR0 regardless of CBNT state. IMO it belongs more to the TPM/measured boot suites

mkopec: Indeed, it does make sense to move this to a generic TPM/measured boot test suite.

SergiiDmytruk: We already have generic tests which check all PCRs: https://github.com/Dasharo/open-source-firmware-validation/blob/develop/dasharo-security/measured-boot.robot#L40-L46

This test is different in that it allows SHA1 bank to fail until coreboot starts to support all active banks.

mkopec (approved):
[cbnt003_logs.tar.gz](https://github.com/user-attachments/files/21429189/cbnt003_logs.tar.gz)

PR #973, opened 2025-07-30
keywords.robot: Import platform configs default.robot

philipanda:
Importing the default platform configs in keywords.robot will fix language servers spamming warnings of undefined variables all over the place. The platform config defined in the ${CONFIG} variable is imported dynamically after all of the static imports, and takes precedence over them.

The documentation is confusing on this:
- not explicitly mentioned how it works: https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#variable-priorities-and-scopes
- https://robotframework.org/robotframework/latest/libraries/BuiltIn.html#Import%20Resource says that `Resources imported with this keyword are set into the test suite scope similarly when importing them in the Setting table using the Resource setting.`. This suggests, that the first imported resource should take precedence.

It is experimentally proven, to not be the case in a little test:
[resource-precedence_log.zip](https://github.com/user-attachments/files/21504717/resource-precedence_log.zip)

### DIFF
--- a/keywords.robot
+++ b/keywords.robot
@@ -16,6 +16,7 @@ Resource    lib/framework.robot
 Resource    lib/me.robot
 Resource    lib/network.robot
 Resource    lib/bsd.robot
+Resource    platform-configs/include/default.robot
 
 
 *** Keywords ***


philipanda:
Sadly, it seems, that only the variables directly defined in the `${CONFIG}` file are taking precedence over the `default.robot` variables. Variables defined in any includes from the `Settings` section of `${CONFIG}` will not override the `default.robot` values:

`${CHECK_POWER_LED_SUPPORT}` was not overridden in this case, as can be seen in the error message and `Log Variables` in `Prepare Test Suite`.
[fwupd-support_log.zip](https://github.com/user-attachments/files/21504947/fwupd-support_log.zip)

In that case it might not be possible to solve this that easily.

Issue #612, opened 2024-11-26
Reuse GH workflows


pietrushnic:
Most of this is copied from existing file, so we could avoid duplication: https://docs.github.com/en/actions/sharing-automations/reusing-workflows

Not a high priority, we can also leave it in issues.

I can also imagine, we would like to reuse this workflow not only in this repo, but use very similar config in e.g. coreboot or edk2 repos in the future.

_Originally posted by @macpijan in https://github.com/Dasharo/open-source-firmware-validation/pull/574#discussion_r1859249529_

macpijan:
I tried in this commit  and had to revert it: https://github.com/Dasharo/open-source-firmware-validation/commit/ab235e5f8e8cd37cff36e71d1f0d5f0577c82f63

The QEMU has been consistently failing after 10th case and I could not get the reason in reasonable time.

<img width="961" height="800" alt="Image" src="https://github.com/user-attachments/assets/0059f5bd-569f-49f1-968b-123b763db58d" />

macpijan:
new datapoint: one of these changes already breaks qemu tests https://github.com/Dasharo/open-source-firmware-validation/pull/979/commits/749650309286a03fc0d83f523881ab0a1328aad5

PR #976, opened 2025-07-31
lib/robot.sh: Disable dirty tree checks on CI

philipanda:
1. GH Actions adds a merge commit on top of the revision it is running on.
2. git is in detached state, not on a branch
3. multiple git processes might run on the same repo (hardware CI)

Due to these reasons, that the main reason for ensuring the git tree being clean is to reduce human error, I believe it would be much less troublesome to just disable these checks on CI instead of creating complex workarounds like https://github.com/Dasharo/open-source-firmware-validation/pull/975/files

### DIFF
--- a/dasharo-compatibility/efi.robot
+++ b/dasharo-compatibility/efi.robot
@@ -15,7 +15,7 @@ Resource            ../keys.robot
 # TODO:
 # - document which setup/teardown keywords to use and what are they doing
 # - go threough them and make sure they are doing what the name suggest (not
-# exactly the case right now)
+# exactly the case right now)s
 Suite Setup         Run Keywords
 ...                     Prepare Test Suite
 ...                     AND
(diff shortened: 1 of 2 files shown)


macpijan:
I don't think we need any extra code, just env variable.
It's been already done but qemu for some reason malfunctioned and I reverted this:https://github.com/Dasharo/open-source-firmware-validation/issues/612#issuecomment-3132388000  and I gave up

I will try to bring back just the env var setting.

philipanda:
@macpijan 
> I don't think we need any extra code, just env variable.

I've moved the code around to reduce indentations.
The only new code is if `$GITHUB_ACTIONS` is set, then no dirty tree checks.

macpijan:
Yeah, and I do not think it is needed since you already have `ALLOW_DIRTY`. Let me inspect this.

macpijan:
Included in: https://github.com/Dasharo/open-source-firmware-validation/pull/979

PR #988, opened 2025-08-01
fuzzysearch

philipanda:


macpijan:
I would add some comments and/or commit messages to the commits changing kwds to `fuzzy` versions. To explain (link) to the original issue and why we changed those instances, not the others.

wiktormowinski:
image

Any attempt to use this new keyword ends with this error. Even tried recloning osfv.

philipanda:
> image
> 
> Any attempt to use this new keyword ends with this error. Even tried recloning osfv.

Could I ask for the whole log? This error has never happened to me. Everything works when I'm cloning a fresh copy of OSFV too.

matmacieje:
Tested with dasharo-security/me-neuter.robot, some tests PASSed.

[0_9_0_rc3_me-neuter_2025_08_06_12_39_48.zip](https://github.com/user-attachments/files/21616825/0_9_0_rc3_me-neuter_2025_08_06_12_39_48.zip)

philipanda:
> Tested with dasharo-security/me-neuter.robot, some tests PASSed.
> 
> [0_9_0_rc3_me-neuter_2025_08_06_12_39_48.zip](https://github.com/user-attachments/files/21616825/0_9_0_rc3_me-neuter_2025_08_06_12_39_48.zip)

I've added some logging to the library, and to my suprise it didn't appear in the logs from the CI. It does in the @matmacieje's logs. This suggests that CI does not use the exact robotframework versions given in requirements.txt. Some weird caching must be happening.

macpijan:
Rebased

macpijan:
> Could I ask for the whole log?

@wiktormowinski Please do provide logs instead of screenshots. It slows us down with the decision here.

[custom-boot-menu-key_log.html](https://github.com/user-attachments/files/21857845/custom-boot-menu-key_log.html)


I think I have run into similar issue @philipanda when trying to start a test on nucbox.

philipanda:
Thanks for the logs @macpijan
I've fixed the issue with the inner library crashing out. Starting from a67ef9077543858e9247b8b60e57397f1f6acad1 it should work fine now.
[serial_stability_log.zip](https://github.com/user-attachments/files/21886769/serial_stability_log.zip)

Make sure to `rm -rf` the venv, and install the newest `requirements.txt`

Details:
The package I used for matching the strings is a bit fussy when it comes to the types of its parameters, and possibly has some bug which sometimes initializes one parameter with an invalid default value. Don't really have time to investigate into it more.

philipanda:
@wiktormowinski @matmacieje @macpijan 

Separately defining the total amounts of max insertions: TELNET_FUZZY_MAX_INSERTIONS, and max deletions: TELNET_FUZZY_MAX_DELETIONS was added https://github.com/Dasharo/open-source-firmware-validation/pull/988/commits/390f6f5da996f18386eb8d53f75fa2287e7a07a6.

It should be perfectly ready to use now
[serial_stability_out.zip](https://github.com/user-attachments/files/21886935/serial_stability_out.zip)

Remember to `rm -rf` the venv and install the newest `requirements.txt` before testing

macpijan:
@philipanda I tried on APU2 with the same error settings as for Novacustom in your MR. Resulting in some failures:
[apu-configuration-menu_log.html](https://github.com/user-attachments/files/21899713/apu-configuration-menu_log.html)

Is it the same problem as on @wiktormowinski side being discussed today?

philipanda:
> @philipanda I tried on APU2 with the same error settings as for Novacustom in your MR. Resulting in some failures:
> [apu-configuration-menu_log.html](https://github.com/user-attachments/files/21899713/apu-configuration-menu_log.html)
> 
> Is it the same problem as on @wiktormowinski side being discussed today?

No, that's a brand new error 🙃
And it looks like it appears somewhere deep in the fuzzysearch package. Seems like it tries to get a slice using a negative index during the search. I wonder If that's some character encoding issue. Otherwise it beats me why it hasn't appeared before.

wiktormowinski:
@philipanda there is an unresolved conflict in this PR

philipanda:
@wiktormowinski conflicts solved

wiktormowinski (approved):

lib/bios/menus.robot
```diff
@@ -365,10 +368,26 @@ Parse Menu Snapshot Into Construction
...
+        END
+
+        IF    ${line_valid}    Append To List    ${construction}    ${line}
     END
     Log    ${construction}
     ${construction}=    Get Slice From List    ${construction}    ${slice_start}    ${slice_end}
+    IF    $first_line is not None
+        ${idx}=    Get Index From List Fuzzy
+        ...    ${construction}
+        ...    Select Language <Standard English>
```
SergiiDmytruk, line 385: @philipanda, should this be `${first_line}` instead?

Also, where is this keyword defined?  Here's what I get even after `pip install -r  requirements.txt`:
```
No keyword with name 'Get Index From List Fuzzy' found. Did you mean:
    Collections.Get Index From List
```

philipanda: > @philipanda, should this be ${first_line} instead?

I think yes... the `Select Language...` makes sense as a default value of $first_line. Checking for $first_line only makes sense if we expect errors on serial, so it should probable also be guarded with something like `${TELNET_FUZZY_MAX_INSERTIONS} + ${TELNET_FUZZY_MAX_DELETIONS} + ${TELNET_FUZZY_MAX_SUBSTITUTIONS} > 0`. It was added because the number of lines before the menu often changes due to errors. I've probably set the `Select Language...` there and forgot about the $first_line argument as it started working, but other menus might not be so lucky?

> Also, where is this keyword defined? Here's what I get even after pip install -r  requirements.txt:

pip is unreliable.
Please try removing your venv completely and regenerating it. If that doesn't work cloning a fresh copy of OSFV was helping a couple times.

the keyword is defined in our RF fork https://github.com/3mdeb/robotframework/blob/fuzzysearch/src/robot/libraries/Collections.py#L267

SergiiDmytruk: > pip is unreliable.

Damn, it's even worse than I thought.

> Please try removing your venv completely and regenerating it. If that doesn't work cloning a fresh copy of OSFV was helping a couple times.

Thank you, that helped.

> I've probably set the Select Language... there and forgot about the $first_line argument as it started working, but other menus might not be so lucky?

Somehow it doesn't seem to break anything (likely because [Get Slice From List](https://robotframework.org/robotframework/latest/libraries/Collections.html#Get%20Slice%20From%20List) does nothing in this case).  I thought it did after trying to replace `Get Index From List Fuzzy` with `Get Index From List` and probably breaking the keyword in the process.

PR #990, opened 2025-08-03
Fixes for capsules on Odroid H4+

SergiiDmytruk:
~~Draft because still need to make sure this works as intended.~~

This contains changes related to DUT timeout which can potentially affect other tests in negative ways (not very likely unless something sets timeout to less than 30s), but the overall effect should be positive.  Timeout is no longer reset to 30s by default on running shell commands, making the behaviour of tests closer to what keywords in the tests suggest.

*ref: ncm-1941*

### DIFF
--- a/dasharo-stability/capsule-update.robot
+++ b/dasharo-stability/capsule-update.robot
@@ -408,15 +408,18 @@ Turn Off Active ME
     END
 
 Display Preparation Instructions
-    Log To Console    ******************************************************************************\n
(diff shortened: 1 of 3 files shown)


SergiiDmytruk:
The tests pass on Odroid now: [odroid-capsule-tests.zip](https://github.com/user-attachments/files/21584847/odroid-capsule-tests.zip)

CI fails because it tries to run capsule tests without capsules.

mkopec (approved):

PR #1001, opened 2025-08-12
lib: terminal.robot: fix VAR assignment (remove '=')

m-iwanicki:
When testing on QEMU I got error that argument with wrong format was used: `180.0=`. Likely when trying to set DUT timeout to previous one by some keyword.

```
ValueError: Invalid time string '180.0='.
```


m-iwanicki:
Merged and fixed in https://github.com/Dasharo/open-source-firmware-validation/pull/1018

PR #1005, opened 2025-08-14
keywords.robot: Default value of SNIPEIT_ALREADY_CHECKED_OUT_MANUALLY

philipanda:
Set to ${TRUE} unless proven otherwise

When testing without an RTE this variable
is undefined and the test teardown fails

### DIFF
--- a/keywords.robot
+++ b/keywords.robot
@@ -495,6 +495,7 @@ Prepare Test Suite
     IF    '${MANUFACTURER}' != 'QEMU' and '${CONFIG}' != 'no-rte'
         Import Osfv Libraries
     END
+    VAR    ${SNIPEIT_ALREADY_CHECKED_OUT_MANUALLY}=    ${TRUE}    scope=GLOBAL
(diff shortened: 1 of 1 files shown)


keywords.robot
```diff
@@ -495,6 +495,7 @@ Prepare Test Suite
     IF    '${MANUFACTURER}' != 'QEMU' and '${CONFIG}' != 'no-rte'
         Import Osfv Libraries
     END
+    VAR    ${SNIPEIT_ALREADY_CHECKED_OUT_MANUALLY}=    ${TRUE}    scope=GLOBAL
```
macpijan, line 498: Am I missing something, or there is some mismatch between commit msg and code (True vs  False)?

philipanda: Yes, @wiktormowinski has already pointed that out to me, but I kept forgetting to update the message. It's fixed now, thanks

macpijan (approved):

Issue #1012, opened 2025-08-18
Check The Platform Is A Laptop returns empty


matmacieje:
### Device

NUC BOX 125H & 155H

### RTE version

none

### OSFV version

develop

### Affected component(s) or functionality

dasharo-performance/cpu-frequency.robot, dasharo-performance/cpu-temperature.robot, ...

### Brief summary

 ${LAPTOP_PLATFORM} is not initialized properly

### How reproducible

10/10

### How to reproduce

Run any tests from dasharo-performance/cpu-frequency.robot, dasharo-performance/cpu-temperature.robot on NUC BOX DUT.

### Expected behavior

 ${LAPTOP_PLATFORM} is initialized properly

### Actual behavior

${LAPTOP_PLATFORM} is not initialized properly

### Link to screenshots or logs

[0_9_0_rc3_cpu-frequency_2025_08_18_16_31_52.zip](https://github.com/user-attachments/files/21837519/0_9_0_rc3_cpu-frequency_2025_08_18_16_31_52.zip)

### Solutions you've tried

Commenting out Check Power Supply keyword from Suite Setup.

philipanda:
This should fix the issue https://github.com/Dasharo/open-source-firmware-validation/pull/1017
Cherry picked from the v54-v1.0.0-rc4 branch

philipanda:
The PR https://github.com/Dasharo/open-source-firmware-validation/pull/1017 was reviewed and merged

Issue #198, opened 2024-01-16
Allow for running tests without RTE


macpijan:
### The problem you're addressing (if any)

Currently, it is assumed in the environment, that each setup needs RTE. This is not true anymore. In some cases we already have some workarounds for that (e.g. in QEMU).

### Describe the solution you'd like

Some tests can be run with no power control at all (we power up the device manually, the tests are executed over serial or SSH).

Some tests can be run with manual power control.

### Where is the value to a user, and who might that user be?

Existing users of OSFV can get rid of workarounds.

New users of OSFV can actually benefit from it more, by running tests themselves. At this moment, it is very demanding on the hardware side, as very few users have RTE, or other equipment.

### Additional context

Some more context can be extracted from this discussion: https://github.com/Dasharo/dasharo-issues/issues/468

macpijan:
Another use-case we faced was testing the [TwPM](https://twpm.dasharo.com/). We did not need RTE here, using it made our setup more complex and testing more time-consuming. We have applied the following diff to disable the power control via RTE, it was assumed the the DUT was powered on with SSH enabled, which can be the case for many end users as well:

```
diff --git a/dasharo-security/tpm2-commands.robot b/dasharo-security/tpm2-commands.robot
index 031f1d4a533b..aec1c174b82e 100644
--- a/dasharo-security/tpm2-commands.robot
+++ b/dasharo-security/tpm2-commands.robot
@@ -248,8 +248,8 @@ TPM2 Test Setup
 
 TPM2 Suite Setup
     Prepare Test Suite
-    Power On
-    Boot System Or From Connected Disk    ubuntu
+    #Power On
+    #Boot System Or From Connected Disk    ubuntu
     Login To Linux
     Switch To Root User
     Detect Or Install Package    tpm2-tools
diff --git a/os-config/ubuntu-credentials.robot b/os-config/ubuntu-credentials.robot
index def24c2674fb..4755a9896f2e 100644
--- a/os-config/ubuntu-credentials.robot
+++ b/os-config/ubuntu-credentials.robot
@@ -1,5 +1,5 @@
 *** Variables ***
-${UBUNTU_USERNAME}=         user
+${UBUNTU_USERNAME}=         ubuntu
 ${UBUNTU_PASSWORD}=         ubuntu
-${UBUNTU_USER_PROMPT}=      ubuntu@3mdeb:~$
[...]

PR #1013, opened 2025-08-19
Temporary changes for NUC BOX MTL testing, do not merge.

matmacieje:


keywords.robot
```diff
@@ -551,7 +551,7 @@ Prepare To SSH Connection
     ...    sections if the communication with the platform based on
     ...    the SSH protocol
     # tu leci zmiana, musimy brać platformy zgodnie z tym co zostało pobrane w dasharo
-    Open Connection And Log In
+    # Open Connection And Log In
```
macpijan, line 554: Why do we need such a change?

lib/options/options-lib_uefi-setup-menu.robot
```diff
@@ -91,7 +91,7 @@ Measure Coldboot Time
     # Do one more iteration than requested, as we may hit first boot which is always longer.
     FOR    ${index}    IN RANGE    0    ${iterations}+1
         Power Cycle On
-        Boot System Or From Connected Disk    ${os_id}
+        #    Boot System Or From Connected Disk    ${os_id}
```
macpijan, line 94: This assumes we just boot to the target OS as a default one?

What is the problem with this kwd?

lib/terminal.robot
```diff
@@ -65,7 +65,7 @@ Set DUT Response Timeout
         ${prev_timeout}=    Telnet.Set Timeout    ${timeout}
     ELSE IF    '${DUT_CONNECTION_METHOD}' in ['SSH', 'open-bmc']
         ${con}=    Get Connection
-        VAR    ${prev_timeout}=    ${con.timeout}=
```
macpijan, line 68: This looks like something to be merged into develop branch.

os-config/202-credentials.py
```diff
@@ -4,7 +4,7 @@
 
 DEVICE_OS_USERNAME = "linux"
 DEVICE_OS_PASSWORD = "linux"
-DEVICE_OS_HOSTNAME = "3mdeb"
+DEVICE_OS_HOSTNAME = "fedora"
```
macpijan, line 7: Why are such changes needed?

We have preseeds out of syncs https://github.com/Dasharo/preseeds/tree/main/fedora or we install OSes manually with different config?

platform-configs/include/default.robot
```diff
@@ -40,6 +40,7 @@ ${SB_ERROR_STRING}=                                 The image signature is inval
 
 # See: https://github.com/Dasharo/dasharo-issues/issues/614
 ${LAPTOP_EC_SERIAL_WORKAROUND}=                     ${FALSE}
+${LAPTOP_PLATFORM}=                                 ${TBD}
```
macpijan, line 43: Are you sure we need it here? Do you remember why?

No other platform defines this variable in configs.

This looks dynamically set in the lib/power.robot

keywords.robot
matmacieje: Because we don't want RTE IP check for RTE-less desktop setup.

lib/options/options-lib_uefi-setup-menu.robot
matmacieje: This is for Fedora testing on RTE-less setup, with no way to select boot entry.

lib/terminal.robot
matmacieje: Yes.

platform-configs/include/novacustom-nuc_box.robot
```diff
@@ -17,7 +17,7 @@ ${WIFI_CARD}=                           Intel Corporation Meteor Lake PCH CNVi W
 
 ${DEFAULT_BOOT_OS_ID}=                  ${ENV_ID_UBUNTU}
 @{TESTED_LINUX_DISTROS}=
-...                                     ${ENV_ID_UBUNTU}    # ${ENV_ID_FEDORA}
+...                                     ${ENV_ID_UBUNTU}
```
macpijan, line 20: You use it to set which OSes to test at the moment - in this case, Ubuntu?

os-config/202-credentials.py
matmacieje: According to @philipanda it's not needed anymore. I've been using Fedora installation media generated with earlier preseed revision.

variables.robot
```diff
@@ -273,6 +273,11 @@ ${OS_UBUNTU}=               ubuntu
 ...                         platform_vendor=Clevo
 ...                         sonoff_ip=192.168.10.229
 
+# NovaCustom NUC BOX desktop
+&{RTE74}=                   ip=192.168.10.254
```
macpijan, line 277: Is it the one on your desk, or in the lab?

platform-configs/include/novacustom-nuc_box.robot
matmacieje: Only Ubuntu vs Fedora, for Windows testing, Ubuntu settings are correct.

variables.robot
matmacieje: None, this is dummy entry.

platform-configs/include/default.robot
matmacieje: No, this is enforced by pre-commit.

keywords.robot
macpijan: @matmacieje So that would be a feature request for something like: https://github.com/Dasharo/open-source-firmware-validation/issues/198 ?

lib/options/options-lib_uefi-setup-menu.robot
macpijan: @matmacieje So with SSH? For SSH this kwd should be skipped: https://github.com/Dasharo/open-source-firmware-validation/blob/develop/lib/bios/menus.robot#L1287

So I wonder what's the exact issue here, so we can solve it properly.

lib/terminal.robot
macpijan: @matmacieje Can you please submit it then?

platform-configs/include/default.robot
macpijan: This is enforced by pre-commit if you define it in the config: https://github.com/Dasharo/open-source-firmware-validation/pull/1013/files#diff-61c5ab51dcb9fecfbf89c1af912362de1f549d4e0d6fb25c74597533c1e9bd25R14

But is it really needed to define?

Please take a look at keywords in lib/power.robot and point me to why is it needed. I need to understand it, so we can properly fix that and avoid such manual changes.

variables.robot
macpijan: A dummy entry to get past through the inability of running tests without RTE? So again https://github.com/Dasharo/open-source-firmware-validation/issues/198 or similar.

Is RTE required by OSFV also when we are running tests over SSH?

platform-configs/include/default.robot
matmacieje: https://github.com/Dasharo/open-source-firmware-validation/issues/1012 - with this issue fixed there will by no need for platform config manipulation.

macpijan: Thank you for pointing to the specific issue. That helps me to understand the context here.

keywords.robot
philipanda: The change is already merged https://github.com/Dasharo/open-source-firmware-validation/pull/1006/files
The line was added somewhat recently and is not needed unless someone wants to use RTE alongside SSH as the main connection method, but that breaks tests without an RTE

lib/options/options-lib_uefi-setup-menu.robot
philipanda: Is this change required?
If you check the definition of the keyword, it only does anything if the chosen OS is not already booted.
Maybe there is some bug in this keyword, that made it not work for you?

lib/terminal.robot
philipanda: Changes of weird tiny bugs like this often stay in larger PRs for an extended amount of time.
Very often multiple people have to fix them independently on their branches.
Maybe we should try to create one-commit PRs and fix them as soon as possible? Maybe even notify others, that such bug was hot-fixed and that they should rebase onto the default branch?

os-config/202-credentials.py
philipanda: I've actually left the PR as a draft by mistake, so no one has approved nor merged it yet. Could you review it @macpijan @matmacieje? https://github.com/Dasharo/preseeds/pull/35

platform-configs/include/default.robot
philipanda: I've cherry picked a commit from my v54 v1.0.0-rc4 testing branch that should fix that

platform-configs/novacustom-nuc_box-125H.robot
```diff
@@ -6,6 +6,17 @@ Resource    include/default.robot
 
 
 *** Variables ***
+${INITIAL_DUT_CONNECTION_METHOD}=       SSH
```
philipanda, line 9: It might be useful to create some common config that defines the testing mode, like RTE-only, RTE+sonoff, SSH+sonoff, SSH+none.
There is a limited amount of working configs, and many other variables are often fully dependent on the selected combination of DUT_CONNECTION_METHOD and POWER_CTRL. Example: TESTS_IN_FIRMWARE_SUPPORT is only TRUE if DUT_CONNECTION_METHOD is either Telnet, or PiKVM
Most of this added block could be separated from most configs to make it easier to choose from the modes, or even use an argument to robot to switch between them instead of doing such modifications on most regressions

variables.robot
philipanda: It was definitely not needed a couple months ago.
I guess modifying variables.robot shouldn't be needed for SSH tests with this line removed https://github.com/Dasharo/open-source-firmware-validation/pull/1013/files#diff-82f3f16c416be69886d34f9df8144e79ff3ebc7fb8f1382f45f69618752fc409R94

lib/terminal.robot
philipanda: merged https://github.com/Dasharo/open-source-firmware-validation/pull/1018

platform-configs/include/default.robot
macpijan: This looks like a proper fix: https://github.com/Dasharo/open-source-firmware-validation/pull/1017/files

philipanda: I've forgot to add a link, thanks

lib/options/options-lib_uefi-setup-menu.robot
macpijan: @matmacieje I see in the 2nd PR you also commented it out. Both me and @philipanda think it's not necessary. Can you please prove us wrong e.g. by attaching some logs?

matmacieje: I did it to avoid recurring errors of Boot Menu parsing (no Ubuntu entry on the list). Example:
[0_9_0_rc4_dcu_2025_08_25_13_18_54.zip](https://github.com/user-attachments/files/22004281/0_9_0_rc4_dcu_2025_08_25_13_18_54.zip)

philipanda: The keywords that parse the boot menu remove a hardcoded amount of lines from the start and from the end of the menu to get rid of unwanted lines that do not contain actual boot entries. If some newline is lost because of unstable serial connection, then this way of parsing the menu fails. In that case it is reasonable to temporarily comment this and allow the device to boot to the default boot option just to allow the tests to run. This change should not be merged though.
Fuzzy string comparisons can't help with newlines being lost when we parse the menus as lists.

A way to fix this would be to search for some characteristic strings in the setup menu instead of hardcoding the amount of lines to drop, but lost newlines separating the boot entries could still break all of this.

I guess the most robust way to overcome this problem would be to re-enter the boot menu if errors are encountered. That would make the execution longer if errors occur, but it won't fail.

Such extreme workaround should be implemented with care, especially that we have reasons to suspect that the errors on serial are caused by the implementation of the USB serial console in EDK2, as there are no serial console errors once the device boots into an OS.

PR #1023, opened 2025-08-20
Sb key protection test

miczyg1:
Addresses issue: https://github.com/Dasharo/dasharo-issues/issues/1478

### DIFF
--- a/dasharo-security/secure-boot.robot
+++ b/dasharo-security/secure-boot.robot
@@ -43,7 +43,6 @@ SBO001.001 Check Secure Boot default state (firmware)
     [Documentation]    This test aims to verify that Secure Boot state after
     ...    flashing the platform with the Dasharo firmware is
     ...    correct.
-    Skip If    not ${TESTS_IN_FIRMWARE_SUPPORT}    SBO001.001 not supported
     Skip If    not ${TESTS_IN_UBUNTU_SUPPORT}    SBO001.001 not supported
     Power On
     ${setup_menu}=    Enter Setup Menu Tianocore And Return Construction
@@ -60,7 +59,6 @@ SBO002.001 UEFI Secure Boot (Ubuntu)
     [Documentation]    This test verifies that Secure Boot can be enabled from
     ...    boot menu and, after the DUT reset, it is seen from
     ...    the OS.
-    Skip If    not ${TESTS_IN_FIRMWARE_SUPPORT}    SBO002.001 not supported
     Skip If    not ${TESTS_IN_UBUNTU_SUPPORT}    SBO002.001 not supported
 
     # 1. Make sure that SB is enabled
@@ -97,7 +95,6 @@ SBO002.002 UEFI Secure Boot (Windows)
     [Documentation]    This test verifies that Secure Boot can be enabled from
     ...    boot menu and, after the DUT reset, it is seen from
     ...    the OS.
-    Skip If    not ${TESTS_IN_FIRMWARE_SUPPORT}    SBO002.002 not supported
     Skip If    not ${TESTS_IN_WINDOWS_SUPPORT}    SBO002.002 not supported
 
     # 1. Make sure that SB is enabled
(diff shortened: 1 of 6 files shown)


miczyg1:
[secure-boot.zip](https://github.com/user-attachments/files/21989044/secure-boot.zip)

Successfully run the tests on ODROID-H4. The protections work as expected.

miczyg1:
It seems like the MSI platform doesn't want to work... Flashing failed.

Ingo-Albrecht:
> It seems like the MSI platform doesn't want to work... Flashing failed.

I get the following error on MSI 690 (just reporting, SPI write lock still enabled):

$ fwupdmgr get-devices 
...
UEFI dbx:
 Device ID:        <redacted>
 Summary:          UEFI revocation database 
 Current version:  20241101
 Minimum Version:  20241101
 Vendor:           UEFI:Microsoft
 Install Duration: 1 second
 Update Error:     Not enough efivarfs space, requested 30.7 kB and got 29.7 kB

Can it be linked?

miczyg1:
> > It seems like the MSI platform doesn't want to work... Flashing failed.
> 
> I get the following error on MSI 690 (just reporting, SPI write lock still enabled):
> 
> $ fwupdmgr get-devices ... UEFI dbx: Device ID: Summary: UEFI revocation database Current version: 20241101 Minimum Version: 20241101 Vendor: UEFI:Microsoft Install Duration: 1 second Update Error: Not enough efivarfs space, requested 30.7 kB and got 29.7 kB
> 
> Can it be linked?

Your error seem to be related to overloaded UEFI variables region space. We use 64KB only, but for some reason you are attempting to write half of the space with some operation.

The error I am getting is from our infrastructre in laboratory which failed to flash the board remotely with external programmer.

Ingo-Albrecht:
> Your error seem to be related to overloaded UEFI variables region space. We use 64KB only, but for some reason you are attempting to write half of the space with some operation.

Ok, yes. It may have been borked due to an old fwupd, similar to https://github.com/fwupd/fwupd/issues/8702. I tried to append the same Howyar's SysReturn dbx a while back.

BeataZdunczyk:
@miczyg1 New tests also need to be added to the JSON file in the same PR. Please follow the OSFV test naming convention and update the IDs accordingly: https://github.com/Dasharo/open-source-firmware-validation/blob/develop/docs/adding-and-naming-test-cases.md

miczyg1:
> @miczyg1 New tests also need to be added to the JSON file in the same PR. Please follow the OSFV test naming convention and update the IDs accordingly: https://github.com/Dasharo/open-source-firmware-validation/blob/develop/docs/adding-and-naming-test-cases.md

Should be fixed now.

miczyg1:
Scope regression fails because:
1. MSI seem to be bricked? Can't find prompt on serial it seems.
2. Basic platform setup failed on APU3 (at least the test with ansible). So the OS did not have efitools installed when my new tests were run, thus they failed. Also most of the SB tets fail, because the SB test data cannot be used on platform without PIKVM, or a plugged USB stick with the SB test data.

miczyg1:
@philipanda @matmacieje ping

dasharo-security/secure-boot.robot
```diff
@@ -133,7 +130,6 @@ SBO002.002 UEFI Secure Boot (Windows)
 SBO003.001 Attempt to boot file with the correct key from Boot Maintenance Manager (firmware)
     [Documentation]    This test verifies that Secure Boot allows booting a
     ...    signed file with a correct key.
-    Skip If    not ${TESTS_IN_FIRMWARE_SUPPORT}    SBO004.001 not supported
```
matmacieje, line 136: How is this TC (& following, up to SBO008) going to work without entering FW setup menu?

philipanda: It almost got me too :laughing: 
This skip is already in the `Suite Setup` so @miczyg1 has just removed the redundant duplicates repeated in every keyword

miczyg1: That's right.

PR #1024, opened 2025-08-21
sleep-lib: Check If Sleep Type... Fix failing to set global var

philipanda:
The global variable PLATFORM_SLEEP_TYPE_SELECTABLE was never being set resulting in fails

### DIFF
--- a/lib/sleep-lib.robot
+++ b/lib/sleep-lib.robot
@@ -7,23 +7,18 @@ Resource            ../keywords.robot
 *** Keywords ***
 Check If Platform Sleep Type Can Be Selected
     [Documentation]    Check if there is a Platform sleep type option
-    IF    not ${TESTS_IN_FIRMWARE_SUPPORT}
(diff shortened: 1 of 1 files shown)


macpijan (approved):
Not tested myself, but code LGTM.

If you attached logs from passing test, we could do it with more confidence.

macpijan:
... unless we will see something in scoped regression :slightly_smiling_face:

philipanda:
~~Sure, that's just what Fedora shows when s0ix is enabled. It either shows `[s2idle] deep` or `s2idle [deep]` meaning that both s2idle and deep suspends are supported, but only one of them is selected at a given moment.
Something must have changed a bit in Dasharo if before it was either `[s2idle] shallow` or `s2idle [deep]`, but it doesn't look like a bug~~

[network-interface-after-suspend_log.html](https://github.com/user-attachments/files/21914991/network-interface-after-suspend_log.html)

ps. WOW, github finally allows to attach a .html file, zipping the logs was so annoying

* Edit, I've described an entirely different issue that I was supposed to make a PR for just now :laughing:, sorry. Incidentally the log is good for showing this PR works

PR #1028, opened 2025-08-22
Vp2440

wiktormowinski:


lib/bios/menus.robot
```diff
@@ -1299,7 +1299,7 @@ Boot System Or From Connected Disk    # robocop: off=too-long-keyword
     IF    '''${boot_menu}''' == 'NOT_SET'
         ${menu_construction}=    Enter Boot Menu Tianocore And Return Construction
     ELSE
-        VAR    ${menu_construction}=    @{boot_menu}
+        VAR    @{menu_construction}=    @{boot_menu}
```
filipleple, line 1311: @wiktormowinski are you sure this is necessary to fix something, and that it doesn't break anything else? This is part of the most elementary code and it's been this way for a while IIRC, so it looks kinda sus

wiktormowinski: - Every  VAR in osfv is either `VAR $= $` or `VAR @= @` (`variable to variable` and `list to list`) or `VAR @= $` (`individual variables to list`)
- all this VAR business is one of the new RF7 changes so it is possible that not everything was handled properly
- later in this kwd menu_construction is expected to be a list anyways

wiktormowinski: the affected test
before:
[usb-detect_log.html](https://github.com/user-attachments/files/22021878/usb-detect_log.html)
after:
[usb-detect_log.html](https://github.com/user-attachments/files/22021893/usb-detect_log.html)

i ran a lot of test suites in recent days and there were no issues with this change on.

philipanda: `${menu}` saves it as a string, that can be interpreted as a list if `evaluated`, i.e. "[1, 2, 3]"
`@{menu}` makes RF treat it as an actual list. The string representation can still be accessed by using the $ sign.

So:
- created as ${menu} - if you try to use the variable `menu` as a string, use ${menu}, if you want to use it as a list, its no longer possible, @{menu} would be invalid, as the variable is a plain string, not anything list-like
- created as @{menu} - accessing it as a string - ${menu}, will work the same as above, the list will be converted into a string representation, but @{menu} will also be possible, as RF remembers its a list

In my opinion this change is 100% safe
reference: https://docs.robotframework.org/docs/variables

filipleple (approved):

PR #1046, opened 2025-08-28
scripts/ci/develop_pr_auto_regression.sh: print results

macpijan:


philipanda (approved):
Might be a little hard to navigate logs from multiple devices, but that's much better than downloading the artifacts each time, so it's an approve from me

PR #1047, opened 2025-08-29
restore-disk-auto.robot: Catenate logs for easier debugging historic runs

philipanda:
Before every line of text had to be manually extended because there is no good builtin for simultaneous scanning the serial and logging it to console at the same time


wiktormowinski (approved):
@philipanda LGTM just pacify the pre-commit and we're good to go

philipanda:
> @philipanda LGTM just pacify the pre-commit and we're good to go

@wiktormowinski thanks for the review, pre-commit's good now

PR #1057, opened 2025-09-02
os-specific-tasks/202.yaml: Add docker installation on Fedora

philipanda:


philipanda:
Logs from a run on Qemu:
[basic-platform-setup_log.html](https://github.com/user-attachments/files/22092459/basic-platform-setup_log.html)

os-config/ansible/os-specific-tasks/202.yaml
```diff
@@ -38,3 +38,40 @@
 - name: Downgrade default kernel
   ansible.builtin.command: grubby --set-default-index {{ default_kernel_index.stdout | int + 1 }}
   when: '"6.13" in default_kernel_version.stdout'
+
+# Install Docker
+- name: Remove potential unofficial packages
```
macpijan: There is maintained a role supporting multiple OSes doing all of that: https://github.com/3mdeb/workstation-config/blob/main/workstation-setup-eft.yml#L12

Perhaps simply including it would work better. We use it in workstation config and I also use it for our internal OSFV Fedora VM setup. I simply include the role and that's it.

philipanda: I've seen it too, but I wasn't sure if I trust it and copied the official instructions. If you say it works, then why not

philipanda: I've replaced the instructions with calling geerlingguy's role.
BPS test suite checks and installs the role in ansible-galaxy if it doesn't exist.
Every 3mdeb workstation should have it installed anyway thanks to workstation-config repo.
[basic-platform-setup_log.html](https://github.com/user-attachments/files/22660405/basic-platform-setup_log.html)

wiktormowinski (approved):
lgtm, tested it meself, rebase and we can merge

philipanda:
@wiktormowinski rebased

PR #1072, opened 2025-09-04
platform-configs: Increase CPT meausrement interval to 60s

philipanda:
The measurements being performed every second cause the test to effectively last for ~ 90 minutes instead of target 60minutes and the logs are over 300 MB in size


philipanda:
merged along with https://github.com/Dasharo/open-source-firmware-validation/pull/1130

PR #1115, opened 2025-09-24
dts-e2e: profiles: remove SMMSTORE migration from Heads transition

m-iwanicki:


ghost (approved):
LGTM, if it passes on qemu, then I'm fine with the changes.

PR #1116, opened 2025-09-25
dts-e2e: platform-configs: odroid-h4-ultra: remove e2e support

m-iwanicki:
We do not support any workflows in ODROID H4 Ultra in DTS, so we cannot run E2E tests


DaniilKl (approved):

DaniilKl:
Please, rephrase the PR comment to `We do not support ODROID H4 Ultra in DTS, hence we cannot run DTS E2E tests.` and add it to the merge commit message. Otherwise it is confusing.

PR #1117, opened 2025-09-26
V560tu v1.0.0 rc6

philipanda:


wiktormowinski (approved):
default os is set as fedora but at the same time it is a dynamically changing parameter so i guess we can leave it be

Issue #748, opened 2025-03-29
Reduce unnecessary reboots in tests


philipanda:
### The problem you're addressing (if any)

Tests that are performed on multiple OSes generally perform a lot of unnecesary reboots. Instead of running all little checks on one OS and the moving to another, we are doing a little check on every supporteed OS before moving to another one.

Example: the DSP test suite. The DSP001, DSP002, DSP003 test on one OS can be done in one go. The total execution time would be very close to executing just one of them, because most of the time is wasted on rebooting the platform and the OS. 

### Describe the solution you'd like

The tests should not perform reboots if it is not necessary.
How to solve this is not straightforward.

### Where is the value to a user, and who might that user be?

A significant decrease of tests run time, especially on regressions.

### Describe alternatives you've considered

Ideas

1. The order of executing the tests has to be changed
  - Just swap the test cases in test suites, so that all cases on one OS are next to each other. RF executes cases in
    the order they appear in the .robot file
  - separate tests on different OSes to their own files
      + the test file name when running robot can include a wildcard character, which might make this a good solution
      + `scripts/run.sh dasharo-compatibility/network*.robot` works and runs `network-boot-utilities`, `network-boot` and
        `network-speed` in one run. Something like `scrupts/run.sh dasharo-compatibility/*_201.robot` could be used to run
        all tests on `201`.
     + There's also the `-t` flag for `robot` which allows filtering by test ID
       - `scripts/regression.sh -- -t "*.201*" -t "*.202*" -t "*.301"` would run all the tests containging `.201`, `.202` or `.301` in
         their names allowing for selecting the environments  manually
     + the test setup could also just contain a `Skip` if the OS, for which tests this file contains, is not set as supported in
        config variables 
2. After the order is altered, we need to modify the tests to not reboot if unnecessary
  - [Keeping the state of the platform](https://github.com/Dasharo/open-source-firmware-validation/issues/686) and making the `Power On` keyword work more like
    `Ensure Power On`.
    - many tests use `Power On` as a way to reboot the device. Such cases would need to be fixed. Maybe a
      separate keyword `Power Off` could be added and used to clearly state when the platform is actually
      rebooted, and when the `Power On` is used only to ensure it is On.
  - If the tests on different OSes would be separated to their own files, they could be simplified, so that Powering
   On the platform only happens during test suite setup, not on test case basis.
  - If tests on different OSes would stay in one file, a dummy test case could be added before OS change, that 
[...]

Issue #1099, opened 2025-09-16
Add a possibility to run tests simultaneously


filipleple:
### The problem you're addressing (if any)

We run hours long tests continuously probing the cpu frequency, then we reboot and  run hours long tests continuously probing the cpu temperature, then we reboot and run hours long tests continuously checking if the platform hasn't rebooted, then the same all over but under load etc....

### Describe the solution you'd like

Boot the platform once, probe frequency, temperature and uptime and then save the results to appropriate test cases (CPF00n, CPT00n, STB00n). Then repeat under load. 

EXTRA: It could also be possible to SSH into the platform in the meantime and do some stateless tests, i.e. not requiring options to be toggled, working on default settings, e.g. dmidecode. Such tests would be tagged, and there would be a mode of running tests on the side that only uses SSH and doesn't do any power toggling

### Where is the value to a user, and who might that user be?

At least 3x speedup of the longest tests.

Furthermore, that way it's way more convenient to set up overnight testing. You would only need to see the first iteration of such combined test to know whether frequency, temperature and stability are checked correctly - no chance of the next test over crashing, ruining your overnight regression. 

### Describe alternatives you've considered

Manually probing multiple things at once. But we have it all automated, the tests just need to be coordinated better.

filipleple:
@BeataZdunczyk  What exactly would need to be done is:

- Take the following test suites:
   - CPU frequency
   - CPU temperature
   - STB00.. Verify if no reboot occurs in the OS
- Each of them just executes a probe command every X minutes until it's been 1 hour or so
- Copy those commands and paste them into one test that runs all of them in sequence and then waits the X minutes
- Ideally log the probed values to the console on the go
- Don't fail and stop the entire process when either of the commands returns a bad value, just make note of it  
- After the 1 hour has passed, collect the results and pass or fail corresponding test cases
  - This might get tricky, not sure how easy is it to pass and modify global variables between test cases in Robot

Of course, this would still need to be done separately for with/without load, Windows/Ubuntu. For each 4 cases (with/without load, Windows/Ubuntu) we take three 1-hour-long tests and turn them into one 1-hour-long test, saving 2 hours. That would mean **we save `4*2=8` hours by doing this,** in each regression testing session.

philipanda:
> [@BeataZdunczyk](https://github.com/BeataZdunczyk) What exactly would need to be done is:
> 
>     * Take the following test suites:
>       
>       * CPU frequency
>       * CPU temperature
[...]

PR #1141, opened 2025-10-29
Reduce unnecessary reboots: part 1, AUD, MNE

philipanda:
implementing changes for https://github.com/Dasharo/open-source-firmware-validation/issues/748


philipanda:
Logs showing:
1. A normal test case over serial that enters the setup menu works as before
[custom-boot-menu-key_log.html](https://github.com/user-attachments/files/23266832/custom-boot-menu-key_log.html)

2. AUD test suite, which had the platform power state control turned on by overriding a variable in `*** Variables ***` section and removing forced Power On calls. By doing so, the platform is rebooted only once per OS, to boot into the OS, which saves about 50% of execution time in this case: [audio-subsystem_log.html](https://github.com/user-attachments/files/23266819/audio-subsystem_log.html)

philipanda:
Splitting the suites into multiple files per OS was supposed to be a way to achieve less reboots during tests. Turns out, that this approach is essentially what was used in a PR to solve https://github.com/Dasharo/open-source-firmware-validation/issues/1099 in https://github.com/Dasharo/open-source-firmware-validation/pull/1130. The approach where we first gather all the data required for all tests (for a given OS) and then we only interpret the collected data in test cases is essentially what the solution for concurrent execution is. Splitting into multiple files, one per OS, allows to run data gathering on every OS independently.

Just monitoring the power state and reducing reboots if not needed can work without this change. Just `${PLATFORM_STATE_CONTROL}=    ${TRUE}` and removing forced power ons (`Power On Ex force_reboot=${True}`) is enough to benefit from reduced reboots. That makes the tests much more tidy though, so I'd stand for keeping the splits from this PR.

Logs after splitting the AUD test suite into files per OS.  [audio-subsystem_log.html](https://github.com/user-attachments/files/23303807/audio-subsystem_log.html)

lib/platform/power.robot
```diff
@@ -7,12 +7,72 @@ Library             Process
...
 Library             SSHLibrary
+
 Resource            ../../variables.robot
 Resource            ../../keywords.robot
 Resource            ../../keys.robot
+Resource            boot_state.robot
 
 
 *** Keywords ***
+Init Power State Control
```
macpijan, line 17: Would it make sense to have some tests for these new keywords on QEMU under `self-tests` to ensure they work properly?

lib/platform/power.robot
```diff
@@ -7,12 +7,72 @@ Library             Process
...
+    IF    $POWER_STATE_POWERED_ON is ${None}    Try Check Power State
+    IF    ${POWER_STATE_POWERED_ON} or $POWER_STATE_POWERED_ON is ${None}
+        Rte Power Off
+        Sleep    10s
+    END
+    Read From Terminal
+    Power Cycle On
+    Boot State Control Notify State    booting
+
+Power On Ex
```
macpijan, line 52: `Ex` stand for what exactly?

keywords.robot
```diff
@@ -57,7 +57,12 @@ Login To Linux
         Login To Linux Via OBMC    root    root
     ELSE
         Press Enter
-        Login To Linux Over Serial Console    ${DEVICE_OS_USERNAME}    ${DEVICE_OS_PASSWORD}
+        ${line}=    Read From Terminal
```
macpijan, line 61: should we explain why do we need to handle this case?

macpijan:
Do you have some metrics - how long it took to run modified test suite - before and after the changes?

lib/platform/power.robot
philipanda: Extended/Expanded, something to differentiate from `Power On` for which `Power On Ex` is a wrapper

lib/platform/power.robot
philipanda: I guess the best way to test them would be to just do the same setup menu tests as now, just with the power state control ON and `Power On Ex    force_reboot=${TRUE}` swapped to `Power On Ex    force_reboot=${FALSE}`. Alternatively a separate test case could be added that just moves between setup/bootmenu/OS repeatedly a couple times.

philipanda:
> Do you have some metrics - how long it took to run modified test suite - before and after the changes?

Sure @macpijan , here's a direct comparison from logs:
- AUD originally - 7:17 [audio-subsystem_log.html](https://github.com/user-attachments/files/23337877/audio-subsystem_log.html)
- AUD with automatic reboot reduction - 4:08 [audio-subsystem_log.html](https://github.com/user-attachments/files/23266819/audio-subsystem_log.html)
- AUD with automatic reboot reduction and split into files -  3:36 [audio-subsystem_log.html](https://github.com/user-attachments/files/23303807/audio-subsystem_log.html)

That's **44% time reduction** for just the automatic reboot reduction. Suprisingly after splitting into files-per-OS it gets even better at **51% time reduction**. I'm nut quite sure what has changed, maybe it's just variation.

keywords.robot
philipanda: I think the commit message should show more or less what's the intention: `Login To Linux: Handle already logged in via serial`. It's added so that if the OS is already booted and the user (or root) is already logged in, we can just continue instead of waiting for the login prompt (which will never appear)

macpijan:
@filipleple I wonder whether the QEMU failures are specific to your changes? I can see that the latest branch job in develop succeeded.

```
Run ./scripts/ci/qemu-self-test.sh
Logs will be saved at logs/qemu-selftests/setup-and-boot-menus_2025_11_05_08_51_30
Watch "logs/qemu-selftests/setup-and-boot-menus_2025_11_05_08_51_30/setup-and-boot-menus_debug.log" to monitor the progress of the test
Setup-And-Boot-Menus :: This suite verifies the correct operation of keywor...
Enter Boot Menu Tianocore :: Test Enter Boot Menu kwd                 | FAIL |
Parent suite setup failed:
No keyword with name 'Rte Check Power Led' found.
```

dasharo-compatibility/os-pfsense.robot
```diff
@@ -119,7 +119,13 @@ PFS006.502 Preseed operating system installer (pfSense)
 
 PFS007.502 Boot operating system installer into rescue shell (pfSense)
     [Documentation]    Boot installer into rescue shell.
+<<<<<<< HEAD
```
macpijan: Need to be fixed.

assessment/txe-secure-boot.robot
```diff
@@ -29,7 +29,7 @@ Test Setup          Restore Initial DUT Connection Method
 TXE001.001 Check if platform is not fused
     [Documentation]    This test aims to verify that the platform is not fused
 
-    Power On
+    Power On Ex    force_reboot=${TRUE}
```
macpijan, line 32: It's not really a `reboot`. From what I can see, we rely on this keyword when we want to perform so-called `coldboot` test case:

https://github.com/Dasharo/open-source-firmware-validation/blob/develop/lib/utc.robot#L282

lib/platform/boot_state.robot
```diff
@@ -0,0 +1,33 @@
+*** Settings ***
```
macpijan, line 1: I am trying to get a grasp of these two state machines.

- [boot state](https://newmo-oss.github.io/mermaid-viewer/?sequence-number=0#eJyNjlFLwzAUhd/9FedZKIKPAwVpOxyUpbQdPoiU0N60YfVeyKLFfy+NnZ0OZXkI4ea73zkHrz0lVndOv0bvt1cA0FpHjbfCyIowCNfz9Qui6B473rOMvMKGrUcuIzmUkwWxsHcyLCszGtaUMauZztIEmYznCJ8Sj7brz5Dv8ImIe2r2WGs7ULuEKmN+2eKPZiAoxg3ycgfFX9xJr8LTzE7bM2bMImXxBGe73kPMsUX4mU6untKiLquHKq3DO01qtcUdtsIUKOI2OP4yHsP+ta31cLhQd0G3yr39lH0Cc2mRhw==)
- [power state](https://newmo-oss.github.io/mermaid-viewer/?sequence-number=0#eJyNkt9rwjAQgN/3V9x8cRuUwR4FB/7oYKBGmsIexpDaXmvAJtJe5vbfj6TartqofQhN7st3d0lKiginIsqKKPe+X+4AABJRYExCSZgFdsEOn09f4HmvsFA0VoowGcC7FAQPS/bhB6tJGMzgfgh9qST2H1tbGO9gh23WDrXbbuNIejcAXxIW1QTmKHUHaX6FzAawVHssgEmYYhrpLTlYozmKzdzlNYXb+ETluy0SNqVW9ZwcyCF9mp4gdX3+j6BqvTEdgu2E/LckzIHrOMayPOM6jmaqCxM11Bnu7PniLmdjNWkVrQY5bjGmKgHbmUd0Dl/2Mu6E4Bn4RlOi9vI/WicPcH3sww5SEUIhsg2BShujjZlvORuFbyyYr8aMhSsejkIfhrBQsrpnlIlVuITNPbplvdJAvZt8hz6uGdcVdruzftxXpDlKfZuV8Ws+VbZNf93XRIQ=)

We can discuss this general idea tomorrow offline.

assessment/txe-secure-boot.robot
philipanda: We do distinguish these terms in OSFV, but I think the word "reboot" is generally used for any sequence of power off then quickly power on when going into details is not needed or details are unknown. My reasoning was to not go into details, but I see your argument. Maybe `force_power_cycle` would be better?

philipanda:
> I wonder whether the QEMU failures are specific to your changes? I can see that the latest branch job in develop succeeded.

I've seen them and do wonder too. I can't imagine for now why would such error happen due to my changes, but I will look into it nevertheless because we don't merge with failed CI

dasharo-compatibility/os-pfsense.robot
philipanda: https://github.com/Dasharo/open-source-firmware-validation/compare/3fdc06b5016ce88595bfa55312d2dd4740e6452c..00573eb5a1d4a5fcf9673f8a44b5dc89dd41d73a

assessment/txe-secure-boot.robot
pietrushnic: Or maybe use standarized nomenclature? https://en.wikipedia.org/wiki/ACPI#Power_states

philipanda: > Or maybe use standarized nomenclature? https://en.wikipedia.org/wiki/ACPI#Power_states

It could be a good thing to implement it in the whole OSFV to make some operations unambiguous. In this case the thing is I don't think it makes a difference what power state would the platform reach during the power cycle. What's important is that it boots back again, that the SoL is printed, that we can enter Setup/Bootmenu, so a `reboot` seemed to make sense.

Anyhow, I think we will be moving this experiment aside for now. Adding this complexity of remembering the state of the platform might not be needed to achieve the reduction in the number of reboots in most test cases. Splitting suites into multiple files with a single Power On in the suite setup should be enough to achieve the same or even better (due to keyword calls reduction) effects.

philipanda:
After an offline discussion we will be closing this PR for now.

Adding this complexity of remembering the state of the platform might not be needed to achieve the reduction in the number of reboots in most test cases.

Splitting suites into multiple files, as experimented on in this PR, with a single Power On in the suite setup should be enough to achieve the same or even better (due to keyword calls reduction) effects.

There will be a followup

assessment/txe-secure-boot.robot
pietrushnic: > Adding this complexity of remembering the state of the platform might not be needed to achieve the reduction in the number of reboots in most test cases.

Such things can be obtained from registers https://docs.amd.com/r/en-US/am011-versal-acap-trm/Reset-Reason-Register

macpijan: It is something else here. Filip here was trying to have a set of global variables (like a state machine) tracking the current (boot and power) state of the hardware. Based on these, keywords would act accordingly (e.g. skip Power On, if the DUT is in the desires state, such as loggeg in to linux, or in Setup Menu. We ended up achieving the same improvements simpler.

PR #1150, opened 2025-11-12
Reducing unnecessary reboots followup - AUD

philipanda:
A followup to https://github.com/Dasharo/open-source-firmware-validation/pull/1141

By splitting a test suite into multiple files it is possible to achieve the same gains as in the original PR without introducing new complexity.

Con - this approach requires individual approach for every suite instead of automating keeping the DUT in correct state


philipanda:
Execution time comparison:

- original `dasharo-compatibility/audio-subsystem.robot`: [audio-subsystem_log.html](https://github.com/user-attachments/files/23502312/audio-subsystem_log.html) - 7:03

- new `dasharo-compatibility/audio-subsystem/` split version: [audio-subsystem_log.html](https://github.com/user-attachments/files/23502191/audio-subsystem_log.html) - 3:46

226 s / 423 s = 0.534, so that's a 47% improvement, roughly the same as in the original PR if we count in some variation

macpijan (approved):

PR #1152, opened 2025-11-13
Fix ids

macpijan:


macpijan:
@philipanda I went ahead with changing the regexp, but then I realized this check applies to checking IDs in .robot files, not in .json. Do we check .json somewhere as well?

macpijan:
The reason for this change was an error when trying to sync the DB:

```
Database has cases which are not present in the local copy.
Make sure you're using correct, up-to-date branch.
[
    {
        "doc": {
            "_id": "DCU001.202",
            "name": "Change the UUID",
            "module": "Dasharo Compatibility"
        }
    },
    {
        "doc": {
            "_id": "DCU002.201",
            "name": "Change the serial number",
            "module": "Dasharo Compatibility"
        }
    },
    {
        "doc": {
            "_id": "DCU002.202",
            "name": "Change the serial number",
            "module": "Dasharo Compatibility"
        }
    },
    {
        "doc": {
            "_id": "DCU003.201",
            "name": "Change the bootsplash logo",
            "module": "Dasharo Compatibility"
        }
    },
    {
        "doc": {
            "_id": "DCU003.202",
            "name": "Change the bootsplash logo",
            "module": "Dasharo Compatibility"
        }
    },
 ```

philipanda:
Why does the regex need to be changed? Extracting the ID should work fine as we are taking the first group not the whole match.

philipanda:
I don't think the IDs in test_cases.json are verified on their own. We assumed that by checking if the IDs in .robot are correct, and then checking if they match the ones in .json, by a transitive relation the IDs in .json will be correct as well. We are ignoring the .robot==.json check for now as they are not in sync and the effort to sync everything up was not done yet, so maybe that's why some errors got there

macpijan:
> Why does the regex need to be changed? Extracting the ID should work fine as we are taking the first group not the whole match.

It does not, I can drop this commit. Initially I thought it was used to validate `id` field in JSON. But later I realized it was not.

> I don't think the IDs in test_cases.json are verified on their own. 

Thanks for confirmation.

A simple check of ID field format in .json could also help here.

philipanda:
> A simple check of ID field format in .json could also help here.

As a temporary patch, yes, I agree. The time when the .robot files would get in sync with the .json is being postponed more than we initially thought. Eventually such test won't be needed though.

PR #1165, opened 2025-12-01
coreboot-redundant-boot: Add slot A protection test

philipanda:


philipanda:
Tested by verifying whether bios-lock feature locks the expected offset range using:
```
&{COREBOOT_REDUNDANT_BOOT_BOOTBLOCK_OFFSET}=    start=0x00900000    end=0x00ffffff
&{COREBOOT_REDUNDANT_BOOT_COREBOOT_OFFSET}=    start=0x00900000    end=0x00ffffff
```
and enabling the bios lock in setup menu manually instead of checking for the not-yet implemented locks of slot A in coreboot redundant boot.
[coreboot-redundant-boot_log.html](https://github.com/user-attachments/files/23856329/coreboot-redundant-boot_log.html)

filipleple:
@philipanda hmmm, looking at the offsets is a nice way to check it, but setting them per each platform and each time the FMAP is adjusted seems a bit tedious. 

I'd suggest either 

* checking by trying to erase the regions that should be protected and expecting errors: 
  `flashrom -p internal -E --fmap -i BOOTBLOCK -i COREBOOT` 
   * erasing because flashrom can skip and succeed when trying to write an identical binary = false fails
   * this doesn't verify that **only** these regions are protected
* reading the FW and parsing `cbfstool read.rom layout`
   * you get:
     `'BOOTBLOCK' (CBFS, size 524288, offset 16252928)`
     `'COREBOOT' (CBFS, size 4161536, offset 11567104)`
   * this is ideal, but will require the extra hoop of converting from normal to hex

philipanda:
@filipleple 
> * erasing because flashrom can skip and succeed when trying to write an identical binary = false fails

Should the protections not work for one reason or another, we would end up with a bricked device unable to continue the regressions tests, so that's not perfect.

> reading the FW and parsing cbfstool read.rom layout
>    you get:
>    'BOOTBLOCK' (CBFS, size 524288, offset 16252928)

does that tell us anything about whether the regions are protected or not? Them existing is not enough

*edit oh you meant reading them and using in the tests instead of defining the offsets in the configs.
This might be a great idea. I can't see why conversion from base 10 to base 16 would be a problem?

philipanda:
@filipleple thanks for the idea with reading the FMAP offsets. It seems natural now.
Tested the new approach by verifying if the `COREBOOT` region is protected once BiosLock is manually set to Enabled in the setup menu, 'cause all vp66xx platforms are checked out currently

[coreboot-redundant-boot_log.html](https://github.com/user-attachments/files/24123383/coreboot-redundant-boot_log.html)

filipleple:
@philipanda not sure what but something funny is happening when I run the test. 

A manual check looks correct:

```
ubuntu@3mdeb:~$ cbfstool read.rom layout | grep "BOOTBLOCK\|COREBOOT"
(...)
'COREBOOT' (CBFS, size 4161536, offset 11567104)
'BOOTBLOCK' (CBFS, size 524288, offset 16252928)
```
Which means `COREBOOT` spans `11567104 - 15728640`

```
ubuntu@3mdeb:~$ sudo flashrom -p internal
(...)
PR0: Warning: 0x00b08000-0x00efffff is read-only.
PR1: Warning: 0x00f80000-0x00ffffff is read-only.
```
Which means `11567104 - 15728639` is protected


But when I run the test, it fails like this:

```
Region COREBOOT(11554816-15728640) is not protected!
Protected ranges: [{'start': 11567104, 'end': 15728640}, {'start': 16252928, 'end': 16777216}]
```


Not sure how the offset of COREBOOT becomes `11554816`, not `11567104` :thinking: 

[coreboot-redundant-boot_log.html](https://github.com/user-attachments/files/24159721/coreboot-redundant-boot_log.html)

philipanda:
@filipleple 
> @philipanda not sure what but something funny is happening when I run the test.

The COREBOOT from your manual run and the tests is moved by 0x3000 (11567104−11554816=12288).
`11554816` is exactly what cbfstool shows for this region in the automated test log. Are you sure it was tested on the same device? The same firmware? Could the COREBOOT somehow move? The `0x3000` does seem suspicious

filipleple:
alright uncommented the flashrom read line and now it works. 
[coreboot-redundant-boot_log.html](https://github.com/user-attachments/files/24165928/coreboot-redundant-boot_log.html)

(skipping the previous test cases here for quicker testing)

dasharo-stability/coreboot-redundant-boot.robot
```diff
@@ -5,18 +5,19 @@ Library             String
 Resource            ../variables.robot
 Resource            ../keywords.robot
 
-Suite Setup         Run Keyword
+Suite Setup         Run Keywords
 ...                     Prepare Test Suite
+# ...    AND    Skip If    not ${COREBOOT_REDUNDANT_BOOT_SUPPORT}    coreboot redundant boot not supported
```
filipleple: uncommenting this

platform-configs/protectli-vp6650.robot
```diff
@@ -3,41 +3,45 @@ Resource    include/protectli-vp66xx.robot
...
+@{ETH_PORTS}=                                   64-62-66-22-84-f5
+...                                             64-62-66-22-84-f6
+...                                             64-62-66-22-84-f7
+...                                             64-62-66-22-84-f8
+
+${SATA_SUPPORT}=                                ${True}
+${TESTS_IN_XCP_NG_SUPPORT}=                     ${True}
+@{TESTED_LINUX_DISTROS}=                        ${ENV_ID_UBUNTU}    ${ENV_ID_XCP_NG}
+
+# TODO - define in platform configs depending on platforms fmap
```
filipleple: TODOn't - not needed anymore, it's obtained automatically

filipleple (changes requested):

filipleple (approved):

PR #1167, opened 2025-12-02
dts/profiles: Add and update profiles for APU2/3/4/6 - DPP

aronowski:


m-iwanicki (changes requested):
Same as: https://github.com/Dasharo/open-source-firmware-validation/pull/1168#pullrequestreview-3530200353

@DaniilKl We should update tests/profiles based on DTS built from develop not newest release. If we want to run tests after releasing newer firmware (by modifying only `dts-configs`) then those tests should probably be run on the same `OSFV` commit as CI does before release.

DaniilKl (approved):

m-iwanicki (approved):
Tested, rebased, squashed, fixed pre-commit error (@DaniilKl please use `pre-commit install` command in osfv repo).

PR #1118, opened 2025-09-29
DTS 2.7.1 tests

m-iwanicki:
Required DTS changes: https://github.com/Dasharo/dts-scripts/pull/115
Tests related to changes done in v2.7.1 (FUM, flashrom read failure in `set_flashrom_update_params` and smmstore migration):

The same tests fail as always (MTL TNx (dGPU) Fuse platform workflow (not supported yet), MSI Heads transition, and some credential related ones)

```text
==============================================================================
Dts-E2E
==============================================================================
E2E010.001 Failure to read flash during update should stop workflo... | PASS |
------------------------------------------------------------------------------
E2E011.001 Aborting update after smmstore migration failure should... | PASS |
------------------------------------------------------------------------------
E2E012.001 Continuing update after smmstore migration failure shou... | PASS |
------------------------------------------------------------------------------
E2E013.001 Verify that FUM update doesn't start automatically :: T... | PASS |
------------------------------------------------------------------------------
E2E014.001 Verify that FUM update succeeds :: Test that FUM update... | PASS |
------------------------------------------------------------------------------
E2E015.001 Verify that entering DTS menu in FUM works :: Test that... | PASS |
------------------------------------------------------------------------------
Dts-E2E                                                               | PASS |
6 tests, 6 passed, 0 failed
==============================================================================
```

Another changes:
- updated prompts after migrating all choice prompts into one function
- changed all main DTS menu checkpoints to write bare (without Enter) - fixes https://github.com/Dasharo/open-source-firmware-validation/issues/1075
- moved `Clean Up DTS Environment` to Test Setup
- added queued checkpoints, fixes https://github.com/Dasharo/open-source-firmware-validation/issues/970

---

All tests:

<details><summary>Details</summary>
<p>

```text
==============================================================================
Dts-E2E
==============================================================================
Create tests                                                          | PASS |
------------------------------------------------------------------------------
E2E001.001 HCL Report test :: Verify that HCL Report is being exec... | PASS |
------------------------------------------------------------------------------
E2E002.001 DCR Initial Deployment On Msi-pro-z690-a-ddr5 With 13th... | PASS |
------------------------------------------------------------------------------
E2E002.002 DCR Initial Deployment On Msi-pro-z690-a-wifi-ddr4 With... | PASS |
------------------------------------------------------------------------------
E2E003.001 DCR UEFI Update On Msi-pro-z690-a-ddr5 With 13th Gen CP... | PASS |
------------------------------------------------------------------------------
[...]

PR #1168, opened 2025-12-02
lib/dts-lib.robot: change case and parentheses to brackets in Variables

aronowski:


m-iwanicki (changes requested):
Changed here: https://github.com/Dasharo/open-source-firmware-validation/pull/1118/commits/b4f2fe92efea6ed3400d2e967b72d995cee5135f and https://github.com/Dasharo/meta-dts/commit/2d87e1ec0f8dde5d7d7c6ea311ac528056c8cd6f

PR should probably be closed. Or maybe we need separate OSFV branches for current release/current develop

PR #1189, opened 2025-12-19
dasharo-stability/capsule-update.robot: Add BtG popup test

mkopec:
*ref: sha-1547*

### DIFF
--- a/dasharo-stability/capsule-update.robot
+++ b/dasharo-stability/capsule-update.robot
@@ -422,23 +422,16 @@ Upload Required Files SSH
     Execute Command In Terminal    rm -rf osfv
     Execute Command In Terminal    git clone https://github.com/dasharo/open-source-firmware-validation osfv
     IF    '${BTG_CAPSULE_FW_FILE}' != '${EMPTY}'
-        VAR    ${commands}=    pushd osfv;
-        ...    git submodule update --init --checkout;
-        ...    export FW_FILE=/root/${fw_filename};
-        ...    export CAPSULE_FW_FILE=/root/${caps_filename};
-        ...    export BTG_CAPSULE_FW_FILE=/root/${btg_caps_filename};
-        ...    ./scripts/capsules/capsule_update_tests.sh /root/${caps_filename};
-        ...    ./scripts/capsules/prepare_capsule_update_tests_drive.sh ${capsule_disk};
-        ...    popd;
-    ELSE
-        VAR    ${commands}=    pushd osfv;
-        ...    git submodule update --init --checkout;
-        ...    export FW_FILE=/root/${fw_filename};
-        ...    export CAPSULE_FW_FILE=/root/${caps_filename};
-        ...    ./scripts/capsules/capsule_update_tests.sh /root/${caps_filename};
-        ...    ./scripts/capsules/prepare_capsule_update_tests_drive.sh ${capsule_disk};
-        ...    popd;
+        Execute Command In Terminal    export BTG_CAPSULE_FW_FILE=/root/${btg_caps_filename}
     END
+    VAR    ${commands}=    pushd osfv;
+    ...    git checkout f672379ed071edc836164fd4f998fe2e5fb8d725;
+    ...    git submodule update --init --checkout;
+    ...    export FW_FILE=/root/${fw_filename};
+    ...    export CAPSULE_FW_FILE=/root/${caps_filename};
+    ...    ./scripts/capsules/capsule_update_tests.sh /root/${caps_filename};
+    ...    ./scripts/capsules/prepare_capsule_update_tests_drive.sh ${capsule_disk};
+    ...    popd;
     Execute Command In Terminal    ${commands}    timeout=120s
 
 Perform Capsule Update
--- a/dasharo-stability/capsule-update.robot
+++ b/dasharo-stability/capsule-update.robot
@@ -417,8 +417,9 @@ Upload Required Files SSH
     Switch To Root User
     Send File To DUT    ${FW_FILE}    /root/${fw_filename}
     Send File To Dut    ${CAPSULE_FW_FILE}    /root/${caps_filename}
-    Send File To Dut    ${CAPSULE_FW_FILE}    /root/${caps_filename}
+    Send File To Dut    ${BTG_CAPSULE_FW_FILE}    /root/${btg_caps_filename}
     ${capsule_disk}=    Identify Path To USB    ${CAPSULE_UPDATE_DISK_MODEL}
+    Execute Command In Terminal    rm -rf osfv
     Execute Command In Terminal    git clone https://github.com/dasharo/open-source-firmware-validation osfv
     IF    '${BTG_CAPSULE_FW_FILE}' != '${EMPTY}'
         VAR    ${commands}=    pushd osfv;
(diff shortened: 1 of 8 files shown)


mkopec:
Managed to run the test - it's the only passing one here:
[capsule-update_btg_V560TU_fused.zip](https://github.com/user-attachments/files/24525934/capsule-update_btg_V560TU_fused.zip)

Got to the prompt:
image

dasharo-stability/capsule-update.robot
```diff
@@ -371,21 +408,37 @@ Upload Required Files Serial
...
+    ELSE
+        VAR    ${commands}=    pushd osfv;
+        ...    git submodule update --init --checkout;
+        ...    export FW_FILE=/root/${fw_filename};
+        ...    export CAPSULE_FW_FILE=/root/${caps_filename};
+        ...    ./scripts/capsules/capsule_update_tests.sh /root/${caps_filename};
+        ...    ./scripts/capsules/prepare_capsule_update_tests_drive.sh ${capsule_disk};
+        ...    popd;
+    END
     Execute Command In Terminal    ${commands}    timeout=120s
```
SergiiDmytruk: Can we add something like this instead?
```robot
IF    '${BTG_CAPSULE_FW_FILE}' != '${EMPTY}'
    Execute Command In Terminal    export BTG_CAPSULE_FW_FILE=/root/${btg_caps_filename}
END
```

The state of the shell shouldn't be reset and the path is absolute, so seems like everything should work the same.

mkopec: Yeah, that makes sense https://github.com/Dasharo/open-source-firmware-validation/pull/1189/commits/981a9cad853b0142fcfef92c917749ad7dcdcfac

SergiiDmytruk: Did you mean to add `git checkout f672379ed071edc836164fd4f998fe2e5fb8d725`?  It wasn't here before the last commit.

LGTM otherwise.

SergiiDmytruk (approved):

dasharo-stability/capsule-update.robot
mkopec: I had that locally but didn't want to have an explicit checkout here in the version that eventually gets merged. Do you think it should be there?

SergiiDmytruk: I don't see a reason for keeping the checkout, it can actually result in use of outdated scripts.  And explicit checkouts should probably use tags rather than commit hashes anyway.

platform-configs/novacustom-v540tu.robot
```diff
@@ -125,6 +125,8 @@ ${DTS_TEST_BOARD_MODEL}=                V540TU
 
 ${HDMI_AUDIO_SUPPORT}=                  ${TRUE}
 
+${INTEL_CBNT_BOOTGUARD_FUSED}=          ${TRUE}
```
philipanda: As its specific to a single device, not the whole model, I don't think such thing should be set in platform-configs

scripts/lib/robot.sh
```diff
@@ -262,6 +262,13 @@ execute_robot() {
     capsule_fw_file_option=""
   fi
 
+  # BTG_CAPSULE_FW_FILE environment variable is required for the capsule update test
```
philipanda, line 265: RF can read environment variables using `Get Environment Variable`

dasharo-stability/capsule-update.robot
```diff
@@ -371,15 +408,24 @@ Upload Required Files Serial
 Upload Required Files SSH
     ${fw_filename}=    Get File Name Without Extension    ${FW_FILE}
     ${caps_filename}=    Get File Name Without Extension    ${CAPSULE_FW_FILE}
+    IF    '${BTG_CAPSULE_FW_FILE}' != '${EMPTY}'
```
philipanda: Accessing the variable by name will crash if its not defined.

scripts/lib/robot.sh
mkopec: I was just following in the example of `CAPSULE_FW_FILE` above

dasharo-stability/capsule-update.robot
mkopec: 0bcde3fef00ff742448f9ed1d65eca772c846b05

platform-configs/novacustom-v540tu.robot
mkopec: Do you have suggestions on how this should be handled instead?

philipanda: basically https://github.com/Dasharo/open-source-firmware-validation/issues/1220

as of now, there is no good way to do that. Had we tested the laptops using RTEs and a static configurations in the lab, the `variables.robot` would be a good place for that information. Otherwise whats left are environment variables or semiauto dialogs. Otherwise this variable will be oscilating back and forth all the time and will cause unexpected failed runs when someone forgets to update it to their device.

dasharo-stability/capsule-update.robot
mkopec: can we close this thread @SergiiDmytruk ?

platform-configs/novacustom-v540tu.robot
mkopec: Not sure what I can do in this PR until https://github.com/Dasharo/open-source-firmware-validation/issues/1220 is fixed

philipanda: I'd say check in cbmem whether `FPFs Committed` is `YES` in suite setup. It can be saved to a global variable so that this check needs to be done only once per run. The same variable should be also settable via an environment variable so that the check can be skipped explicitly.

My suggestion that will not break osfv stability and allow for some flexibility:
1. `${INTEL_CBNT_BOOTGUARD_FUSING_SUPPORT}` - permamently set to TRUE on platforms that *can* be fused
2. `${INTEL_CBNT_BOOTGUARD_FUSED}` set to `${TBD}` in default.robot
3. Use a keyword in Suite Setup that:
   1. checks if env variable `INTEL_CBNT_BOOTGUARD_FUSED` is set
   2. `INTEL_CBNT_BOOTGUARD_FUSED` can be set to True/False to skip checks
   3. if not set, then perform the check for fuses being commited in cbmem and save the result to `${INTEL_CBNT_BOOTGUARD_FUSED}` as a global variable
4. If etiher `${INTEL_CBNT_BOOTGUARD_FUSING_SUPPORT}` or `${INTEL_CBNT_BOOTGUARD_FUSED}`  is not `True`, then  tests that require fusing are skipped

dasharo-stability/capsule-update.robot
SergiiDmytruk: @mkopec, I thought we both don't see a need for keeping `git checkout` and you intended to removed it, no?

platform-configs/novacustom-v540tu.robot
mkopec: b05860a4e351837db256db9e930626fd2041cde6

philipanda: great

dasharo-stability/capsule-update.robot
mkopec: Sorry, thought it was removed. Removed now. https://github.com/Dasharo/open-source-firmware-validation/pull/1189/commits/ae45876fbe68746e05c8abfab3ae7c0a8ac3cf52

PR #1203, opened 2026-01-19
dts: msi-pro-z690-a-wifi-ddr4: regenerate profiles

DaniilKl:
These profiles have been regenerated without testing on hardware as 3mdeb does not have hardware for testing currently. As it is not the correct way to regenerate profiles as described in
https://blog.3mdeb.com/2025/2025-10-29-maintaining-and-testing-dts/ it needs some reasoning to be written down.

The reasoning behind trusting a profile without verification on hardware:

1. The same MSI boards but for different DDR types have similar mocking configurations, that differ only by the MSI model in dmidecode variables.
2. The same MSI boards but for different DDR types have the same profiles for the same mocking configurations.
3. Apart from the above two facts the "msi-pro-z690-a-wifi-ddr4 UEFI->Heads Transition - DPP.profile" will be verified by senior Dasharo firmware developer to make sure all commands that modify the firmware state are expected and correct for this DTS workflow on this platform.

### DIFF
--- a/dts/profiles/msi-pro-z690-a-wifi-ddr4 UEFI->Heads Transition - DPP.profile
+++ b/dts/profiles/msi-pro-z690-a-wifi-ddr4 UEFI->Heads Transition - DPP.profile
@@ -1,4 +1,3 @@
-# Couldn't verify: https://github.com/Dasharo/dasharo-issues/issues/1533
 dmidecode -s system-manufacturer 0
 dmidecode -s system-product-name 0
 dmidecode -s baseboard-product-name 0
@@ -31,12 +30,11 @@ flashrom -p internal 0
 ifdtool -d /tmp/biosupdate 0
 fsread_tool test -d /sys/class/pci_bus/0000:00/device/0000:00:16.0 1
 cbmem -1 0
-cbmem -1 0
 flashrom -p internal -N --ifd -i bios -r /tmp/bios.bin 0
 cbfstool /tmp/bios.bin layout -w 0
 cbfstool /tmp/biosupdate layout -w 0
 flashrom -p internal -N --ifd -i fd -w /tmp/biosupdate 0
-flashrom -p internal --ifd -i bios -i fd -w /tmp/biosupdate 0
-flashrom -p internal --ifd -i bios -i fd -w /tmp/biosupdate 0
+flashrom -p internal --ifd -i bios -i fd -i me -w /tmp/biosupdate 0
+flashrom -p internal --ifd -i bios -i fd -i me -w /tmp/biosupdate 0
 reboot  0
 dmidecode  0
--- a/dts/profiles/msi-pro-z690-a-wifi-ddr4 UEFI->Heads Transition - DPP.profile
+++ b/dts/profiles/msi-pro-z690-a-wifi-ddr4 UEFI->Heads Transition - DPP.profile
@@ -23,6 +23,13 @@ flashrom -p internal --flash-name 0
 flashrom -p internal --flash-size 0
 fsread_tool test -e /sys/class/power_supply/AC/online 1
 flashrom -p internal 0
+cbfstool /tmp/biosupdate layout -w 0
+cbfstool /tmp/biosupdate print -r COREBOOT 0
+flashrom -p internal -r /tmp/rom.bin --ifd -i bios 0
+cbfstool /tmp/rom.bin layout -w 0
+cbfstool /tmp/rom.bin read -r ROMHOLE -f /tmp/romhole.bin 0
+cbfstool /tmp/biosupdate remove -r COREBOOT -n msi_romhole.bin 0
+cbfstool /tmp/biosupdate add -r COREBOOT -n msi_romhole.bin -f /tmp/romhole.bin -b 0xff7c0000 -t raw 0
 flashrom -p internal -r /tmp/dasharo_dump.rom --fmap -i FMAP -i BOOTSPLASH 0
 cbfstool /tmp/dasharo_dump.rom extract -r BOOTSPLASH -n logo.bmp -f /tmp/logo.bmp 1
 cbfstool /tmp/biosupdate extract -r COREBOOT -n config -f /tmp/biosupdate_config 0
--- a/platform-configs/msi-pro-z690-a-wifi-ddr4.robot
+++ b/platform-configs/msi-pro-z690-a-wifi-ddr4.robot
@@ -46,7 +46,7 @@ ${DTS_TEST_BOARD_MODEL}=                    PRO Z690-A WIFI DDR4(MS-7D25)
 &{DTS_TEST_EXPORTS_PER_WORKFLOW}=
 ...                                         &{DTS_TEST_EXPORTS_PER_WORKFLOW_BASE}
 ...                                         UEFI Update=&{{ {"TEST_IS_COREBOOT": "true"} }}
-...                                         UEFI->Heads Transition=&{{ { "TEST_IS_COREBOOT": "true", "TEST_ME_DISABLED": "false", "TEST_ROMHOLE_MIGRATION_FROM": "flashmap", "TEST_ROMHOLE_MIGRATION_TO": "cbfs" } }}
+...                                         UEFI->Heads Transition=&{{ { "TEST_IS_COREBOOT": "true", "TEST_ME_HAP_DISABLED": "false", "TEST_ME_DISABLED": "true", "TEST_ROMHOLE_MIGRATION_FROM": "flashmap", "TEST_ROMHOLE_MIGRATION_TO": "cbfs" } }}
 ...                                         Initial Deployment=&{{ {"TEST_HCI_PRESENT": "true", "TEST_FMAP_REGIONS": "", "TEST_ROMHOLE_MIGRATION_FROM": "flashmap", "TEST_ROMHOLE_MIGRATION_TO": "flashmap"} }}
 # robocop: off=LEN08
 &{DTS_TEST_EXPORTS_PER_FULL_WORKFLOW}=
--- a/platform-configs/msi-pro-z690-a-wifi-ddr4.robot
+++ b/platform-configs/msi-pro-z690-a-wifi-ddr4.robot
@@ -50,5 +50,5 @@ ${DTS_TEST_BOARD_MODEL}=                    PRO Z690-A WIFI DDR4(MS-7D25)
 ...                                         Initial Deployment=&{{ {"TEST_HCI_PRESENT": "true", "TEST_FMAP_REGIONS": "", "TEST_ROMHOLE_MIGRATION_FROM": "flashmap", "TEST_ROMHOLE_MIGRATION_TO": "flashmap"} }}
 # robocop: off=LEN08
 &{DTS_TEST_EXPORTS_PER_FULL_WORKFLOW}=
-...                                         ${{ ("UEFI Update", "DCR") }}=${{ {"TEST_BIOS_VERSION": "Dasharo (coreboot+UEFI) 0.0.0", "TEST_FMAP_REGIONS": "", "TEST_ME_DISABLED": "false", "TEST_ROMHOLE_MIGRATION_FROM": "flashmap", "TEST_ROMHOLE_MIGRATION_TO": "flashmap" } }}
+...                                         ${{ ("UEFI Update", "DCR") }}=${{ {"TEST_BIOS_VERSION": "Dasharo (coreboot+UEFI) 0.0.0", "TEST_FMAP_REGIONS": "", "TEST_ME_HAP_DISABLED": "false", "TEST_ME_DISABLED": "true", "TEST_ROMHOLE_MIGRATION_FROM": "flashmap", "TEST_ROMHOLE_MIGRATION_TO": "flashmap" } }}
 ...                                         ${{ ("UEFI Update", "DPP") }}=${{ {"TEST_ME_OP_MODE": "2", "TEST_ME_HAP_DISABLED": "true"} }}
--- a/dts/profiles/msi-pro-z690-a-wifi-ddr4 Initial Deployment - DCR.profile
+++ b/dts/profiles/msi-pro-z690-a-wifi-ddr4 Initial Deployment - DCR.profile
@@ -73,6 +73,9 @@ fsread_tool test -d /sys/class/pci_bus/0000:00/device/0000:00:16.0 0
 setpci -s 00:16.0 42.B 0
 cbfstool /tmp/biosupdate extract -r COREBOOT -n config -f /tmp/biosupdate_config 0
 cbfstool /tmp/biosupdate layout -w 0
+cbfstool /tmp/biosupdate print -r COREBOOT 0
+flashrom -p internal -r /tmp/rom.bin --ifd -i bios 0
+cbfstool /tmp/biosupdate write -r ROMHOLE -f /tmp/romhole.bin -u 0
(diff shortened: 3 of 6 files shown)


dts/profiles/msi-pro-z690-a-wifi-ddr4 UEFI->Heads Transition - DPP.profile
DaniilKl, line 1: @miczyg1, you should only review this profile. We need to know whether the commands that are executed (and logged to this profile) for MSI Z690 DDR4 transition from Dasharo (coreboot+UEFI) v1.1.6 to Dasharo (coreboot+Heads) v0.9.0 are expected and correct.

m-iwanicki:
> could not detect that ME is HAP disabled, hence DTS assummed it is soft disabled

It assumed it is enabled, as it is safer assumption and that's why ME wasn't flashed.

platform-configs/msi-pro-z690-a-wifi-ddr4.robot
```diff
@@ -41,3 +41,7 @@ ${DTS_TEST_BOARD_MODEL}=                PRO Z690-A WIFI DDR4(MS-7D25)
 ...                                     ${{ ("UEFI Update", "DCR") }}
 ...                                     ${{ ("Initial Deployment", "DCR") }}
 ...                                     ${{ ("Initial Deployment", "DPP") }}
+
+&{DTS_TEST_EXPORTS_PER_WORKFLOW}=
+...                                     &{DTS_TEST_EXPORTS_PER_WORKFLOW_BASE}
+...                                     UEFI->Heads Transition=&{{ { "TEST_IS_COREBOOT": "true", "TEST_ME_HAP_DISABLED": "false", "TEST_ME_DISABLED": "true" } }}
```
m-iwanicki: You might want to copy and modify `DTS_TEST_EXPORTS_PER_WORKFLOW` from `msi-z690-common`

m-iwanicki (changes requested):

platform-configs/msi-pro-z690-a-wifi-ddr4.robot
DaniilKl: Interesting. On the current `develop` (commit `fe1acf5fc94771bbe344b52cae1a111bfb66d99a`) the mocking configuration for `msi-pro-z690-a-wifi-ddr4 Initial Deployment - DCR` and `msi-pro-z690-a-wifi-ddr4 Initial Deployment - DPP` contain `export TEST_BIOS_VENDOR="3mdeb"`. But as I check the DTS v2.7.2 release test results the E2E tests for these mocking configurations pass. 🤔 . The same value for `TEST_BIOS_VENDOR` will be with the change https://github.com/Dasharo/open-source-firmware-validation/compare/db36e3ef0a2b0eac32c812ec482cc7d2ccb48d2a..38e110fe4a24068aead3fab945e4482f72e300c7 .

DaniilKl: Somehow I thought it should not pass if BIOS informatin contain `3mdeb`, but it seems to be looking for `Dasharo` :man_shrugging: . Because I have checked it manually now and with the follwoing mocking configuration:

<details>

```bash
export TEST_BIOS_VERSION="v0.0.0"
export DTS_CONFIG_REF="refs/heads/main"
export DTS_TESTING="true"
export TEST_SYSTEM_MODEL="MS-7D25"
export TEST_BOARD_MODEL="PRO Z690-A WIFI DDR4(MS-7D25)"
export TEST_SYSTEM_VENDOR="Micro-Star International Co., Ltd."
export TEST_BIOS_VENDOR="3mdeb"
export TEST_CPU_VERSION="TBD_variable_not_set_and_should_be_defined_in_platform_config_if_needed"
export TEST_INTERNAL_PROGRAMMER_CHIPNAME="Opaque flash chip"
export TEST_USING_OPENSOURCE_EC_FIRM="false"
export TEST_BOARD_HAS_BOOTSPLASH="false"
export TEST_VBOOT_KEYS="true"
export TEST_FMAP_REGIONS=""
export TEST_SOUND_CARD_PRESENT="false"
export TEST_BOARD_HAS_GBE_REGION="false"
export TEST_HCI_PRESENT="true"
```

</details>

The `msi-pro-z690-a-wifi-ddr4 Initial Deployment - DCR` passed.

m-iwanicki: remove `DTS_TEST_EXPORTS_PER_WORKFLOW` from `msi-z690-common` as it isn't used at all currently (it is overwritten in each specific platform config). That way it won't confuse others.

miczyg1:
> > could not detect that ME is HAP disabled, hence DTS assummed it is soft disabled
> 
> It assumed it is enabled, as it is safer assumption and that's why ME wasn't flashed.

Assumptions are almost always bad... Why not check with cbmem if ME is disabled? coreboot should print:
`ME is disabled.` When soft disable method is used.

dts/profiles/msi-pro-z690-a-wifi-ddr4 UEFI->Heads Transition - DPP.profile
```diff
@@ -31,12 +30,11 @@ flashrom -p internal 0
...
 cbmem -1 0
-cbmem -1 0
 flashrom -p internal -N --ifd -i bios -r /tmp/bios.bin 0
 cbfstool /tmp/bios.bin layout -w 0
 cbfstool /tmp/biosupdate layout -w 0
 flashrom -p internal -N --ifd -i fd -w /tmp/biosupdate 0
-flashrom -p internal --ifd -i bios -i fd -w /tmp/biosupdate 0
-flashrom -p internal --ifd -i bios -i fd -w /tmp/biosupdate 0
+flashrom -p internal --ifd -i bios -i fd -i me -w /tmp/biosupdate 0
+flashrom -p internal --ifd -i bios -i fd -i me -w /tmp/biosupdate 0
```
miczyg1, line 45: Why the command appears twice?

dts/profiles/msi-pro-z690-a-wifi-ddr4 UEFI->Heads Transition - DPP.profile
miczyg1: Generally looks ok in terms of the chanegs and flashing, but I don't see ROMHOLE migration anywhere in the profile.

Also we seem to extract bootsplash:

```
flashrom -p internal -r /tmp/dasharo_dump.rom --fmap -i FMAP -i BOOTSPLASH 0
cbfstool /tmp/dasharo_dump.rom extract -r BOOTSPLASH -n logo.bmp -f /tmp/logo.bmp 1
```

But we don't do anything with it here.

m-iwanicki:
> Why not check with cbmem if ME is disabled

Similar issue: https://github.com/Dasharo/dasharo-issues/issues/853#issuecomment-2197646733

We do, previously we were testing with ME HAP disabled, which couldn't be detected. Current changes are with ME soft disabled, that's why there is one less cbmem (first one checks if ME is soft disabled, second one checks if ME is HAP disabled)

dts/profiles/msi-pro-z690-a-wifi-ddr4 UEFI->Heads Transition - DPP.profile
m-iwanicki: That's how it was implemented for a long time

* https://github.com/Dasharo/dts-scripts/blob/3b88b776b92559aaa0357d688b75d38bdfee1e2c/scripts/dasharo-deploy.sh#L944
* https://github.com/Dasharo/dts-scripts/blob/3b88b776b92559aaa0357d688b75d38bdfee1e2c/scripts/dasharo-deploy.sh#L987

We would probably need to simplify this logic, or remove identical flashing jobs

dts/profiles/msi-pro-z690-a-wifi-ddr4 UEFI->Heads Transition - DPP.profile
m-iwanicki: Last number is return value of the command, we don't do anything with it, since it failed (likely no `logo.bmp`)

> but I don't see ROMHOLE migration anywhere in the profile

We only do ROMHOLE migration during initial deployment and in transition workflow (UEFI<->SeaBIOS, UEFI->Slimbootloader). Currently Heads transition is baked into update workflow.
Should we add this migration to Heads/UEFI update?

miczyg1: Migration between heads and UEFI (regardless of direction) should preserve ROMHOLE, because it contains board's unique SMBIOS data.

platform-configs/msi-pro-z690-a-wifi-ddr4.robot
DaniilKl: 7e0da33950557c272a00a76927b81c371c45035b , right, thank you for the comment.

m-iwanicki (changes requested):
pre-commit, and remove WIP commit.

dts/profiles/msi-pro-z690-a-wifi-ddr4 UEFI->Heads Transition - DPP.profile
DaniilKl: The ROMHOLE problem has been fixed here: 85e85842fbc3848445f09e447d886e2e26780743 . The BOOTSPLASH problem is not critical and will be addressed here: https://github.com/Dasharo/open-source-firmware-validation/issues/1210 .

DaniilKl:
I have added here several others MSI Z690 a DDR4 profiles regeneration after ROMHOLE fix, becaus we cannot verify them on hardware too.

m-iwanicki (approved):

dts/profiles/msi-pro-z690-a-wifi-ddr4 UEFI->Heads Transition - DPP.profile
miczyg1: I think the problem is that we allow FLASHROM_ADD_OPT_REGIONS even if FLASHROM_ADD_OPT_UPDATE_OVERRIDE is set. Isn't FLASHROM_ADD_OPT_UPDATE_OVERRIDE overriding any flash operations and skip autodetection?

The whole block from lines 966-992 should be moved to line 963 (before the last `fi`).

DaniilKl: Thank you for this note. So far it has not caused any problems. Currently the `FLASHROM_ADD_OPT_UPDATE_OVERRIDE` is [being always overwritten by `FLASHROM_ADD_OPT_REGIONS`](https://github.com/Dasharo/dts-scripts/blob/3b88b776b92559aaa0357d688b75d38bdfee1e2c/scripts/dasharo-deploy.sh#L943), or the [`FLASHROM_ADD_OPT_UPDATE` is being used directly](https://github.com/Dasharo/dts-scripts/blob/3b88b776b92559aaa0357d688b75d38bdfee1e2c/scripts/dasharo-deploy.sh#L954) (except [`RW_SECTION_B`](https://github.com/Dasharo/dts-scripts/blob/3b88b776b92559aaa0357d688b75d38bdfee1e2c/scripts/dasharo-deploy.sh#L961C24-L961C36) which is outside `me` and `fd` that are being [overwritten here](https://github.com/Dasharo/dts-scripts/blob/3b88b776b92559aaa0357d688b75d38bdfee1e2c/scripts/dasharo-deploy.sh#L987)). Hence, it is still a issue, but has a lower priority and will be fixed later.

DaniilKl: https://github.com/Dasharo/dasharo-issues/issues/1742

m-iwanicki: > FLASHROM_ADD_OPT_UPDATE_OVERRIDE is [being always overwritten by FLASHROM_ADD_OPT_REGIONS](https://github.com/Dasharo/dts-scripts/blob/3b88b776b92559aaa0357d688b75d38bdfee1e2c/scripts/dasharo-deploy.sh#L943)

It isn't, one line above we are using: [set_intel_regions_update_params](https://github.com/Dasharo/dts-scripts/blob/3b88b776b92559aaa0357d688b75d38bdfee1e2c/include/dts-functions.sh#L783-L792):

```sh
set_intel_regions_update_params() {
  local fd_me_locked="no"
  if [ $BOARD_HAS_FD_REGION -eq 0 ]; then
    # No FD on board, so no further flashing
    FLASHROM_ADD_OPT_REGIONS=""
  else
    # Safe defaults, only BIOS region and do not verify all regions,
    # as some of them may not be readable. First argument is the initial
    # params.
    FLASHROM_ADD_OPT_REGIONS=$1
```

So `FLASHROM_ADD_OPT_REGIONS` contains `FLASHROM_ADD_OPT_UPDATE_OVERRIDE` and possibly additional elements e.g. `me` and `fd`, unless board doesn't have FD region, then `FLASHROM_ADD_OPT_REGIONS` is empty, which would result in whole binary being flashed.

miczyg1: > Thank you for this note. So far it has not caused any problems. Currently the `FLASHROM_ADD_OPT_UPDATE_OVERRIDE` is [being always overwritten by `FLASHROM_ADD_OPT_REGIONS`](https://github.com/Dasharo/dts-scripts/blob/3b88b776b92559aaa0357d688b75d38bdfee1e2c/scripts/dasharo-deploy.sh#L943), or the [`FLASHROM_ADD_OPT_UPDATE` is being used directly](https://github.com/Dasharo/dts-scripts/blob/3b88b776b92559aaa0357d688b75d38bdfee1e2c/scripts/dasharo-deploy.sh#L954) (except [`RW_SECTION_B`](https://github.com/Dasharo/dts-scripts/blob/3b88b776b92559aaa0357d688b75d38bdfee1e2c/scripts/dasharo-deploy.sh#L961C24-L961C36) which is outside `me` and `fd` that are being [overwritten here](https://github.com/Dasharo/dts-scripts/blob/3b88b776b92559aaa0357d688b75d38bdfee1e2c/scripts/dasharo-deploy.sh#L987)). Hence, it is still a issue, but has a lower priority and will be fixed later.

FLASHROM_ADD_OPT_UPDATE_OVERRIDE is the starting point for FLASHROM_ADD_OPT_REGIONS, i.e. contains the must-have regions. `set_intel_regions_update_params` is only adding mroe regions to FLASHROM_ADD_OPT_REGIONS if possible to flash them. That is why FLASHROM_ADD_OPT_UPDATE_OVERRIDE gets overwritten by FLASHROM_ADD_OPT_REGIONS.

This doesn't change the fact that the block I mentioned should be moved. Because whatever regions were added by `set_intel_regions_update_params` to FLASHROM_ADD_OPT_REGIONS, will be included in FLASHROM_ADD_OPT_UPDATE_OVERRIDE and flashed. No need to flash it second time with FLASHROM_ADD_OPT_REGIONS in lines 944/987

DaniilKl: I am resolving this thread, further discussions should be done in the issue https://github.com/Dasharo/open-source-firmware-validation/pull/1203#discussion_r2716577135 .

PR #1221, opened 2026-01-28
Creating custom linux bootentries

philipanda:
This prevents the bootoerder from changing during tests via SSH.
The `DEFAULT_BOOT_OS_ID` should always stay as the first bootentry and always boot after a reboot/warmboot/coldboot etc.
Running Basic Platform Setup is required to setup non-volatile default boot os.


filipleple (approved):
LGTM

PR #1225, opened 2026-02-02
OSFV stable release infrastructure

### LINKED ISSUE #1224: OSFV Stable release preparation checklist
- [x] CI respects SnipeIT
- [x] Regression CI allows for pre-defined test lists instead of running depending on the changes
- [ ] ~~Separating the CI into multiple workflows, one per platform, so that they are independent. Now they either run all at once or none~~ won't do now, no need to fiddle with gh actions for now
- [x] Defining new platforms in `scripts/regression-scope` configs, they will not be used for PR tests, but will be there to run a big regression prior to planned OSFV releases
- [ ] optional name/terminology change of `scripts/ci/regression-scope` as it will be more multifunctional and the name might be misleading
- [x] Tools/scripts for quick text summary of regression, % of passes, grouping per platfrom, per suite. Mostly `rebot` + pandas
- [x] Potentially some script/action that will run the full release-grade regression on target machines. 
   - A) be async and manage running actions - probably overcomplicated
   - B) be able to save some state, like action IDs to come back later and get the results of the actions it triggered - seems more reasonable
      - maybe some GH actions mechanism could help, like child actions or smth
   - No matter the solutions it should be able to produce viable results even if tests fail for any reason like platforms being dead. It should be able to continue a failed run based on results, maybe even automatically re-run failed tests or flash the device if it fails like 2 times a row - that would be great
   - should be able run the whole regression a couple times to gather info about results variation should there be such need

philipanda:
Scripts and CI workflows for OSFV release tests and documenting stability.
Check README.md changes


philipanda:
`osfv_stability_run.py` is a wrapper for `develop_pr_auto_regression.sh` which reuses the code made for CI on PR.

It runs the suites defined in `scripts/ci/regression-scope/configs/tests-list.txt` on devices from `scripts/ci/regression-scope/configs/pr-regression-devices.csv`. 

It produces logs in `ci-logs/` directory, which then can be parsed by `osfv_stability_reports.py`, which just prints some stats for now, the frontend can be adjusted later if needed or saved into some CSVs:
image

scripts/ci/regression-scope/configs/devices/msi-pro-z690-a-wifi-ddr4_00039.json
```diff
@@ -0,0 +1,11 @@
+{
+    "env_vars": {
+        "ASSET_ID": "00039",
+        "RTE_IP": "192.168.10.199",
+        "CONFIG": "msi-pro-z690-a-wifi-ddr4",
+        "SONOFF_IP": "192.168.10.169",
+        "PIKVM_IP": "192.168.10.120",
+        "DEVICE_IP": "192.168.10.155",
+        "FW_FILE": "_roms/msi_ms7d25_v1.1.6_ddr5_dev_signed.rom"
```
filipleple, line 9: @philipanda  I think it has to be:
```suggestion
        "FW_FILE": "_roms/msi_ms7d25_v1.1.6_ddr4_dev_signed.rom"
```

I think the DDR5 ROM isn't compatible with the DDR4 board?

scripts/ci/regression-scope/configs/devices/protectli-vp6650_00002.json
```diff
@@ -0,0 +1,10 @@
+{
+    "env_vars": {
+        "ASSET_ID": "00002",
```
filipleple, line 3: So shouldn't the filename be `protectli-vp6650_00002.json`?

scripts/ci/regression-scope/configs/devices/pcengines-apu3_00252.json
```diff
@@ -0,0 +1,9 @@
+{
+    "env_vars": {
+        "ASSET_ID": "00252",
+        "RTE_IP": "192.168.10.171",
+        "CONFIG": "pcengines-apu3",
+        "DEVICE_IP": "192.168.10.197",
+        "FW_FILE": "_roms/pcengines_apu6_v0.9.1_dev_signed.rom"
```
filipleple, line 7: Shouldn't it be

```suggestion
        "FW_FILE": "_roms/pcengines_apu3_v0.9.1_dev_signed.rom"
```

since there is such file on cloud? again, I'm not up to date on APU binary compatibility so pls feel free to resolve

filipleple (changes requested):

scripts/ci/regression-scope/lib/snipeit_checkout.sh
```diff
@@ -0,0 +1,58 @@
...
+snipeit_cleanup() {
+    MAX_TRIES=10
+    echo "To check_in: ${CHECKED_OUT_DEVICES[*]}"
+    for id in "${CHECKED_OUT_DEVICES[@]}"; do
+        trial=0
+        out=$(osfv_cli snipeit check_in --asset_id $id)
+        rc=$?
+        while [[ $rc != 0 ]]; do
+            if [ "$trial" -ge "$MAX_TRIES" ]; then
+                echo "Failed ${MAX_TRIES} times, skipping check_in"
```
filipleple: It looks to me like the value of `rc` is never changed in this loop? and the checkin command is never rerun?

So it either succeeds at first attempt, or goes into an infinite loop of printing `failed to check in`

philipanda: That's true, luckily that's dead code I've left over by mistake during rebase

philipanda: https://github.com/Dasharo/open-source-firmware-validation/compare/591910ccf0ed0f90220a042de5d4cb375219b499..06af23086d8464d813ab73a970c6c6f29691f680

scripts/ci/regression-scope/configs/devices/protectli-vp6650_00002.json
philipanda: https://github.com/Dasharo/open-source-firmware-validation/pull/1225/changes/14f7de42a9992ba0181c13ebf0526fecf838d7da

filipleple (approved):

PR #1233, opened 2026-02-10
Osfv stable release improvements

### LINKED ISSUE #1211: Detect booted OS (SSH tests) (OSFV stability)
### The problem you're addressing (if any)

OSFV assumes that the DEFAULT_BOOT_OS_ID is booted when starting tests. After test run ends, the device will be booted to an unknown OS.

### Describe the solution you'd like

Detect which OS is booted, for example by trying all ssh user/password pairs we have.

### Where is the value to a user, and who might that user be?

OSFV stability - the tests could continue without the need to go back to DEFAULT_BOOT_OS

### Describe alternatives you've considered

cleaning up - returning back to DEFAULT_BOOT_OS after test ends
 - cant be guaranteed to happen - less robust
 - takes more time

### Additional context

_No response_

### LINKED ISSUE #1240: OSFV Stable release improvements
- [x] https://github.com/Dasharo/open-source-firmware-validation/issues/1211
- [x] Fix UEFI  Shell deployment which sometimes breaks bootorder
- [x] take care of Capsule Update test suite (CUP) to allow running it remotely without constant supervision and manual steps
- more TODO

philipanda:


README.md
```diff
@@ -453,6 +454,51 @@ necessary corrections. Example workflow:
...
+from `scripts/ci/regression-scope/configs/release_tests_devices.txt`, more
+precisely defined in `scripts/ci/regression-scope/devices/`.
+It reuses the system used for automatic CI runs on PRs.
+
+The `osfv_stability_run.py` script will try to run the whole test scope on
+all supported devices 2 times and wait until they're free for check-out on
+Snipe-IT. It can easily take multiple hours, so always make sure to run it
+on a stable device that always has the connection to the lab network.
+
+Use `LOGS_DIR` envv variable to redirect logs to NFS for future reference:
```
macpijan:
```suggestion
Use `LOGS_DIR` env variable to redirect logs to NFS for future reference:
```

macpijan (commented):
I just skimmed through it, won't test right now. Overall sounds ok to me. I am most interested in seeing some results - e.g. how reports from this run look like, whether we have any improvement already due to the changes introduced here.

README.md
```diff
@@ -453,6 +454,51 @@ necessary corrections. Example workflow:
...
+After the results are created, use `scripts/ci/osfv_stability_reports.py` to
+create a summary of all the runs. It can take a couple minutes to parse all
+that logs.
+
+Use `LOGS_DIR` env var to direct the parser to the directory with logs.
+E.g. The NFS with all the historic official runs on OSFV releases.
+`export LOGS_DIR=/srv/nfs/logs/osfv_stability/ci_logs`
+
+Use `--json` option to format the output in json for easy processing
+`./scripts/ci/osfv_stability_reports.py --json > reports.json`
```
macpijan, line 500: Can we easily compare these reports and gather some stats out of that? Example use-case:
- we run this right now and record results
- we make some improvements according to our hypothesis that should improve resiliency and stability
- we run again to verify - we need an easy method to compare these results.

philipanda: We can just compare the numbers saved in the json for the historic runs. Would you like something that compares them too?
That would be suitable for the PR https://github.com/Dasharo/open-source-firmware-validation/pull/1225

README.md
philipanda: https://github.com/Dasharo/open-source-firmware-validation/compare/4c6588bf81854eddb17a7285dfd58266ca5005a4..b1818bff4f910f3308007324a7a5ae4e04232793

philipanda:
## Booted OS recovery fallback

When testing via SSH it sometimes so happens that we get ourselves into an unexpected Os.
This should never happen unless smoe errors or unhandled exceptions happen.
It will probably still happen pretty often, especially it might not be possible to guarantee we always end up in a known Os every time when running the whole regression via SSH.
To counteract that we can try to recover from an unknown state by trying to log in into all known Oses if logging in to the expected one fails.


Runtimes of KW "Login To Linux" were analysed to find the best
time to fallback to rotating the username&password across different OSes.
Only `PASS`es were analysed, as we're interested in how much to wait to have
a reasonable chance of continuing the test with given credentials.


V540tu v1.0.1-rc3

Keyword: Login To Linux
Executions: 286
Average runtime (s): 10.647
Median runtime (s): 0.364
Stddev runtime (s): 24.841
Max runtime (s): 287.088
Min runtime (s): 0.221
100-th percentile (s): 141.326

Keyword: Login To Linux Via SSH
Executions: 284
Average runtime (s): 10.122
Median runtime (s): 0.360
Stddev runtime (s): 23.772
Max runtime (s): 287.084
Min runtime (s): 0.218
100-th percentile (s): 103.107


V540tnd v1.0.0-rc10

Keyword: Login To Linux
Executions: 1175
Average runtime (s): 23.970
Median runtime (s): 0.305
Stddev runtime (s): 77.692
Max runtime (s): 875.654
Min runtime (s): 0.203
100-th percentile (s): 437.272

Keyword: Login To Linux Via SSH
Executions: 961
Average runtime (s): 17.269
Median runtime (s): 0.284
Stddev runtime (s): 55.271
Max runtime (s): 469.345
Min runtime (s): 0.200
100-th percentile (s): 292.312

about 300s, or 30 tries 10 seconds wait between each should be more than enough to cover 99% cases.

### Implementation

Were using different keywords for logging into OSes.
1. Login To Linux
2. Login To Linux Via SSH
3. Login To Windows
4. Login To Windows Via SSH
5. SSHLibrary.Login
6. Login To OS

Instead of forcing every usage, It might be best to rework the `Login To OS` keyword
so that it uses the BOOTED_OS_ID context and implements the fallback if needed.
Keywords like `Login To Windows` would stay untouched as they might rely on
NOT logging to an OS different than Windows.

Logs from Basic Platform Setup, the device was purposely booted into Windows while the DEFAULT_BOOT_OS_ID references Ubuntu. OSFV was able to recover from this state and continue the suite.
[basic-platform-setup.html](https://github.com/user-attachments/files/25212348/basic-platform-setup.html)

philipanda:
## UEFI Shell breaks bootorder
1. BPS runs uefi shell deployment script unsupervised in ansible
2. `Deploy Uefi Shell` is used in tests to deploy it if needed
3. Removing this step from ansible and doing it in BPS using RF keywords with bootorder guards

https://github.com/Dasharo/open-source-firmware-validation/pull/1233/changes/0f2cae2383921d5b9d1a8c4d0e73d46a862e528d
https://github.com/Dasharo/open-source-firmware-validation/pull/1233/changes/7e719420417f911d3ca3bfac98276d9cdc6b9ef6
https://github.com/user-attachments/files/25212348/basic-platform-setup.html

philipanda:
## Updating packages as part of basic platform setup
1. implemented in ansible, only apt and dnf supported for now
https://github.com/Dasharo/open-source-firmware-validation/pull/1233/changes/5f68afde438d11d6ee75e8590ccc59fd7f7d93d2

philipanda:
# take care of Capsule Update test suite (CUP) to allow running it remotely without constant supervision and manual steps

Capsule Update tests now pass completely unattended without any additional setup like attaching usb drives or running any scripts etc. making it suitable for CI. And it works without serial connection while not being more complex than the serial console variants. The same mechanism as for deploying  UEFI Shell is being used. A directory is being created on the boot partition of the system drive. The directory contains the UEFI Shell app, to which a bootentry is created, and contains all the files needed for capsule update tests via SSH.

All that is required for the tests to rungiven three files are provided to robot:
- tested firmware (`FW_FILE`)
- tested firmware capsule (`CAPSULE_FW_FILE`)
- some previous firmware to run the update from, e.g. the RC0 (`CAPSULE_UPDATE_RC0_FW_FILE`)

I'd suggest removing any other execution paths (via serial console) as they don't help with anything and only create more bug vectors. Tests via SSH are pretty much universal and will work on any device.

[merged_log.html](https://github.com/user-attachments/files/25268922/merged_log.html)

wiktormowinski:
[merged_log.html](https://github.com/user-attachments/files/25417706/merged_log.html)
@philipanda there is some "to boot directly" action going on when running the capsule update suite on vp6670
CUP001 CUP002 and CUP150 are affected 

All 3 have this `Perform Capsule Update` kwd. What happens is that:
-  `Perform Caspsule Update`>` '${OPTIONS_LIB}' == 'options-lib_uefi-setup-menu' `>`Read From Terminal Until`  reads `to boot directly` 
- the next kwd: `Perform Capsule Update` > `Boot And Login To OS` is expecting `to boot directly` again, even though the DUT already booted ubuntu


**Tested on:
RC0 FILE:** 0.9.3-rc3
**Capsule file:** rc4 -  rom, built and converted to capsule on [this commit](https://github.com/Dasharo/coreboot/commit/e617bc6b6cf5949d9169165cb89e8a699a96a21c)
**Platform:** protectli vp6670 (asset 00004)

philipanda:
> [merged_log.html](https://github.com/user-attachments/files/25417706/merged_log.html) @philipanda there is some "to boot directly" action going on when running the capsule update suite on vp6670 CUP001 CUP002 and CUP150 are affected
> 
> All 3 have this `Perform Capsule Update` kwd. What happens is that:
> 
> * `Perform Caspsule Update`>`'${OPTIONS_LIB}' == 'options-lib_uefi-setup-menu'`>`Read From Terminal Until`  reads `to boot directly`
> * the next kwd: `Perform Capsule Update` > `Boot And Login To OS` is expecting `to boot directly` again, even though the DUT already booted ubuntu
> 
> **Tested on: RC0 FILE:** 0.9.3-rc3 **Capsule file:** rc4 - rom, built and converted to capsule on [this commit](https://github.com/Dasharo/coreboot/commit/e617bc6b6cf5949d9169165cb89e8a699a96a21c) **Platform:** protectli vp6670 (asset 00004)

Let's unwrap this:

First we're in Ubuntu, changing nextboot to uefi shell, reboot
***FIRST** read from terminal until.... - we need to skip this one to allow UEFI Shell to boot and do its thing
UEfi shell reboots after loading capsule:
***SECOND** read from terminal until... - we need to skip that one too, after this one either FUM screen appears, or not, 
regardless the Capsule update should start and and reboot the device whether it succeeds or not
***THIRD** read from terminal until... - now we're after the capsule update and we can boot an OS

From what I can see the amount of expected appearings of `to boot directly` is OK. If that fails, then it looks like the capsule update was not started at all, or there was no reboot after it and the device went directly to an OS.

Could you please check how this looks in practice? Maybe there's one less reboot on protectli devices? Or maybe the capsule file was corrupted and the CapsuleApp did not even try to load it? Or maybe the UEFI Shell was not deployed correctly? From what I can see there's something happening after booting UEFI Shell, because the device reboots just like if the CapsuleApp has loaded the capsule, but I can't see any output from UEFI Shell in the logs. Could you verify if you can boot UEFI Shell and start the capsule update using the files in `fsX:\capsule_testing` manually?

I've been testing that thoroughly on an MSI z690 and a Novacustom v540tnd and it was working perfectly, so something clearly must different here.

PR #1242, opened 2026-02-13
tpm.robot: Validate PCRs Against Event Log: Accept some pcrs being invalid

philipanda:
*ref: prot-1820*

### DIFF
--- a/lib/tpm.robot
+++ b/lib/tpm.robot
@@ -4,6 +4,10 @@ Library     OperatingSystem
 Library     String
 
 
+*** Variables ***
+@{ACCEPTABLE_PCR_FAILS}=    10
+
+
 *** Keywords ***
 Get PCRs State From Linux
     [Documentation]    Returns list of strings containing
@@ -59,7 +63,8 @@ Validate PCRs Against Event Log
     ${tpm2_eventlog}=    Execute Command In Terminal
     ...    tpm2_eventlog ${binary_log_file}
     Should Not Contain    ${tpm2_eventlog}    ERROR: Unable to run tpm2_eventlog
-    VAR    @{invalid_prcs}=    @{EMPTY}
+    VAR    @{acceptable_invalid_prcs}=    @{EMPTY}
+    VAR    @{critical_invalid_pcrs}=    @{EMPTY}
     FOR    ${algo}    IN    sha1    sha256
         ${eventlog_pcrs}=    Get PCRs From Eventlog    ${tpm2_eventlog}    ${algo}
         FOR    ${pcr_element}    IN    @{eventlog_pcrs}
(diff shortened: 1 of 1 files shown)


philipanda:
The tests were failing when PCR10 reconstruction was turning out invalid before:
[merged_log.html](https://github.com/user-attachments/files/25286356/merged_log.html)
```
Measured-Boot                                                                 
MBO001.201 Measured Boot support (Ubuntu) :: Check whether Measure... | PASS |
MBO002.201 Check if event log PCRs match actual values (Ubuntu) ::... ..[{'pcr': 'pcr10-sha256', 'log_hash': '0xdc54f29f4bf4c7584a8d9915b09359959f8e13f13009c1eab4b00c21b85f1930', 'real_hash': '1ED747252BF06B3D0ECCDB51F4126D76193C8AE51B5FA22BA5DC1D768E3A9F8C'}]
MBO002.201 Check if event log PCRs match actual values (Ubuntu) ::... | FAIL |
Invalid PCRs found
Since the asset has been checkout manually by you prior running this script, it will NOT be checked in automatically. Please return the device when work is finished.
Measured-Boot                                                         | FAIL |
2 tests, 1 passed, 1 failed
```


Now there's a list of PCR numbers that will be accepted by OSFV and only a warning will be stated:
[merged_log.html](https://github.com/user-attachments/files/25286361/merged_log.html)
```
Measured-Boot                                                                 
MBO001.201 Measured Boot support (Ubuntu) :: Check whether Measure... | PASS |
[ WARN ] Invalid PCRs found, but they don't necessarily mean a firmware issue 
MBO002.201 Check if event log PCRs match actual values (Ubuntu) ::... [{'pcr': 'pcr10-sha256', 'log_hash': '0xdc54f29f4bf4c7584a8d9915b09359959f8e13f13009c1eab4b00c21b85f1930', 'real_hash': '1ED747252BF06B3D0ECCDB51F4126D76193C8AE51B5FA22BA5DC1D768E3A9F8C'}]
MBO002.201 Check if event log PCRs match actual values (Ubuntu) ::... | PASS |
Since the asset has been checkout manually by you prior running this script, it will NOT be checked in automatically. Please return the device when work is finished.
Measured-Boot                                                         | PASS |
2 tests, 2 passed, 0 failed
```

lib/tpm.robot
```diff
@@ -4,6 +4,10 @@ Library     OperatingSystem
 Library     String
 
 
+*** Variables ***
+@{ACCEPTABLE_PCR_FAILS}=    10
```
SergiiDmytruk, line 14: I think this needs a comment explaining why a failure to reconstruct PCR-10 is not a problem:
 - this PCR is not supposed to be extended by firmware
 - it's the default PCR used by IMA implementation in Linux, which uses a separate log for its measurements
 - if compiled into the kernel, IMA cannot be completely disabled at runtime

philipanda: https://github.com/Dasharo/open-source-firmware-validation/pull/1242/changes/fe912bd2c5fd2861403c5963d8cd5802503937cb

SergiiDmytruk (approved):

PR #1243, opened 2026-02-17
platform-configs/novacustom-*: Bump benchmark results for 155h platforms

mkopec:
Update benchmark values with ones from https://openbenchmarking.org/vs/Processor/Intel+Core+Ultra+7+155H .
After recent performance fixes, the performance exceeds the previous values. Set the expected values to average 155h performance.

*ref: ncm-2109*

### DIFF
--- a/platform-configs/novacustom-nuc_box-155H.robot
+++ b/platform-configs/novacustom-nuc_box-155H.robot
@@ -24,11 +24,11 @@ ${USB_MODEL}=                           ${TBD}
 ${USB_DEVICE}=                          Linux
 
 # cpu performance Ubuntu
-${ZIP_MULTI_COMPRESSION}=               63476    # MIPS
-${ZIP_MULTI_DECOMPRESSION}=             39336    # MIPS
-${CRAY_5_K_RENDER}=                     654.5    # sec
-${CRAY_4_K_RENDER}=                     356.9    # sec
-${CRAY_1080_P_RENDER}=                  90.8    # sec
+${ZIP_MULTI_COMPRESSION}=               69245    # MIPS
+${ZIP_MULTI_DECOMPRESSION}=             47858    # MIPS
+${CRAY_5_K_RENDER}=                     549    # sec
+${CRAY_4_K_RENDER}=                     319    # sec
+${CRAY_1080_P_RENDER}=                  79    # sec
 ${COREMARK_SINGLE}=                     400079.5    # iterations/s
 
 # cpu performance Windows
--- a/platform-configs/novacustom-nuc_box-155H.robot
+++ b/platform-configs/novacustom-nuc_box-155H.robot
@@ -3,6 +3,7 @@ Resource    include/novacustom-nuc_box.robot
 Resource    include/novacustom-mtl.robot
 Resource    include/novacustom-common.robot
 Resource    include/default.robot
+Resource    include/cpu-perf-155h.robot
 
 
 *** Variables ***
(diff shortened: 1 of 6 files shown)


platform-configs/novacustom-nuc_box-155H.robot
```diff
@@ -24,11 +24,11 @@ ${USB_MODEL}=                           ${TBD}
 ${USB_DEVICE}=                          Linux
 
 # cpu performance Ubuntu
-${ZIP_MULTI_COMPRESSION}=               63476    # MIPS
-${ZIP_MULTI_DECOMPRESSION}=             39336    # MIPS
-${CRAY_5_K_RENDER}=                     654.5    # sec
-${CRAY_4_K_RENDER}=                     356.9    # sec
-${CRAY_1080_P_RENDER}=                  90.8    # sec
+${ZIP_MULTI_COMPRESSION}=               69245    # MIPS
```
SergiiDmytruk: Should https://openbenchmarking.org/vs/Processor/Intel+Core+Ultra+7+155H be mentioned in the comment like for some other sections in these files?

platform-configs/novacustom-v560tnd.robot
```diff
@@ -53,11 +53,11 @@ ${DGPU_ONLY_SUPPORT}=                   ${TRUE}
 ${DISK_IO_PERFORMANCE_TESTS}=           ${TRUE}
 
 # cpu performance Ubuntu
-${ZIP_MULTI_COMPRESSION}=               79729    # MIPS
-${ZIP_MULTI_DECOMPRESSION}=             52410    # MIPS
-${CRAY_5_K_RENDER}=                     585.333    # sec
-${CRAY_4_K_RENDER}=                     326.895    # sec
-${CRAY_1080_P_RENDER}=                  80.547    # sec
+${ZIP_MULTI_COMPRESSION}=               69245    # MIPS
+${ZIP_MULTI_DECOMPRESSION}=             47858    # MIPS
```
SergiiDmytruk: So the expected CPU performance went down for V56 variants despite configuration being adjusted to increase the performance?  I've noticed that `COREMARK_SINGLE` wasn't updated and thought that's because it's higher here than on OpenBenchmarking.

In general, do we expect 155H behave the same for all devices?  If so, maybe all this should be defined once in a file like `include/cpu-perf-155h.robot` to make updates simpler.

mkopec: Done: https://github.com/Dasharo/open-source-firmware-validation/pull/1243/commits/f30fa95d1c6babba851488cc6baa3a59ef5b7b2f

Updated coremark results to match OpenBenchmarking and moved to separate include. 

> So the expected CPU performance went down for V56 variants despite configuration being adjusted to increase the performance? 

I'm not sure how the original results were obtained. I think we should expect to be within some margin from the average 155H results in any case.

platform-configs/novacustom-nuc_box-155H.robot
mkopec: Added https://github.com/Dasharo/open-source-firmware-validation/commit/f30fa95d1c6babba851488cc6baa3a59ef5b7b2f

platform-configs/novacustom-v560tnd.robot
SergiiDmytruk: Forgot to update `platform-configs/novacustom-v560tne.robot` in the new commit?

PR #1256, opened 2026-03-13
scripts: add FW_URI support for regression firmware download

### LINKED ISSUE #980: Automate managing firmware binaries in OSFV
### The problem you're addressing (if any)

Right now the testers are downloading the firmware binary for testing from *some* source.
The preferred way is to either download from dl.3mdeb.com, or from the corresponding repo (most often coreboot) CI system.

This allows for user error if the tester forgets to get and flash the correct firmware binary from the correct source or reuses an old binary they already have without making sure it is exactly the same as the one that should be tested.

### Describe the solution you'd like

Instead of `FW_FILE=<path>`, add a way to give the version of the binary and get it automatically.
For example: 
- `FW_REPO=https://github.com/dasharo/coreboot`, `FW_VERSION=novacustom_v54x_mtl_v0.9.1` - tag name
- `FW_URI=https://dl.3mdeb.com/open-source-firmware/Dasharo/novacustom_v54x_mtl/v0.9.1/novacustom_v54x_mtl_v0.9.1.rom` - direct download URL

The `FW_FILE` option must obviously not be removed, because it would hinder debugging locally

Ideally the binaries should be cached somewhere in order to reduce unnecessary downloads.

Might be related to https://github.com/Dasharo/open-source-firmware-validation/issues/632, as making OSFV aware of firmware binary versions would allow to set some DMIDECODE parameters automatically.

### Where is the value to a user, and who might that user be?

Better reproducibility
Freeing the testers from managing firmware release binaries
Separating release binaries from possibly multiple debug binaries built by the tester, which might not be renamed by them and be mistaken for the release binaries

### Describe alternatives you've considered

_No response_

### Additional context

Problems:
- Not every Dasharo flavor builds the firmware in CI:
  - https://github.com/dasharo/heads
  - https://github.com/dasharo/seabios

WuXieSec:
## Summary
Add `FW_URI` support to regression wrappers so testers can provide a direct firmware download URL instead of managing local binary paths manually.

## What changed
- added `resolve_fw_file` helper in `scripts/lib/robot.sh`
  - accepts either `FW_FILE` (existing behavior) or `FW_URI`
  - downloads from `FW_URI` to cache (`$FW_CACHE_DIR` or `~/.cache/osfv/firmware`)
  - reuses cached file on subsequent runs
  - exports resolved path as `FW_FILE`
- switched wrappers to use resolver:
  - `scripts/regression.sh`
  - `scripts/regression-rerun-failed.sh`
- updated README with `FW_URI` usage example

## Why
This addresses issue #980 goal of reducing manual firmware file handling and improving reproducibility while preserving `FW_FILE` for local debugging.

Closes #980

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,6 +226,8 @@ Parameters should be defined as follows:
 * $FW_FILE - path to and name of the coreboot firmware file. This is usually
   not required when running single tests or suites, where flashing is not
   necessary.
+* $FW_URI - optional direct URL to firmware binary used by regression wrappers
+  when `$FW_FILE` is not set. Downloaded once and cached locally.
 * $CONFIG - platform config - see the `platform-configs` directory for
   available configurations.
 * $TEST_MODULE - name of the test module (i.e. `dasharo-compatibility`),
@@ -295,6 +297,12 @@ platform config file.
 
 ```bash
 FW_FILE=$FW_FILE DEVICE_IP=$DEVICE_IP RTE_IP=$RTE_IP CONFIG=$CONFIG ./scripts/regression.sh
+```
+
+Or provide a direct download URL instead of a local file:
+
+```bash
+FW_URI=https://dl.3mdeb.com/path/to/firmware.rom DEVICE_IP=$DEVICE_IP RTE_IP=$RTE_IP CONFIG=$CONFIG ./scripts/regression.sh
 ```
 
 Running regression tests without snipeit works the same way as
--- a/scripts/lib/robot.sh
+++ b/scripts/lib/robot.sh
@@ -19,6 +19,45 @@ check_env_variable() {
   fi
 }
 
+resolve_fw_file() {
+  # Accept either FW_FILE (local path) or FW_URI (remote URL downloaded to cache)
+  if [ -n "${FW_FILE}" ]; then
+    if [ ! -f "${FW_FILE}" ]; then
+      echo "Error: Environment variable FW_FILE doesn't point to a file."
+      exit 1
+    fi
+    return 0
+  fi
+
+  if [ -z "${FW_URI}" ]; then
+    echo "Error: provide FW_FILE (local path) or FW_URI (download URL)."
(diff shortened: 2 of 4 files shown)


WuXieSec:
Rebased on latest `develop` and force-pushed to remove `BEHIND` status.

Updated head commit: `88e2e28b`

Could you please re-review when you have a moment? Thanks!

WuXieSec:
Friendly follow-up on this PR.

It has been rebased on the latest `develop` branch and is ready for review. Could you please take a look when convenient?

WuXieSec:
Quick ping on this one. It is rebased, CI is clean, and I am happy to adjust anything if you want the scope trimmed further.

Rishet11:
Two firmware-cache cases reproduce at `88e2e28bcb160af857e0fa07888662ec2858fae4`:

- **Different URLs reuse the wrong bytes.** In `scripts/lib/robot.sh:40-47`, `/a/firmware.rom` and `/b/firmware.rom` both map to `cache/firmware.rom`. After downloading A, requesting B returns A without contacting B.
- **A failed download becomes a successful cache hit.** Curl writes directly to the final path at lines 46-55. A truncated HTTP response produces curl error 18 and leaves a partial file. Retrying the **same URL** after the server recovers returns that partial file without another request.

[Reproduction script, setup and observed output](https://gist.github.com/Rishet11/4da3ca139c584d2abefdb9e49edeca41). It executes the unchanged resolver with real `/usr/bin/curl` and synthetic HTTP responses on `127.0.0.1`. Both cases were independently rerun by the orchestrating AI assistant. This is automated resolver validation, not human or device testing.

A URI-based cache key and download-to-temporary-file followed by rename on success would address the two cases. The existing PR remains unchanged.

Could a separate regression-test follow-up be assigned under #980, and would this validation qualify for the bounty program's financial support? Please confirm the scope and support amount before implementation; neither has been agreed.

PR #1276, opened 2026-07-05
feat: documentation build-test proof of concept (dasharo-issues#1153)

ameyypawar:
## Summary

A proof of concept toward Dasharo/dasharo-issues#1153 (automatically test that
following the build documentation reproduces the published firmware). It's a
**draft**, and I'm treating it as concrete input to the discussion on that
issue rather than a finished solution — happy to move, reshape, or close it.

## Where this fits (and what it deliberately doesn't do)

This follows the "parse the docs, then hash-compare" plan @macpijan sketched in
#545. There's an alternative in that thread — standardising the build (a
universal script like Dasharo/coreboot#579, or Jinja-generated docs) so parsing
becomes unnecessary — and this PoC **doesn't preclude** it; it tackles the
"read the existing docs" slice.

It intentionally does **not** attempt the harder half of the issue:

- **Fresh-OS dependency capture** — verifying the steps run on a clean OS with
  nothing missing (the `sudo`/toolchain gap hit when trying doc-detective).
  That's the part that guarantees "no dependency is missing from the docs," and
  it's out of scope here.
- Running the (Docker, multi-minute) builds in CI, and discovering the
  published release set + hashes.

## Relationship to Dasharo/docs#1240

@WuXieSec's Dasharo/docs#1240 ("phase-1 build-docs command checks") is already
in this space, in the docs repo. This harness is **complementary** — it
resolves a single build path through the MkDocs tabbed decision tree, generates
a runnable build script, and adds a reproducibility comparison — but overlaps
on the "extract commands from building-manual.md" part. I'd like to reconcile
with it rather than duplicate; guidance on whether this belongs here (osfv) or
in docs alongside Dasharo/docs#1240 would help.

## What the tool does

A `building-manual.md` page is a MkDocs Material decision tree of
`pymdownx.tabbed` tabs (firmware x device x options). Extracting *every* fenced
block concatenates mutually exclusive branches; this resolves a **single path**
for a selection, following MkDocs' `content.tabs.link` semantics.

- `list` / `extract` / `script` — enumerate targets, resolve one, emit a
  runnable build script.
- `verify` — sha256 compare; on a difference it surfaces `romscope compare`
  output to interpret (it does **not** guess a verdict — a correct reproducible
  build legitimately differs from the signed release).
- `diagnose` — flags doc patterns that block mechanical testing (version
  choices hidden in prose, inconsistent checkout/build tab labels, ambiguous
  selections). Findings come from running it against the live manuals; the
  committed tests use small inlined fixtures.

Standard library only; `pre-commit` (black, isort, codespell, markdownlint,
reuse) passes.

## Asking first

@macpijan — you asked people to discuss on the issue before picking this up,
so: is this direction useful, and if so where should it live and how should it
relate to Dasharo/docs#1240? I'll follow your lead.

### DIFF
--- a/scripts/ci/doc-build-test/README.md
+++ b/scripts/ci/doc-build-test/README.md
@@ -7,8 +7,9 @@ SPDX-License-Identifier: Apache-2.0
 # Documentation build test
 
 A proof-of-concept harness that reads a Dasharo build manual from
-`docs.dasharo.com`, resolves the exact commands a reader would run to build
-one firmware variant, and checks the result against the published release.
+`docs.dasharo.com` and resolves the exact commands a reader would run to build
+one firmware variant, so a resulting binary can be compared against a published
+release.
 
 It is a first step towards
 [dasharo-issues#1153](https://github.com/Dasharo/dasharo-issues/issues/1153):
@@ -38,8 +39,10 @@ Off-the-shelf "runnable docs" tools such as `codedown`, `doc-detective` and
 `tuttest` extract _every_ fenced block on the page and run them in order. On
 the page above that means running the DDR4 build, the DDR5 build and the Heads
 build back to back - incompatible branches concatenated into one broken
-script. This is why a plain run of such a tool against the Dasharo docs has
-mostly failed in earlier experiments on the issue.
+script. Branch concatenation is one failure mode. A separate one the
+maintainers hit when trying doc-detective - its container lacked host
+dependencies the docs assume, such as `sudo` - is about capturing what a fresh
+OS needs, and is out of scope here (see Scope below).
 
 This harness instead resolves a _single path_ through the tree. Tab
 resolution follows MkDocs' own `content.tabs.link` semantics: tabs that share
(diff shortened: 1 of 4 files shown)


Rishet11:
Automated validation at `261229e4b617231459b88ebefbd913c9ca948dde` found one traversal case worth covering in this PoC: a tab group inside an admonition is parsed but silently omitted from the resolved recipe.

`_admonition()` recursively parses its children. `_absorb()` then feeds those children back into `_absorb()`, which has no `TabGroup` case, so the selected shell commands and artifacts disappear. With the same tabs outside the admonition, resolution works.

The 21 existing self-tests pass. This additional synthetic fixture reproduces the omission using the actual resolver:

```sh
PYTHONPATH=scripts/ci/doc-build-test python3 - <<'PY'
import mkdocs_build_extractor as ext
doc = '''# Build

!!! note "Choose target"
    === "Board A"
        ```bash
        ./build.sh board-a
        ```

        The resulting image will be placed at `out-a.rom`.
    === "Board B"
        ```bash
        ./build.sh board-b
        ```
'''
flat = '\n'.join(line[4:] if line.startswith('    ') else line
                 for line in doc.splitlines() if not line.startswith('!!!'))
for label, text in [('nested', doc), ('unwrapped control', flat)]:
    r = ext.resolve(text, select=['Board A'])
    print(label, r.commands, r.artifacts, r.caveats)
PY
```

Observed:

```text
nested [] [] ['Choose target']
unwrapped control ['./build.sh board-a'] ['out-a.rom'] []
```

Resolving admonition children through the path walker, while preserving the caveat, should prevent silently incomplete scripts. This was prepared and independently rerun by an AI assistant; impact on existing live manuals and full firmware builds remains unverified.

If a separately assigned regression-test follow-up under Dasharo/dasharo-issues#1153 would help, please confirm the scope and support amount under the bounty program before implementation.

